PR #1, opened 2021-05-21
Fix icij/datashare url that was 404ing

gtrufitt:
Was previously relative, this makes it absolute so it goes to the right place..

image


mbarton (approved):
Ah whoops, thanks @gtrufitt!

PR #4, opened 2021-05-24
OCR temp folder fixes

mbarton:
Follow up to #3 which set `TMPDIR` for OCRmyPDF:

- Remove `Uri( )` toString wrapper from temporary file name
- Write the output PDF to the same scratch directory we pass as `TMPDIR`

### DIFF
--- a/backend/app/utils/Ocr.scala
+++ b/backend/app/utils/Ocr.scala
@@ -47,9 +47,9 @@ object Ocr {
 
   // TODO MRB: allow OcrMyPdf to read DPI if set in metadata
   // OCRmyPDF is a wrapper for Tesseract that we use to overlay the OCR as a text layer in the resulting PDF
-  def invokeOcrMyPdf(lang: String, inputFileName: String, dpi: Option[Int], stderr: mutable.Buffer[String], tmpDir: Path): Path = {
-    val tempFile = Files.createTempFile("ocrmypdf", ".pdf")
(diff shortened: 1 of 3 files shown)


mbarton:
[[Giant] OCR temp dir fixes](https://trello.com/c/t2xFZkoS/1468-giant-ocr-temp-dir-fixes)

backend/app/utils/Ocr.scala
```diff
@@ -47,9 +47,9 @@ object Ocr {
 
   // TODO MRB: allow OcrMyPdf to read DPI if set in metadata
   // OCRmyPDF is a wrapper for Tesseract that we use to overlay the OCR as a text layer in the resulting PDF
-  def invokeOcrMyPdf(lang: String, inputFileName: String, dpi: Option[Int], stderr: mutable.Buffer[String], tmpDir: Path): Path = {
-    val tempFile = Files.createTempFile("ocrmypdf", ".pdf")
-    val cmd = s"ocrmypdf --redo-ocr -l $lang ${dpi.map(dpi => s"--image-dpi $dpi").getOrElse("")} ${inputFileName} ${tempFile.toAbsolutePath}"
+  def invokeOcrMyPdf(lang: String, inputFilePath: Path, dpi: Option[Int], stderr: mutable.Buffer[String], tmpDir: Path): Path = {
```
joelochlann, line 50: would it be any clearer/less error-prone to just have one param representing full path to file?

joelochlann (approved):

backend/app/utils/Ocr.scala
mbarton: Ah do you mean for the temporary file? I'm not sure it would be as we'd repeat the same logic in all the calling extractors

PR #5, opened 2021-05-24
Reprocess blob endpoint

mbarton:
We have an endpoint to reprocess everything in a workspace but sometimes it's better to just reprocess a single blob, especially if OCR is involved which can be quite slow.

The new endpoint has a query parameter `rerunSuccessful` which can be set to false to only re-run failed extractors. The default is true though to match how the existing reprocess workspace endpoint works.

To be consistent with the delete blobs API, this endpoint requires you have administrator access. Currently there is no UI so you must trigger it manually, through cURL with a token or through the `pfi-cli`:

```
pfi-cli api --verb POST /api/blobs/<blob id>/reprocess
```

### DIFF
--- a/backend/app/controllers/api/Blobs.scala
+++ b/backend/app/controllers/api/Blobs.scala
@@ -7,12 +7,13 @@ import play.api.mvc.{AnyContent, Request}
 import services.ObjectStorage
 import services.index.Index
 import services.manifest.Manifest
+import utils.Logging
 import utils.attempt.Attempt
 import utils.controller.{AuthApiController, AuthControllerComponents}
 
 class Blobs(override val controllerComponents: AuthControllerComponents, manifest: Manifest, index: Index,
             objectStorage: ObjectStorage, previewStorage: ObjectStorage)
-  extends AuthApiController {
+  extends AuthApiController with Logging {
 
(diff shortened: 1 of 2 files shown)


mbarton:
[[Giant] Reprocess blob endpoint](https://trello.com/c/fClGUr6H/1469-giant-reprocess-blob-endpoint)

joelochlann:
just checking how it looks when you pass the param - is it this?

```
pfi-cli api --verb POST /api/blobs/<blob id>/reprocess?rerunSuccessfulParam=false
```

joelochlann (approved):

mbarton:
Yep, exactly. Seeing that though I don't know why I added `param` on the end so I've renamed them and added two separate parameters for clarity, both of which default to true. I've also added them to the existing reprocess workspace endpoint.

```
run api --verb POST /api/blobs/v8EuuK2pyCfFuXTS6Iy_9MK3n7-BJe1WWPJUKha-gEkmKqBmOKkyV5TlAjQibCkWMk5-f8A_Oo8if│atkgEyjtQ/reprocess?rerunFailed=false
run api --verb POST /api/workspaces/708ec751-1fa6-4e6c-8b40-fbf7413bfa56/reprocess
run api --verb POST /api/workspaces/708ec751-1fa6-4e6c-8b40-fbf7413bfa56/reprocess?rerunSuccessful=false
```

PR #6, opened 2021-05-24
Incremental OCR logging and retry OCR after deploy (again)

mbarton:
Other OCR extractors catch the subprocess dying (such as during a Riff-Raff deploy) and don't mark it as a failure so the work can be picked up by a future worker. Unfortunately we didn't include this behaviour for the OCRmyPDF extractors. This PR adds it back in a base class to try and make it less likely we regress this again. Ideally we'd have integration tests for it too but this is a decent first step.

It also adds incremental logging for stderr of the subprocess. Before we only logged once everything was complete and it was impossible to get a sense of how far through OCRing a very large PDF we were.

Bonus feature, the last stderr log line is passed through the client as a progress note:

https://user-images.githubusercontent.com/395805/119389744-ed407080-bcc3-11eb-8fe6-9413d1a0bab5.mov

### DIFF
--- a/backend/app/utils/Ocr.scala
+++ b/backend/app/utils/Ocr.scala
@@ -8,13 +8,22 @@ import scala.collection.mutable
 import scala.sys.process._
 
 class OcrStderrLogger(setProgressNote: Option[String => Either[Failure, Unit]]) extends Logging {
+  val LOG_THROTTLE_RATE_MILLIS = 5000
+
   val acc = mutable.Buffer[String]()
+  var lastLogTime: Option[Long] = None
 
   def append(line: String): Unit = {
     acc.append(line)
 
     logger.info(line)
-    setProgressNote.foreach(f => f(line))
+
+    // Avoid spamming the database should the extractor output a lot of stderr logging quickly
+    val now = System.currentTimeMillis()
+    if(lastLogTime.isEmpty || lastLogTime.exists(t => (now - t) > LOG_THROTTLE_RATE_MILLIS)) {
+      setProgressNote.foreach(f => f(line))
+      lastLogTime = Some(now)
+    }
   }
 
   def getOutput: String = {
--- a/backend/app/utils/Ocr.scala
+++ b/backend/app/utils/Ocr.scala
@@ -1,12 +1,26 @@
 package utils
 
 import java.nio.file.{Files, Path}
-
 import services.TesseractOcrConfig
+import utils.attempt.Failure
 
 import scala.collection.mutable
 import scala.sys.process._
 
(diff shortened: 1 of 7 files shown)


mbarton:
[[Giant] Incremental OCR logging and note in UI](https://trello.com/c/AgfE0KQL/1470-giant-incremental-ocr-logging-and-note-in-ui)

backend/app/utils/Ocr.scala
```diff
@@ -41,19 +64,19 @@ object Ocr {
...
   // OCRmyPDF is a wrapper for Tesseract that we use to overlay the OCR as a text layer in the resulting PDF
-  def invokeOcrMyPdf(lang: String, inputFileName: String, dpi: Option[Int], stderr: mutable.Buffer[String], tmpDir: Path): Path = {
+  def invokeOcrMyPdf(lang: String, inputFileName: String, dpi: Option[Int], stderr: OcrStderrLogger, tmpDir: Path): Path = {
     val tempFile = Files.createTempFile("ocrmypdf", ".pdf")
     val cmd = s"ocrmypdf --redo-ocr -l $lang ${dpi.map(dpi => s"--image-dpi $dpi").getOrElse("")} ${inputFileName} ${tempFile.toAbsolutePath}"
 
     val stdout = mutable.Buffer.empty[String]
     val process = Process(cmd, cwd = None, extraEnv = "TMPDIR" -> tmpDir.toAbsolutePath.toString)
-    val exitCode = process.!(ProcessLogger(stdout.append(_), stderr.append(_)))
+    val exitCode = process.!(ProcessLogger(stdout.append(_), stderr.append))
```
joelochlann, line 79: is the removal of the second `_` significant or just syntactic sweetening?

joelochlann (approved):
Nice! I guess the number output by ocrmypdf is pages complete? Maybe in the client we could make the message a little friendlier. Will require a bit of string wrangling though, might be flaky

backend/app/utils/Ocr.scala
mbarton: Just sweetening, without it IntelliJ highlighted as possible to remove. I'm not sure why it was required before but not now this component is a class? I think the underscore is not necessary at all in Scala 3: https://dotty.epfl.ch/docs/reference/changed-features/eta-expansion.html

mbarton:
Yeah the output varies by file and what processing stages are triggered internally in OCRmyPDF, eg:

```
1 redoing OCR
Postprocessing...
Optimize ratio: 1.00 savings: 0.1%
Output file is a PDF/A-2B (as expected)
```

For the actual OCR though it includes the page number so reads like a counter. I figured it was probably going to be the most descriptive as is

PR #9, opened 2021-05-28
Fix fallback viewer if pages are unavailable

mbarton:
Pages can be missing in a few situations:

- An old upload that we haven't reprocessed yet
- `PdfOcrExtractor` failed whilst processing the file

In these cases there was code to fallback to the old viewer but it wasn't working. This PR fixes it by taking advantage of the page state now being in Redux to move the decision up a level to `Viewer` rather than `PageViewer`.

It also fixes a server-side error where we did `minBy` on the empty list of pages. That now returns a 404. As `Viewer` is now so complex ironically the simplest fix was to convert that to an empty list of pages in the Redux state and then fallback in that case. This is not ideal and the whole `Viewer` is very much due a refactor. I've added comments inline explaining our ideas for this.


docs/cheat-sheet.md
```diff
@@ -0,0 +1,36 @@
+# Cheat sheet
```
mbarton, line 1: @joelochlann should we move this into the runbook? Thinking about it there's a strong argument to start a `docs` folder in this repo with explainers, cheat sheets and tips per service

frontend/src/js/components/viewer/PageViewer/PageViewer.tsx
```diff
@@ -39,13 +30,6 @@ export default function PageViewer({ uri, q, fallbackRenderFn }: Props) {
         }
     }, [currentHighlightElement]);
 
-    useEffect(() => {
```
mbarton, line 42: The initial load is now done in `Viewer`

frontend/src/js/actions/pages/loadPages.ts
```diff
@@ -3,7 +3,11 @@ import { fetchPages } from "../../components/viewer/PageViewer/pageViewerApi";
 import { GiantAction, PagesActionType } from "../../types/redux/GiantActions";
 import { GiantState } from "../../types/redux/GiantState";
 
-export function loadPages(pageApiUri: string, viewportTop: number, viewportBottom: number, q?: string): ThunkAction<void, GiantState, void, GiantAction> {
+export function loadPages(uri: string, q?: string, maybeViewportTop?: number, maybeViewportBottom?: number): ThunkAction<void, GiantState, void, GiantAction> {
```
mbarton, line 6: I re-arranged the parameters to make the defaulting cleaner in `Viewer`

frontend/src/js/components/viewer/Viewer.tsx
```diff
@@ -195,6 +215,7 @@ class Viewer extends React.Component<Props, State> {
     componentWillUnmount() {
         document.title = "Giant";
         this.props.resetResource();
+        this.props.resetPages();
```
mbarton, line 218: Without this the old page state hangs around and we can't properly calculate the fallback logic if you click back to Search and then into the Viewer again via a seach result

scripts/setup.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
mbarton, line 1: I forgot to include this script when we opened the repo 🤦

mbarton:
[Targeted fix for the `empty minBy` error when using the Page Viewer in PROD](https://trello.com/c/yq1YSTS9/1349-targeted-fix-for-the-empty-minby-error-when-using-the-page-viewer-in-prod)

docs/cheat-sheet.md
joelochlann: yeah anything here that's not in runbook already, let's move there

joelochlann (approved):
I think this looks merge-able! I'm assuming the race between pages & full resource request doesn't make anything too weird for the user. And that intermediate loading states don't look too crazy etc

Would be good to refactor the pages state a bit soon, along the lines we discussed. enum between

- first page request hasn't happened yet
- page request pending
- request came back successfully
- request came back with an error

PR #10, opened 2021-07-21
Do not alarm on failure to parse token

joelochlann:
We shouldn't alarm on "failed to parse token", since this happens whenever the token is missing, as a result of vulnerability scanners and occasionally developer testing/debugging (e.g. just opening https://giant.pfi.gutools.co.uk/api/search in your browser will fire this alarm). We get spurious alarms most weeks because of this

### DIFF
--- a/backend/app/utils/auth/AuthActionBuilder.scala
+++ b/backend/app/utils/auth/AuthActionBuilder.scala
@@ -1,7 +1,5 @@
 package utils.auth
 
-import pdi.jwt.JwtSession
-
 import java.time.{Clock, Instant, LocalDateTime, ZoneId, ZoneOffset}
 import pdi.jwt.JwtSession._
 import play.api.Configuration
@@ -76,29 +74,37 @@ class DefaultAuthActionBuilder(val controllerComponents: ControllerComponents, f
   private[auth] def invokeBlockWithTime[A](request: Request[A], block: (UserIdentityRequest[A]) => Future[Result],
                                            now: Long): Future[Either[Failure, Result]] = {
     implicit val implicitReq = request
+    val claimData = request.jwtSession.claimData
 
-    // Why can't we do request.jwtSession.isEmpty?
-    // Because request.jwtSession gets initialised with some default properties if there's no
-    // actual JWT token in the request. This seems a little odd to me, but it's in the pdi.jwt
-    // library so we can't control this behaviour.
-    if (request.headers.get(JwtSession.REQUEST_HEADER_NAME).isEmpty) {
-      val msg = s"No token in request"
-      logger.warn(msg)
-      Future.successful(Left(AuthenticationFailure(msg, reportAsFailure = false)))
-    } else {
-      val claimData = request.jwtSession.claimData
-      val maybeToken = claimData.validate[Token]
-      maybeToken match {
-        case (JsSuccess(token, _)) if token.loginExpiry > now =>
-          handleValidUnexpiredToken(token, request, block, now)
+    val maybeToken = claimData.validate[Token]
+    maybeToken match {
+      case (JsSuccess(token, _)) if token.loginExpiry > now =>
+        handleValidUnexpiredToken(token, request, block, now)
 
-        case JsSuccess(token, _) => {
-          val msg = s"Token is older than $maxLoginAge"
-          logger.info(token.user.asLogMarker, msg)
-          Future.successful(Left(AuthenticationFailure(msg, reportAsFailure = false)))
+      case JsSuccess(token, _) => {
+        val msg = s"Token is older than $maxLoginAge"
+        logger.info(token.user.asLogMarker, msg)
+        Future.successful(Left(AuthenticationFailure(msg, reportAsFailure = false)))
+      }
+
+      case JsError(errors) => {
+        // It would be nice to just call request.jwtSession.isEmpty right at the top,
+        // instead of all this. But request.jwtSession gets initialised with some default
+        // properties if there's no actual JWT token in the request (which seems a little
+        // odd to me, but it's in the pdi.jwt library so we can't control this behaviour).
+        val isTokenMissing = errors.forall {
+          case (_, pathErrors) =>
+            pathErrors.forall {
+              case JsonValidationError(messages) => messages.forall(_ == "error.path.missing")
+              case _ => false
+            }
+          case _ => false
         }
-
-        case JsError(errors) => {
+        if (isTokenMissing) {
+          val msg = s"No token in request"
+          logger.warn(msg)
+          Future.successful(Left(AuthenticationFailure(msg, reportAsFailure = false)))
+        } else {
           val msg = s"Failed to parse token: $errors"
           logger.warn(msg)
           Future.successful(Left(AuthenticationFailure(msg, reportAsFailure = true)))
--- a/backend/app/utils/auth/AuthActionBuilder.scala
+++ b/backend/app/utils/auth/AuthActionBuilder.scala
@@ -1,5 +1,7 @@
 package utils.auth
 
+import pdi.jwt.JwtSession
+
 import java.time.{Clock, Instant, LocalDateTime, ZoneId, ZoneOffset}
 import pdi.jwt.JwtSession._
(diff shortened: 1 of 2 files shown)


joelochlann:
[Investigate Giant failure alarms triggered by "Authentication failure: Failed to parse token:"](https://trello.com/c/YBuTHscq/1515-investigate-giant-failure-alarms-triggered-by-authentication-failure-failed-to-parse-token)

joelochlann:
[Investigate Giant and Playground alarms that fired on Thurs 15th July and Mon 19th July](https://trello.com/c/qcPFacZK/1528-investigate-giant-and-playground-alarms-that-fired-on-thurs-15th-july-and-mon-19th-july)

backend/app/utils/auth/AuthActionBuilder.scala
```diff
@@ -33,60 +34,75 @@ class DefaultAuthActionBuilder(val controllerComponents: ControllerComponents, f
...
-        Future.successful(Left(AuthenticationFailure(msg, reportAsFailure = true)))
+  private[auth] def invokeBlockWithTime[A](request: Request[A], block: (UserIdentityRequest[A]) => Future[Result],
+                                           now: Long): Future[Either[Failure, Result]] = {
+    implicit val implicitReq = request
+
+    // Why can't we do request.jwtSession.isEmpty?
+    // Because request.jwtSession gets initialised with some default properties if there's no
+    // actual JWT token in the request. This seems a little odd to me, but it's in the pdi.jwt
+    // library so we can't control this behaviour.
+    if (request.headers.get(JwtSession.REQUEST_HEADER_NAME).isEmpty) {
```
joelochlann, line 84: FYI here's where it sets `request.jwtSession`:

https://github.com/jwt-scala/jwt-scala/blob/v4.3.0/play/src/main/scala/JwtPlayImplicits.scala#L19

backend/app/utils/auth/AuthActionBuilder.scala
```diff
@@ -33,60 +34,75 @@ class DefaultAuthActionBuilder(val controllerComponents: ControllerComponents, f
...
 
-      case JsError(errors) => {
-        val msg = s"Failed to parse token: $errors"
-        logger.warn(msg)
-        Future.successful(Left(AuthenticationFailure(msg, reportAsFailure = true)))
+  private[auth] def invokeBlockWithTime[A](request: Request[A], block: (UserIdentityRequest[A]) => Future[Result],
+                                           now: Long): Future[Either[Failure, Result]] = {
+    implicit val implicitReq = request
+
+    // Why can't we do request.jwtSession.isEmpty?
```
amyhughes, line 80: Hmm yes, the jwt-scala docs on the [JwtSession](https://jwt-scala.github.io/jwt-scala/api/pdi/jwt/JwtSession.html) say this about the two components of the `JwtSession`, `headerData` and `claimData`:

`
The data is separated into two attributes: headerData and claimData. There is also an optional signature. Most of the time, you should only care about the claimData which stores the claim of the token containing the custom values you eventually put in it. That's why all methods of JwtSession (such as add and removing values) only modifiy the claimData.
`


The comments on [this example](https://jwt-scala.github.io/jwt-scala/jwt-play-jwt-session.html) suggest that a default session would have empty `claimData`, meaning `session.isEmpty()` would be `true` (despite the fact that `headerData` is not empty):
```
var session = JwtSession()
// session: JwtSession = JwtSession(
//   headerData = JsObject(
//     underlying = Map(
//       "typ" -> JsString(value = "JWT"),
//       "alg" -> JsString(value = "HS256")
//     )
//   ),
//   claimData = JsObject(underlying = Map()),
//   signature = ""
// )
```

When our app creates a session, the session is populated from what is in `application.conf`. Since we set maxAge [here](https://github.com/guardian/giant/blob/main/backend/conf/application.conf#L4), `exp` is set on the `claimData` and the session is not considered empty (see [here](https://www.playframework.com/documentation/2.8.x/SettingsSession#Session-Timeout-/-Expiration) in the Play docs):

```
play {
  # 30 minute JWT expiry
  http.session {
    maxAge = 1800000
    jwtResponseName = "X-Offer-Authorization"
  }
...
}
```

Printing the `headerData` and `claimData` on a request to an authenticated endpoint I get:
```
claimData: {"permissions":{"granted":["CanPerformAdminOperations"]},"refreshedAt":1627037211879,"issuedAt":1627037211879,"exp":1627039011,"user":{"username":"amy","displayName":"amy"},"verificationExpiry":1627037271879,"loginExpiry":1627066011879}
headerData: {"typ":"JWT","alg":"HS256"}
``` 
Whereas when I hit /api/search I get:
```
claimData: {"exp":1627039113}
headerData: {"typ":"JWT","alg":"HS256"}
```

I wonder whether we shouldn't be checking the `claimData` rather than the `headerData` given this? We can't use `session.isEmpty()` but we could check whether it only has the `exp` property set (which is a little like the case where we fail to validate it as a `Token`, maybe we should just add a bit to the existing log message for `JsError(errors)`?)

joelochlann: We're not current checking `headerData`, we're just checking the presence of the header itself, which seemed to me the quickest route to diagnose "token not present in request", which is the event actually triggering the alarm.

We could look at `claimData` but then we'd need to infer that the token was missing or have a more convoluted message like "permissions not present in `claimData`".

Wasn't quite sure what you meant by adding to the existing `JsError(errors)` log message. This is a catch-all for any JSON parsing error. I think we still want to alarm on other parsing errors since it's an exceptional circumstance (haven't seen it before, would want to diagnose). So you need to do some further checking or parsing in that condition to determine whether the token was entirely missing or not, which was what I had earlier but seemed needlessly complex:

https://github.com/guardian/giant/pull/10/commits/98d66e491cd4ccd9f21204a9508531a7a4ddd111

joelochlann: The reason you see that minimal `claimData` and `headerData` even when you've supplied no token is because of this: https://github.com/jwt-scala/jwt-scala/blob/v4.3.0/play/src/main/scala/JwtSession.scala#L150

So that example from the docs seems misleading, or maybe applies to a different version of the library. `claimData` will be populated with the [default claim](https://github.com/jwt-scala/jwt-scala/blob/v4.3.0/play/src/main/scala/JwtSession.scala#L140-L143) when there's no token.

backend/app/utils/auth/AuthActionBuilder.scala
joelochlann: and here's where the `JwtSession` case class gets initialised with default stuff if called with no params: https://github.com/jwt-scala/jwt-scala/blob/v4.3.0/play/src/main/scala/JwtSession.scala#L150

backend/app/utils/auth/AuthActionBuilder.scala
amyhughes: > `claimData` will be populated with the [default claim](https://github.com/jwt-scala/jwt-scala/blob/v4.3.0/play/src/main/scala/JwtSession.scala#L140-L143) when there's no token.

I agree, I think this is consistent with the docs. In our case, we set the a value for MAX_AGE [here](https://github.com/guardian/giant/blob/main/backend/conf/application.conf#L4) in our `application.conf`, which is why our default claim is not empty (responding to your comment `// Why can't we do request.jwtSession.isEmpty?`).

None of this is really the point though, I think what makes me a little uneasy about this change is that it introduces a fair bit of extra logic for the case of verifying that there was in fact no token in the request rather than some other reason for us not being able to parse a token and handling these differently. I'd be tempted to keep the code simple and just downgrade the `reportAsFailure = true` to `false` on the failed token parsing. It's your call though, if you think there is value in seperating out the different types of failure which is worth the extra complexity to the code, I agree that looking for the header in the request is probably the most explicit way of doing it.

amyhughes (approved):

backend/app/utils/auth/AuthActionBuilder.scala
joelochlann: I take this point, it's a smaller change just to not report as failure. Have changed

PR #12, opened 2021-10-01
Allow downloading in prod

itsibitzi:
Enable downloading in AWS so that people can download stuff. 

Reasoning has been discussed offline.


philmcmahon (approved):
Love what you did with line 34

PR #13, opened 2021-11-09
Correct es link in dev tools

joelochlann:
I *think* these links should've changed when we upgraded to ES 7

https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html

> **Elasticsearch 7.x**
> Specifying types in requests is deprecated. For instance, indexing a document no longer requires a document type. 
> The new index APIs are PUT {index}/_doc/{id} in case of explicit ids and POST {index}/_doc for auto-generated ids. : > Note that in 7.0, _doc is a permanent part of the path, and represents the endpoint name rather than the document type.*


philmcmahon (approved):

joelochlann:
here's where they are, for reference...

Screenshot 2021-11-09 at 14 35 19

PR #15, opened 2021-11-10
Add new metrics for ingest failures

philmcmahon:
## What does this change?
We recently performed a large ingest on giant. This resulted in a lot of alarms. The majority of these were triggered by the failuretoresultmapper metric - which alarms whenever there's more than 1 error. The link between worker errors and failuretoresultmapper was added in https://github.com/guardian/pfi/pull/925 in order to alarm when workers fail. However, grouping frontend errors (e.g. auth failure 5xx/4xx errors) that users experience with ingest errors isn't very helpful when it comes to setting a threshold for triggering an alarm on - as we have a small number of users we want to be notified for every time a user hits an error, but we don't want an alarm when, during a 10,000 file ingest there is a single failure.

This PR disconnects the workers from failuretoresultsmapper and instead adds 4 new metrics to try and capture ingest failures:
 - itemsIngested
 - itemsFailed
 - batchesIngested
 - batchesFailed

I'm not convinced about splitting these up into batches and items, but mixing them feels like a bad idea. The motiviation for measuring success as well as failures is so that we can see at a glance what proportion of files in an ingest are failing. This could be used to e.g. alarm when > 10% of stuff in an ingest fails - or we may still want to alarm on every worker failure

## How to test
I've tested this on playground using a small ingest and a manually generated dodgy PDF file and the metrics came out [as predicted](https://eu-west-1.console.aws.amazon.com/cloudwatch/home?region=eu-west-1#metricsV2:graph=~(metrics~(~(~(expression~'100*2a*28m1*2fm2*29~label~'Expression1~id~'e1))~(~'PFI~'IngestItemsFailed~'Stage~'rex~'Stack~'pfi-playground~(id~'m1))~(~'.~'IngestedItems~'.~'.~'.~'.~(id~'m2)))~view~'timeSeries~stacked~false~region~'eu-west-1~start~'-PT1H~end~'P0D~stat~'Maximum~period~300);query=~'*7bPFI*2cStack*2cStage*7d).

A new alarm will be needed to maintain the existing alarm behaviour https://github.com/guardian/investigations-platform/pull/392

ingest problems

### DIFF
--- a/backend/app/extraction/Worker.scala
+++ b/backend/app/extraction/Worker.scala
@@ -5,17 +5,16 @@ import java.util.concurrent.atomic.AtomicInteger
 import cats.syntax.either._
 import extraction.Worker.Batch
 import model.manifest.{Blob, WorkItem}
-import services.ObjectStorage
+import services.{Metrics, MetricsService, ObjectStorage}
 import services.manifest.WorkerManifest
 import utils.Logging
 import utils.attempt._
-import utils.controller.FailureToResultMapper
 
 import scala.concurrent.ExecutionContext
 import scala.language.postfixOps
 import scala.util.control.NonFatal
 
-object Worker {
+object Worker extends Logging {
   type Batch = List[(Extractor, Blob, ExtractionParams)]
 }
 
@@ -24,7 +23,7 @@ class Worker(
   manifest: WorkerManifest,
   blobStorage: ObjectStorage,
   extractors: List[Extractor],
-  failureToResultMapper: FailureToResultMapper)(implicit executionContext: ExecutionContext) extends Logging {
+  metricsService: MetricsService)(implicit executionContext: ExecutionContext) extends Logging {
 
   private val maxBatchSize = 1000 // tasks
   private val maxCost = 100 * 1024 * 1024 // 100MB
@@ -36,7 +35,9 @@ class Worker(
       completed
     }.recoverWith {
       case err =>
-        failureToResultMapper.failureToResult(err)
+        metricsService.updateCloudwatchMetric(Metrics.batchesFailed)
+        // on failure, fetchBatch just returns the first failure
+        logger.error("Error executing batch", err)
         manifest.releaseLocks(name)
         Attempt.Left(err)
     }
--- a/backend/app/extraction/Worker.scala
+++ b/backend/app/extraction/Worker.scala
@@ -88,7 +88,8 @@ class Worker(
 
         case Left(failure) =>
           markAsFailure(blob, extractor, failure)
-          failureToResultMapper.failureToResult(failure)
+          metricsService.updateCloudwatchMetric(Metrics.itemsFailed)
+          logger.warn(s"Ingest batch execution failure, ${failure.msg}", failure.toThrowable)
           completed
       }
     }
--- a/backend/app/extraction/Worker.scala
+++ b/backend/app/extraction/Worker.scala
@@ -35,7 +35,7 @@ class Worker(
       completed
     }.recoverWith {
       case err =>
-        metricsService.updateCloudwatchMetric(Metrics.batchesFailed)
+        metricsService.updateMetric(Metrics.batchesFailed)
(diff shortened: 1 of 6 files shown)


backend/app/services/MetricsService.scala
```diff
@@ -0,0 +1,76 @@
...
+      .withMetricName(name)
+      .withTimestamp(new Date())
+      .withDimensions(dimensions: _*)
+      .withValue(value)
+  }
+}
+
+trait MetricsService {
+  def updateCloudwatchMetrics(metrics:List[MetricUpdate]): Unit
+  def updateCloudwatchMetric(metricName: String, metricValue: Double = 1): Unit
```
itsibitzi, line 34: Can these be `updateMetric` and `updateMetrics` - I don't think it makes sense to refer to a specific implementation of a metric service from the trait.

backend/app/services/MetricsService.scala
```diff
@@ -0,0 +1,76 @@
...
+}
+
+class DefaultMetricsService() extends MetricsService {
```
itsibitzi, line 37: Can this be called `NoOpMetricsService` - I think that would be more descriptive and accurate?

backend/app/extraction/Worker.scala
```diff
@@ -87,7 +88,8 @@ class Worker(
 
         case Left(failure) =>
           markAsFailure(blob, extractor, failure)
-          failureToResultMapper.failureToResult(failure)
+          metricsService.updateCloudwatchMetric(Metrics.itemsFailed)
+          logger.warn(s"Ingest batch execution failure, ${failure.msg}", failure.toThrowable)
```
itsibitzi, line 92: Should be an error instead of warn?

backend/app/ingestion/phase2/IngestStorePolling.scala
```diff
@@ -71,12 +72,15 @@ class IngestStorePolling(
...
-                  failureToResultMapper.failureToResult(failure)
+                  metricsService.updateCloudwatchMetric(Metrics.itemsFailed)
                   logger.warn(s"Failed to process $key: $failure")
                 case _ => ingestStorage.delete(key)
               }
               result
             }.collect { case Right(success) => success }
+            metricsService.updateCloudwatchMetrics(List(
+              MetricUpdate(Metrics.itemsIngested, results.size),
+              MetricUpdate(Metrics.batchesIngested, 1)))
```
itsibitzi, line 83: A bit confused at this number. Should it be the number of batches ingested as part of the overall dataset? I'm not sure the worker has the information to know - feel free to ignore.

backend/app/services/MetricsService.scala
```diff
@@ -0,0 +1,76 @@
...
+  val namespace = "PFI"
+
+  val itemsIngested = "IngestedItems"
+  val itemsFailed = "IngestItemsFailed"
+  val batchesIngested = "IngestBatchesIngested"
+  val batchesFailed = "IngestBatchesFailed"
+  val failureToResultMapper = "ErrorsInGiantFailureToResultMapper"
+
+
+  def metricDatum(name: String, dimensions: List[Dimension], value: Double): MetricDatum = {
```
itsibitzi, line 23: If we don't support non-stack/stage metrics, maybe make this `protected`?

itsibitzi (changes requested):
Just a few minor changes requested.

backend/app/ingestion/phase2/IngestStorePolling.scala
philmcmahon: ah the '1' is the number that failed and I think in this case that's just a single batch that's failed?

philmcmahon:
@itsibitzi I think this is ready to go now

itsibitzi:
Argh, the protected method would have to be in the `MetricsService` too apols for that.

philmcmahon:
Ah yeah I only just noticed the TC build failure

joelochlann (approved):

PR #19, opened 2021-12-17
Bump sbt version

philmcmahon:
## What does this change?
Bump SBT version following advice from DevX

Tested on playground - seems to work!


MarSavar (approved):
👍

PR #20, opened 2021-12-21
Upgrade log4j version to 2.17.0

MarSavar:
Another log4j vulnerability has been identified - this one is high severity (instead of critical) but we should still upgrade within 14 days.

The version we're on (2.16) doesn't protect us from infinite recursion in lookup evaluation, meaning intentional recursive lookups could cause stack overflow errors that would lead to a crashing app (DOS attack).

More info here: https://logging.apache.org/log4j/2.x/index.html

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -16,7 +16,7 @@ val compilerFlags = Seq(
 )
 
 val awsVersion = "1.11.566"
-val log4jVersion = "2.16.0"
+val log4jVersion = "2.17.0"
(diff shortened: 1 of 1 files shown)


joelochlann (approved):

MarSavar:
[Upgrade log4j to v2.17.0](https://trello.com/c/6tYAoNod/1812-upgrade-log4j-to-v2170)

PR #32, opened 2022-02-25
Page viewer spanning newlines and rotation

itsibitzi:
Break up highlights into spans which can be put across some new lines. These spans count as a single highlight in the next-next-next button so you don't have to press it any extra times.

Also does some basic math to make rendering work with rotations. Which revealed some shortcomings in how we're able to find text which is at weird angles. But i guess this is uncommon. Or it could be an issue with the open office export to PDF fucntion - which is what I used for the test docs.

NB: I also removed the routine for doing the highlighting inline inside the PDF as it prevents us from doing next-next-next. I feel the next-next-next feature is super important so I got rid of the code from `HEAD` for now.

## Example

Searching for `testing how highlights`

### Before:

image

### After:

image


backend/conf/evolutions/default/1.sql
```diff
@@ -0,0 +1,39 @@
+CREATE EXTENSION "uuid-ossp";
+
+CREATE TABLE collections (
```
joelochlann, line 3: did you mean to commit this?

itsibitzi: whoops.

itsibitzi: Removed!

philmcmahon (approved):
Banging. 
highlight gif

backend/app/utils/PDFUtil.scala
```diff
@@ -75,47 +76,56 @@ object PDFUtil {
...
-        highlightBox.setQuadPoints(quadPoints)
+        // The highlight is rendered slightly off, we need to move the position by about 0.75 * the max character height
+        // while also factoring in rotation. Here were getting the x/y offsets by rotating a Y axis unit vector and then
+        // multiplying the result by the offset magnitude. The Y axis has to be negated due to the coordinate spaces.
 
-        highlightBox.setColor(highlightColour)
-        highlightBox.setPrinted(true)
+        val offsetMagnitude = height * 0.75
+        val offsetX = offsetMagnitude * sin(rotation)
+        val offsetY = -offsetMagnitude * cos(rotation)
```
joelochlann, line 123: omg the joy of getting to use trig functions in actual code! well jel

PR #33, opened 2022-02-25
JDeb packaging symlink fix

itsibitzi:
Ronseal.

~We weren't including the public folder in the build.~

It looks like teamcity wasn't resolving the symlink correctly in the JDeb build - so manually `cp` it instead.

### DIFF
--- a/backend/app/AppComponents.scala
+++ b/backend/app/AppComponents.scala
@@ -228,7 +228,8 @@ class AppComponents(context: Context, config: Config)
       usersController,
       authController,
       appController,
-      genesisController
+      genesisController,
+      assets
     )
   } catch {
     case NonFatal(e) =>
--- a/backend/app/controllers/frontend/App.scala
+++ b/backend/app/controllers/frontend/App.scala
@@ -18,10 +18,6 @@ class App (components: ControllerComponents, assets: Assets, config: Config,
     assets.at("index.html")
   }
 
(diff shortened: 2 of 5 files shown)


joelochlann (approved):
oh wow, how did page viewer ever work then? think we must've had some other mechanism to get pdf.js served in prod, at some point... can't remember what it was though

itsibitzi:
The issue seemed to be that the JDeb build step in sbt wasnt following a symbolic link when packaging, but only on teamcity.

It's possible this used to work on Teamcity but broke somehow, which would explain why it used to work but doesn't any more.

For now I'm happy just manually copying stuff.

itsibitzi:
Big thanks to Phil for helping uncover this weirdness. I'd like to raise the issue in DevX on Monday.

philmcmahon:
Just a note on this - when we say 'teamcity' it's just an ubuntu machine running a bash script so there has to be something different between the environments of running locally versus in TC... Maybe it's the JVM version? TC runs java 8

philmcmahon:
(Having said that, I think it's a  waste of time debugging TC as it might be gone in a year or two, so probably fine to just leave this)

itsibitzi:
Yeah it's pretty confusing. FWIW the local machine I was testing on was also an ubuntu machine running the same bash script... It could possibly be some subtle bug caused by a different JVM, even if they're running the same Java version.

Ultimately, as Phil says I don't think its worth investing too much time root causing this issue.

PR #34, opened 2022-03-03
PageViewer: Mk. III Mod. A

itsibitzi:
Just thought I'd open a PR now because this could get pretty long. This is not production ready but has approximate feature parity with the existing page viewer.

The main intention of this rework is to remove the synchronous nature of the old API in order to give us a well performing foundation for the page viewer MVP.

Rather than modify the existing code I've added new Typescript/Scala files. The older versions can be retired in the future if we decide this is a good approach.

You can access the new version of the page viewer by changing `/viewer/:uri` to `/viewer2/:uri` in this build.

I'll try to be very descriptive in this PR for future info but will happily run through after standup or whatever.

## Previous version

https://user-images.githubusercontent.com/5560113/156780098-b6732ee0-7e95-44e6-af71-ae8fdfb450f8.mov

Previously a few queries to Elasticsearch were sent off and those got the visible pages using a pixel offset in the client's window, once we had the page numbers from ES we could then fetch the individual PDF pages from S3. The first API call did *4* separate calls to Elasticsearch - which in a production environment where ES is on the other side of a network connection produced some pretty unfortunate performance characteristics. Additionally - since the page numbers aren't known until after this call - downloading the PDF pages was blocked until this was complete. In my rough testing I've seen the time to paint a single PDF page be ~800ms in just network latency.

## New version

https://user-images.githubusercontent.com/5560113/156780154-a1eb618f-243b-4bdb-9aa9-d5b22641328b.mov

The new version does the same fundamental work, but reorganised to decrease perceived latency, which you can think of as "time to paint" an individual page to the DOM. It mainly achieves this by allowing the pages to be fetched asynchronously to the text and highlights, caching, and by prefetching results which are off page using a virtual scrolling mechanism.

Information that is only needed once, currently only total page count, is fetched individually to reduce the amount of work done per-page request.

Pages are fetched individually which is a latency/throughput trade-off, I've come down hard on the side of latency. 

Pages are fetched by their page number, which removes the synchronous API call, but at the cost of some rendering complexity. Because we don't know how big the pages dimension are until we fetch the results it's impossible to size the parent scroll area to be the correct height. To get around this we contain individual pages inside a bounding box, which is 1000x1000 pixels at the moment. This also makes having margins and whatnot between pages very easy!

Note that this bounding box approach makes it possible to link to a specific page, which previously would have been expressed as linking to a pixel offset. This would have worked fine, if a bit weird.

I've also made a cache where we render the PDFs to off-DOM canvases which are added to the containers as they scroll into view.

Highlights and text overlays are done separately to the canvas paint - right now the computation of the text overlays is not cached which is not ideal but is much more straight forward.

## Other stuff...
A side benefit of the rewrite is it uses our more modern understanding of react hooks - the original version was written as a first foray into hooks and was a bit messy (at least, according to a Señor engineer formally of this parish who shall go unnamed). I hope the code is easier to follow.

# Production performance data to come...

### DIFF
--- a/frontend/src/js/components/PageViewer/PageCache.ts
+++ b/frontend/src/js/components/PageViewer/PageCache.ts
@@ -1,8 +1,12 @@
 import _ from "lodash";
 import authFetch from "../../util/auth/authFetch";
 import * as pdfjs from "pdfjs-dist";
-import { Page } from "./model";
+import { CONTAINER_SIZE, Page } from "./model";
 
+export type Preview = {
+  canvas: HTMLCanvasElement;
+  scale: number;
+};
 export class PageCache {
   uri: string;
 
@@ -10,7 +14,7 @@ export class PageCache {
 
   // Tracking of text (Page}) and rendered previews (HTMLCanvasElement) is handled seperately incase an
   // eviction happens to an item that is in flight
-  previews: HTMLCanvasElement[] = [];
+  previews: Preview[] = [];
   cachedPreviewsLru: number[] = [];
 
   pages: Page[] = [];
@@ -24,10 +28,7 @@ export class PageCache {
   // Previews //
   //////////////
 
-  private addToPreviewCache = (
-    pageNumber: number,
-    canvas: HTMLCanvasElement
-  ) => {
+  private addToPreviewCache = (pageNumber: number, canvas: Preview) => {
     this.previews[pageNumber] = canvas;
     this.cachedPreviewsLru.push(pageNumber);
     if (this.cachedPreviewsLru.length > PageCache.MAX_CACHED_PAGES) {
@@ -45,7 +46,7 @@ export class PageCache {
   };
 
   // Get a cached canvas with the page rendered into it
-  getPagePreview = async (pageNumber: number): Promise<HTMLCanvasElement> => {
+  getPagePreview = async (pageNumber: number): Promise<Preview> => {
     if (this.previews[pageNumber]) {
       this.bumpPreviewRecency(pageNumber);
 
@@ -66,8 +67,8 @@ export class PageCache {
       const unscaledViewport = pdfPage.getViewport({ scale: 1.0 });
       const isLandscape = unscaledViewport.width > unscaledViewport.height;
 
-      const widthScale = 1000 / unscaledViewport.width;
-      const heightScale = 1000 / unscaledViewport.height;
+      const widthScale = CONTAINER_SIZE / unscaledViewport.width;
+      const heightScale = CONTAINER_SIZE / unscaledViewport.height;
 
       const scale = isLandscape ? widthScale : heightScale;
 
@@ -83,9 +84,11 @@ export class PageCache {
       });
 
       // Handle caching
-      this.addToPreviewCache(pageNumber, canvas);
+      const preview = { canvas, scale };
 
-      return canvas;
+      this.addToPreviewCache(pageNumber, preview);
+
+      return preview;
     }
   };
 
--- a/frontend/src/js/components/PageViewer/PageCache.ts
+++ b/frontend/src/js/components/PageViewer/PageCache.ts
@@ -1,100 +1,58 @@
 import authFetch from "../../util/auth/authFetch";
-import { CachedPreview, Page } from "./model";
+import { LruCache } from "../../util/LruCache";
+import { CachedPreview, PageData } from "./model";
 import { renderPdfPreview } from "./PdfHelpers";
+
+export type CachedPageData = {
+  previewAbortController: AbortController;
+  preview: Promise<CachedPreview>;
+  dataAbortController: AbortController;
+  data: Promise<PageData>;
+};
 
 export class PageCache {
   uri: string;
+  query?: string;
 
   static MAX_CACHED_PAGES = 100;
 
-  // Tracking of text (Page}) and rendered previews (HTMLCanvasElement) is handled seperately incase an
-  // eviction happens to an item that is in flight
-  previews: CachedPreview[] = [];
-  cachedPreviewsLru: number[] = [];
+  private cache: LruCache<number, CachedPageData>;
 
-  pages: Page[] = [];
-  cachedPagesLru: number[] = [];
-
-  constructor(uri: string) {
+  constructor(uri: string, query?: string) {
     this.uri = uri;
+    this.query = query;
+    this.cache = new LruCache(25, this.onCacheMiss, this.onCacheEvict);
   }
 
-  //////////////
-  // Previews //
-  //////////////
+  private onCacheMiss = (pageNumber: number): CachedPageData => {
+    const previewAbortController = new AbortController();
+    const preview = authFetch(`/api/pages2/${this.uri}/${pageNumber}/preview`, {
+      signal: previewAbortController.signal,
+    })
+      .then((res) => res.arrayBuffer())
+      .then((buf) => renderPdfPreview(buf));
 
-  private addToPreviewCache = (pageNumber: number, canvas: CachedPreview) => {
-    this.previews[pageNumber] = canvas;
-    this.cachedPreviewsLru.push(pageNumber);
-    if (this.cachedPreviewsLru.length > PageCache.MAX_CACHED_PAGES) {
-      const lruPageNumber = this.cachedPreviewsLru.shift();
-      if (lruPageNumber) {
-        delete this.previews[lruPageNumber];
-      }
-    }
+    const dataAbortController = new AbortController();
+    const data = authFetch(
+      `/api/pages2/${this.uri}/${pageNumber}/text${
+        this.query ? `?q=${this.query}` : ""
+      }`,
+      { signal: dataAbortController.signal }
+    ).then((res) => res.json());
+
+    return {
+      previewAbortController,
+      preview,
+      dataAbortController,
+      data,
+    };
   };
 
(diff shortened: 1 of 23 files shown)


frontend/src/js/components/PageViewer/VirtualScroll.tsx
```diff
@@ -0,0 +1,82 @@
...
+  totalPages,
+  initialPage,
+  renderPage,
+}) => {
+  // Tweaked this and 2 seems to be a good amount on a regular monitor
+  // The fewer pages we preload the faster the initial paint will be
+  // Could possibly make it dynamic based on the visible of the container
+  const PRELOAD_PAGES = 2;
+
+  const pageHeight = CONTAINER_AND_MARGIN_SIZE;
```
mbarton, line 22: Shower though: could you store the maximum page height as metadata on the document along with page count? That way you know you’re guaranteed to get the next page even if for whatever reason it’s bigger than the hard coded bounding box

itsibitzi: Yeah I was thinking about this too, the thing that put me off the idea was if there's some weird PDF where a given page is *massive* then it might be preferable that we actually constrain it?

frontend/src/js/components/PageViewer/model.ts
```diff
@@ -0,0 +1,57 @@
...
+export type PdfText = {
+  value: string;
+  left: string;
+  top: string;
+  fontSize: string;
+  fontFamily: string;
+  transform: string;
+};
+
+export type Page = {
```
joelochlann, line 41: maybe `PageData` to distinguish from the page contents itself

frontend/src/js/components/PageViewer/VirtualScroll.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+      const scrollTo =
+        initialPage !== 1 ? initialPage * pageHeight - v.clientHeight / 2 : 0;
+
+      v.scrollTop = scrollTo;
+    }
+  }, [viewport, initialPage, pageHeight]);
+
+  return (
+    <div ref={viewport} className="viewer__main" onScroll={onScroll}>
```
joelochlann, line 61: as per conversation, let's copy all existing styles scoped to the page viewer into the CSS module, so the frontend code is more clearly contained

frontend/src/js/components/PageViewer/VirtualScroll.tsx
```diff
@@ -0,0 +1,82 @@
...
+      setBotPage(botPage);
+    }
+  };
+
+  useLayoutEffect(() => {
+    if (viewport?.current && initialPage) {
+      const v = viewport.current!;
+
+      const scrollTo =
+        initialPage !== 1 ? initialPage * pageHeight - v.clientHeight / 2 : 0;
```
joelochlann, line 54: I'd love some redundant brackets here for readability. Or separated out by line

joelochlann: assume this scrolls to centre of page... maybe top is nicer? haven't tried it though, might be totally wrong. worth a comment to say what it does in any case

frontend/src/js/components/PageViewer/VirtualScroll.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+      setTopPage(topPage);
+      //setMidPage(midPage);
+      setBotPage(botPage);
+    }
+  };
+
+  useLayoutEffect(() => {
+    if (viewport?.current && initialPage) {
+      const v = viewport.current!;
```
joelochlann, line 51: why the need for a non-null assertion operator?

frontend/src/js/components/PageViewer/VirtualScroll.tsx
```diff
@@ -0,0 +1,82 @@
...
+  //const [midPage, setMidPage] = useState(1); // Todo hook up to URL
+  const [botPage, setBotPage] = useState(1 + PRELOAD_PAGES);
+
+  const onScroll = () => {
+    if (viewport?.current) {
+      const v = viewport.current;
+
+      const currentMid = v.scrollTop + v.clientHeight / 2;
+
+      const topEdge = currentMid - PRELOAD_PAGES * pageHeight;
```
joelochlann, line 36: with all of these arithmetic expressions, I'd always prefer explicit brackets even when redundant because of BODMAS - I find it much easier to read brackets than recite BODMAS to myself and then try and mentally insert the brackets

frontend/src/js/components/PageViewer/PageCache.ts
```diff
@@ -0,0 +1,100 @@
...
+import { renderPdfPreview } from "./PdfHelpers";
+
+export class PageCache {
+  uri: string;
+
+  static MAX_CACHED_PAGES = 100;
+
+  // Tracking of text (Page}) and rendered previews (HTMLCanvasElement) is handled seperately incase an
+  // eviction happens to an item that is in flight
+  previews: CachedPreview[] = [];
```
joelochlann, line 12: why is this type called `CachedPreview` and the other `Page` (not `CachedPage`, or `CachedPageData`)?

I reckon they should be called `PagePreview` and `PageData` - there's nothing intrinsic in the type that means it's definitely cached, so could be misleading otherwise

frontend/src/js/components/PageViewer/PageCache.ts
```diff
@@ -0,0 +1,100 @@
+import authFetch from "../../util/auth/authFetch";
+import { CachedPreview, Page } from "./model";
+import { renderPdfPreview } from "./PdfHelpers";
+
+export class PageCache {
```
joelochlann, line 5: let's add some tests for this

frontend/src/js/components/PageViewer/PdfHelpers.ts
```diff
@@ -0,0 +1,71 @@
...
+  await pdfjs.renderTextLayer({
+    textContentStream,
+    container: document.createElement("div"),
+    viewport,
+    textDivs,
+    enhanceTextSelection: true,
+    textContentItemsStr,
+  }).promise;
+
+  return textDivs.flatMap((textDiv) => {
```
joelochlann, line 59: seems like this could just be a `map` returning an object?

joelochlann (approved):
Some naming things, but other than that: nice work.

Having had a play around with it, it feel very slick and the performance difference from the old viewer is palpable. In fact it's so quick loading in new pages that it's hard to tell the difference between cached & uncached.

Interestingly, from an anecdotal UX research perspective, in the previous approach I find the lag for new pages acceptable - seems to be about 200ms from the start of the first request to the response from the second. This may be because of the analogy with page turn speed of a physical document or the time it takes the human eye to apprehend new material.

But what does feel jarring is scrolling back to pages that were previously loaded and having to wait for them to reload. So for this particular user (me), it's the caching that makes the biggest difference rather than the reduction in first load latency.

Of course it's pretty cool that in your implementation it's hard to even tell them apart!

frontend/src/js/components/PageViewer/PageCache.ts
itsibitzi: `CachedPreview` contains information which is relevant to how the data is going to be accessed in practice by the systems that rely on the cache. Specifically the rendered canvas, the scale used, and the `PDFPageProxy`.

It's named like this because this collection of fields only makes sense in the context of working with cached results, where we want to minimise computation.

frontend/src/js/components/PageViewer/VirtualScroll.module.css
```diff
@@ -0,0 +1,24 @@
+.scrollArea {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
```
joelochlann, line 4: Don't think these have any effect because you have the `.pages` element intervening between the `.scrollArea` and the `.page` elements.

If you put

```css
  display: flex;
  flex-direction: column;
  justify-content: center;
  ```
on `.pages`, it will center each `.page` element which is nice

joelochlann:
Would also be good to center the pages horizontally before merging (see my CSS comments above)

frontend/src/js/components/PageViewer/PdfHelpers.ts
itsibitzi: Ah yeah, it was like that in the old version I copied across - will remove the flattening!

frontend/src/js/components/PageViewer/PageCache.ts
itsibitzi: `CachedPreview` is named like that because it contains data pertinent to consumers of the cache, in this case the rendered canvas (the "Preview" in this case), the scale, and the raw `PDFPageProxy`. 

Consumers of the cache need to consume these 3 elements but for different things, in no particularly coherent or abstract way so the type is a bit tricky to name... `CachedPreviewResult` is the only thing I can think of but that's not really adding anything... idk

itsibitzi:
Updated the caching logic to abort in-flight page requests when you're aggressively scrolling.

Here's a video where I scroll ~100 pages in a second or so. You can see that the render performance does take a hit but the browser survives.

https://user-images.githubusercontent.com/5560113/157227110-78a69077-577d-4619-a76c-a3a45cb93098.mp4

The performance hit seems to be effected by the fact we're fetching `pdf.worker.min.js` for every page. Maybe if we solve this in another PR we can improve the perf here a little bit, but it's really just a nice-to-have.

PR #36, opened 2022-03-16
Move giant to Java 11

philmcmahon:
## What does this change?
This was originally part of work to get giant onto graviton. I'm not sure we'll achieve that but might as well make use of this work. This moves giant onto Java 11, with the various errors I encountered fixed by the age old method of 'google copy paste'

Upgrade to scala 2.12.6: https://stackoverflow.com/questions/50034684/simulacrum-macro-implementation-not-found
`.lines` now returns a stream https://stackoverflow.com/questions/68864659/problem-rewriting-scala-method-post-upgrade-to-scala-2-12-from-2-11 (I used a different solution to convert from the stream to a list)

## How to test
This is currently live on playground https://playground.pfi.gutools.co.uk/workspaces/56b36105-498f-45a2-a7bb-ddf3a34c0c6c I've tested search and file upload

Newer better faster shinier onwards and upwards
ooooh

### DIFF
--- a/scripts/teamcity.sh
+++ b/scripts/teamcity.sh
@@ -25,7 +25,7 @@ export CI=true
 
 #Use java 11
 export JAVA_HOME=/usr/lib/jvm/java-11-amazon-corretto
-export PATH=$PATH:$JAVA_HOME
+PATH=$PATH:$JAVA_HOME
 echo $JAVA_HOME
 java -version
 
--- a/scripts/teamcity.sh
+++ b/scripts/teamcity.sh
@@ -4,9 +4,6 @@ set -e
 # Make Create React App treat warnings as errors
 export CI=true
 
-#Use java 11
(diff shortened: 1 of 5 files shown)


scripts/teamcity.sh
```diff
@@ -23,6 +23,10 @@ cp -r frontend/build/* backend/public
 # On Teamcity the JDeb build doesn't seem to follow the symbolic link while packaging, weirdly
 cp frontend/node_modules/pdfjs-dist/build/pdf.worker.min.js backend/public/third-party/pdf.worker.min.js
 
+#Use java 11
+export JAVA_HOME=/usr/lib/jvm/java-11-amazon-corretto
```
itsibitzi, line 27: I'm guessing this is implicitly defined by the AMI we expect Giant to be running on?

itsibitzi (approved):
Nice!

PR #38, opened 2022-03-17
Add "Find In Page" (aka Ctrl-F) to Page Viewer

itsibitzi:
As said in the title this PR adds the ability to make impromptu searches to the current document without having to go back to the main search page.

A lot of this work will be reused to do next/previous search highlights.

https://user-images.githubusercontent.com/5560113/158852217-de003fc5-41ef-4989-8b30-37b318e88759.mov

#### Searching
When a user enters an impromptu query a request is sent to the pages API which fetches the first 501 pages which match the query. This array of page numbers is stored in the client, a reference is kept to the last page the user pressed "next" to arrive at so that if they scroll around the next and previous buttons still make sense. 

I copied the behaviour of Firefox where if you've got hits on pages 10, 20, 25 and 30 and you're already on page 12 the initial "next" page will be 20. If you then scroll past 25 and press next it will go *back up* to 25. 

Each time you jump to a hit the next and previous 3 pages are prefetched using the existing page cache code which makes the experience more buttery.

It's worth noting that you can no longer jump between specific highlights. I initially implemented next/previous going page-by-page because it simplifies the code but I think it's reasonable that it should stay like this. Hammering next/previous over and over to jump through specific highlights doesn't add a lot of value. If people are eager to add this feature back then it should be fairly easy.

As I said the API only returns 501 pages to avoid an unbounded number of pages coming back which would be bad for performance if the user puts in a query with a lot of hits while on a large document. Here's the error message that shows, note it says only 500 will be shown which isn't exactly true since I return all 501 but I guess it's a small concession.

Capture

This could be fixed by implementing a paginated search system which somehow searches "around" the current middle page. I actually started with this implementation but decided the system but decided it would complicate this already chonky PR.

#### Highlights
Since we need to highlight impromptu results as well as the existing search highlights I had to make changes to how we get the highlights.

When you get a page's highlights you can now send an impromptu query (`iq`) along side the normal query (`q`). If we have both strings defined we issue both queries in Elasticsearch using a [multisearch query](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-multi-search.html). Highlights are pulled out as before, except there's now both regular hightligths and impromptu highlights.

Highlights are deduplicated across languages to minimise the number of fetches to the S3 page store before geometry is calculated.

The highlights are then passed to the frontend as before, with the impromptu ones `type` set appropriately so we can render them in a different colour.

At this point we need to bust the existing cached highlight geometry so we re-request the highlights for the "current" pages. The `PageCache` has been split into 2 caches so that geometry and PDF rendering can be handled separately. This is because we don't need to re-request and re-render the PDF just because the highlights changed. It also has the added benefit of reducing flicker when you change your impromptu query.

## Other thoughts

Wiring the impromptu search bar to the virtual scroller was a bit of a pain. I had to move the page cache into the virtual scroller for ergonomic reasons with the highlight refresh but this made page prefetching a bit ugly (you have to pass in an array of pages to prefetch).

Maybe it's time to break out the caching and prefetching logic into a redux store? We might be able to use more modern tools like RTK along side the older raw Redux stuff we have.

Additionally it might be useful to have some prioritisation on the cache busting, currently we refresh the geometry for all pages currently being displayed by the `VirtualScroll` component at the same time which isn't super spiffy.


joelochlann:
One naming thought, while it's in my head: what about "find in document" (or even just "find") instead of "impromptu search"? Aligns with desktop app/browser terminology and hence the F in Cmd+F

joelochlann:
Also, haven't looked at the code yet, but this is super exciting 🔎 📄 🎉 👏

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '3'
+dersion: '3'
```
philmcmahon, line 1: is this deliberate?

itsibitzi: oh. weird. you'd think that would have broken something lol?

joelochlann:
> It's worth noting that you can no longer jump between specific highlights. I initially implemented next/previous going page-by-page because it simplifies the code but I think it's reasonable that it should stay like this. Hammering next/previous over and over to jump through specific highlights doesn't add a lot of value. If people are eager to add this feature back then it should be fairly easy.

For me this is an important one. The model that people (including myself) have internalised from "find" in almost all applications (MacOS preview, browsers, Google Docs, word processors) navigates by individual result, not by page.

Giant itself in its existing text/OCR view follows this model, saying "Result x of y" and up/down buttons to navigate between them.

When playing around on the branch before reading the PR description, the numbers (e.g. `1/4`) in the search widget really confused me because I would never have guessed they referred to pages.

So I'd be a strong advocate of navigation and counters based on individual highlights. Doesn't have to be in this PR if it's a pain to implement but I'd want it before we unleashed on users

joelochlann:
An easily addressed (but, I think, significant) UI point: can we swap the up/down arrows? There was something subconsciously off about them and I've realised it's because everything else puts the "previous" arrow first:

Screenshot 2022-03-21 at 16 35 58

philmcmahon:
I've scanned through the code and happy with this, haven't looked at behaviour

itsibitzi:
Can certainly swap the up and down buttons around.

I'm not sure what value a close button adds if we're going to have the bar there permanently anyway? It just harms discoverability and means you have to figure out where to put it in the context of the bar. You can see in the Pinpoint example they seem to cover up the next/previous match buttons, which is pretty poor.

Can look at getting highlight level next/previous in but it may increase the size of this PR by a good bit.

frontend/src/js/components/PageViewer/VirtualScroll.tsx
```diff
@@ -131,7 +131,7 @@ export const VirtualScroll: FC<VirtualScrollProps> = ({
 
       setCurrentPages(newCurrentPages);
     }
-  }, [triggerHighlightRefresh, pageCache, currentPages]);
+  }, [triggerHighlightRefresh, pageCache]);
```
philmcmahon, line 137: I think you could wrap `setCurrentPages` in an `if (newCurrentPages !== currentPages)` then you'd avoid the infinite loop but there would still be an extra execution, so maybe fighting the liter is better

philmcmahon:
Had a scan through the latest commits, all looks good to me

joelochlann (approved):

PR #42, opened 2022-03-29
Allow admins to delete stuff

philmcmahon:
## What does this change?
This adds a delete button to the viewer page of giant. This button is wired up to the exiting deleteBlob endpoint, which has been modified to support the deletion of 'parent File' and 'workspace Node'. 

For context, in the UI this is a parent file:
Screenshot 2022-03-29 at 13 52 12

And this is a workspace node
Screenshot 2022-03-29 at 13 52 25

They kinda look the same. The idea is these should disappear when the blob itself is deleted because the filename may contain stuff we want deleted.

I spent a while trying to work out the most beautiful verison of the deletemodal thing - went back and forth between the current version with loads of conditionals and just having a component that takes some values. Decided it was ok as it is for now...

The feature is currently disabled for emails and any resources with child resources (e.g. zip files). That could be done as a a separate bit of work.

## How to test
I've tested this on playground - feel free to have a go as well or there's a video below showing deletion from search results.

We can easily delete individual items out of giant

If I've messed up the neo4j queries then who knows maybe this feature could unexpectedly delete all the content in giant 😱 

## Images

https://user-images.githubusercontent.com/3606555/160611691-c70025cd-0f2e-42c8-8bbf-f76c89b53ba2.mov

delet button gif

### DIFF
--- a/frontend/src/js/components/viewer/DeleteButtonModal.tsx
+++ b/frontend/src/js/components/viewer/DeleteButtonModal.tsx
@@ -12,9 +12,12 @@ export function DeleteButtonModal({ resource }: { resource: Resource | null }) {
     const [modalOpen, setModalOpen] = useState(false);
     const [deleteStatus, setDeleteStatus] = useState<DeleteStatus>("unconfirmed");
 
-    if (!resource || resource.children.length > 0) {
+    if (!resource) {
         return null;
     }
+
+    const hasChildResources = resource.children.length > 0
+    const tooltip = hasChildResources ? "Item cannot be deleted as it has child resources" : "Delete item"
 
     const deleteItem = async () => {
         try {
@@ -61,7 +64,7 @@ export function DeleteButtonModal({ resource }: { resource: Resource | null }) {
             </div>
         </Modal>
 
-        <button className="btn" onClick={() => setModalOpen(true)}>
+        <button className="btn" onClick={() => setModalOpen(true)} title={tooltip} disabled={hasChildResources}>
             Delete
         </button>
     </React.Fragment>


backend/conf/routes
```diff
@@ -12,7 +12,7 @@ POST          /api/collections/:collection/:ingestion/verifyFiles           cont
 
 GET           /api/blobs                                                    controllers.api.Blobs.getBlobs
 POST          /api/blobs/:id/reprocess                                      controllers.api.Blobs.reprocess(id, rerunSuccessful: Option[Boolean], rerunFailed: Option[Boolean])
-DELETE        /api/blobs/:id                                                controllers.api.Blobs.delete(id)
+DELETE        /api/blobs/:id                                                controllers.api.Blobs.delete(id, deleteFolders: Boolean, checkChildren: Boolean)
```
itsibitzi, line 15: It's not an issue with this PR since it was in the CLI before but I think we ought to think of a way of partitioning off CLI-only endpoints from the normal REST API.

backend/app/commands/DeleteResource.scala
```diff
@@ -0,0 +1,53 @@
...
+     }
+
+     // Deletes resource after checking it has no child nodes
+     def deleteBlobCheckChildren(id: String, deleteFolders: Boolean): Attempt[_ <: Unit] = {
+       val uri = Uri(id)
+
+       // casting to an option here because Attempt[Resource] and Attempt[Unit] are incompatible - so can't use a for comprehension with toAttempt
+       val deleteResult = manifest.getResource(uri).toOption map { resource =>
+         if (resource.children.isEmpty) deleteResource(uri, deleteFolders)
+         else Attempt.Left(DeleteFailure(s"Cannot delete $uri as it has child nodes - is it a zip file or an email?"))
```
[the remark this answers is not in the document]
itsibitzi: We could possibly just use an `IllegalStateFailure` here instead?

frontend/src/js/components/viewer/DeleteButtonModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+                    {deleteStatus === "deleted" ? "This item has been successfully deleted."  :
+                        "Are you sure you want to delete this item? This action will permanently delete the resource from all collections and datasets."}
+                </div>
+                <div className='form__row'>
+                    { deleteStatus === "deleted" ?
+                        <>
+                            <button className="btn" onClick={()=>document.location.href="/"}>
+                                Giant Home
+                            </button>
+                            {window.history.length > 1 &&  <button className="btn" onClick={() => window.history.back()}>Return to Search</button>}
```
itsibitzi, line 51: There are states where `window.history.back()` won't result in going to the search page. For example if you've gone to the "next document". 

You could possibly use `buildLink` to go back to the search page instead.

frontend/src/js/components/viewer/DocumentMetadata.js
```diff
@@ -104,7 +104,7 @@ export class DocumentMetadata extends React.Component {
                 <div className='sidebar__title'>
                     Actions
                 </div>
-                <ViewerActions resource={this.props.resource} config={this.props.config} />
+                <ViewerActions resource={this.props.resource} config={this.props.config} isAdmin={this.props.myPermissions.includes('CanPerformAdminOperations')} />
```
itsibitzi, line 107: We can check if there's child elements using the state we have in the redux store to hide this button if there are subresources.

itsibitzi (changes requested):

frontend/src/js/components/viewer/DocumentMetadata.js
philmcmahon: This is handled by https://github.com/guardian/giant/blob/pm-delete/frontend/src/js/components/viewer/DeleteButtonModal.tsx#L15

backend/app/services/manifest/Neo4jManifest.scala
```diff
@@ -947,28 +947,49 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
...
-      } else {
-        Attempt.Left(IllegalStateFailure(s"Error deleting blob $uri. Nodes deleted: ${counters.nodesDeleted()}"))
+      if (correctResultsCount(counters.nodesDeleted())) Attempt.Right(())
+      else {
+        Attempt.Left(IllegalStateFailure(s"$errorText of  blob $uri. Nodes deleted: ${counters.nodesDeleted()}"))
       }
     }
   }
 
+  def deleteBlobWorkspaceNode(uri: Uri): Attempt[Unit] = processDelete(
```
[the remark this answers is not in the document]
philmcmahon: discussed in person, agreed that current behaviour is inefficient (3 round trips) but ok for admin functionality

frontend/src/js/components/viewer/DeleteButtonModal.tsx
philmcmahon: thanks, have updated the text to make it more accurate

backend/conf/routes
philmcmahon: ok, sounds sensible for the future

itsibitzi (approved):
LGTM.

Minor improvement could be disabling the button and giving it hover text of something like "you can't delete this document because it has subresources". Not very important for admin functionality but I guess UX still matters - esp. if ESD folx are going to be administering Giant in the future?

PR #51, opened 2022-05-25
Fix Snyk workflow: Remove rogue package-lock.json file in root

philmcmahon:
## What does this change?
We had a rogue package-lock file which was causing snyk to think that there should be a package.json file in the root of the project, rather than where it lives in `frontend/`. Removing this file fixes the snyk job. 

I've also dropped the `SKIP_NODE=false` line, because that's the default setting


itsibitzi (approved):
Nerf rogues!

PR #52, opened 2022-05-30
Use same worker thread for all pages

joelochlann:
If we don't pass a `worker` to `pdfjs.getDocument()`, pdf.js creates a new worker on every call:

https://github.com/mozilla/pdf.js/blob/5cf116a958548f6596674bf8d5ca0fe64aa2df3c/src/display/api.js#L331-L342

This causes a request to fetch `third-party/pdf.worker.min.js` and a new web worker thread per page loaded. When dragging across many pages, this thread leak crashes the browser when it hits about 30 or 40 threads.

Instead, we can create a worker (via pdf.js's wrapper) when we instantiate the page cache, and pass this in to `getDocument`

### DIFF
--- a/frontend/src/js/components/PageViewer/PageCache.ts
+++ b/frontend/src/js/components/PageViewer/PageCache.ts
@@ -2,6 +2,7 @@ import authFetch from "../../util/auth/authFetch";
 import { LruCache } from "../../util/LruCache";
 import { CachedPreview, PageData } from "./model";
 import { renderPdfPreview } from "./PdfHelpers";
+import * as pdfjs from 'pdfjs-dist';
 
 export type CachedPage = {
   previewAbortController: AbortController;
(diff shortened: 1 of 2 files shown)


joelochlann:
[Fix thread leak when rendering PDFs](https://trello.com/c/ADoiIyjd/1981-fix-thread-leak-when-rendering-pdfs)

itsibitzi (approved):
Nice! 

Was this API actually documented anywhere or did you just dig through the code?

joelochlann:
dug through code! I haven't really found any useful pdf.js docs anywhere: https://mozilla.github.io/pdf.js/

joelochlann:
I've always found with pdf.js the only way to figure out how to use it properly is to read the source code.

Even their TypeScript definitions are weak, e.g. this `PDFWorker` param for `getDocument` is typed `any` so you need to read every line to find out what to pass in

PR #53, opened 2022-05-31
Add rotate buttons to pdf viewer

philmcmahon:
## What does this change?
This adds rotate clockwise/anticlockwise buttons to giant. It's pretty nasty dom manipulation, but is what's recommended in the pdfjs docs https://pdfjs.express/blog/how-to-build-pdf-viewer-jquery-pdfjs

The rotate buttons normally live in the secondary menu, which we hide in order to get rid of the download button. It might be worth changing our approach there and just hiding download rather than the entire right hand menu. I think even if we do that though it's a good idea to promote the rotate buttons to the main toolbar so they are easy to find.

Screenshot 2022-05-31 at 16 19 15

### DIFF
--- a/frontend/src/js/components/viewer/EmbeddedPdfViewer.tsx
+++ b/frontend/src/js/components/viewer/EmbeddedPdfViewer.tsx
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types';
 
 const viewerLocation = '/third-party/pdfjs-2.4.456-dist/web/viewer.html';
 
+
+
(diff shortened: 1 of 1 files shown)


joelochlann (approved):
Wow! Gotta love pdfjs's API and docs 🤯

joelochlann:
class `toolbarButton` helps with correct top/bottom margins though you can't get rid of `secondaryToolbarButton` because it sets padding and stuff (otherwise the element has zero width because it contains an absolutely positioned icon)

PR #54, opened 2022-05-31
Do not setPagesToLoad twice on init

joelochlann:
We set initial state values for `botPage`, `midPage`, `topPage`, but then we have a `useEffect()` which re-computes them from the viewport on init.

This causes it to unnecessarily unmount the third page on every load, which exposes an underlying problem with promise callbacks firing for pages after they've been unmounted.

The promise [cleanup problem](https://reactjs.org/docs/hooks-effect.html#effects-with-cleanup) needs a separate solution (think we may need to wrap those promises into [cancellable promises](https://dev.to/praveenkumarrr/cancellable-promises-in-react-and-why-is-it-required-5ghf)) but this stops us pointlessly mounting then unmounting the third page on every fresh page load.

# before
Screenshot 2022-05-31 at 16 27 11

# after
Screenshot 2022-05-31 at 16 27 42


philmcmahon (approved):
Typescript LGTM

PR #56, opened 2022-06-16
Navigate between individual highlights

joelochlann:
Since our existing text view (as well as Google Docs) allows navigation between individual highlights, I thought it would be good to have it behave this way in the new page viewer.

The backend now gets the page hits (which is fast), and then for each page hit parses the highlights (which is slower).

So for searches with hits on lots of page, performance is noticeably slower. But more specific searches return quickly.

In Playground searching Functional Programming in Scala:
- "the" takes 10 seconds
- "function" takes 6 seconds
- "monoidal" takes 366 milliseconds

## Other changes
- Spinner while find is pending
- Do not quote the query when getting find hits, i.e. allow stemming by default and let the user put quotes if they don't want stemming. (Beforehand we were quoting when getting the page hits, but not when rendering highlights, leading to a mismatch between visible highlights and the navigation)

## Notes on stemming/tokenisation
- Because we use elasticsearch to do the find operation, it searches for occurrences of word stems by default
  - i.e. `functional` will match `function`, `functions`, `functional`, etc
- You can search for exact matches by quoting
  - i.e. `"functional"` will only match `functional`
- However you cannot search within tokens/words regardless of quotes
  - i.e. neither `"sc"` nor `sc` will match `scala`, but `s` will match usages of `s` as a variable name in code examples
- This **differs** from typical find functionality (Google Docs, Preview, browser Cmd+F etc), but is **consistent** with Giant's primary search functionality
- There is a [detailed ADR](https://github.com/guardian/pfi/blob/1dca05c24b24e79322d7e00dbe7da2c723b92d24/docs/adrs/2020-09-28_quoted-searches.md) about our approach to quoted searches in Giant more generally

# before

https://user-images.githubusercontent.com/5122968/174096748-7ec8d1be-3724-47e7-a2cf-fc8398faf2e6.mov

# after

https://user-images.githubusercontent.com/5122968/174096836-ee85d5de-5c61-4afb-9db1-0afc6241ebdb.mov

### DIFF
--- a/backend/app/controllers/api/PagesController.scala
+++ b/backend/app/controllers/api/PagesController.scala
@@ -124,12 +124,15 @@ class PagesController(val controllerComponents: AuthControllerComponents, manife
       pagesWithHits <- pagesService.findInPages(uri, findQuery)
       pageData <- Attempt.sequence(pagesWithHits.map(pageData(uri, _, req.user.username, None, Some(findQuery))))
     } yield {
-      val highlights = pageData.flatMap { page =>
-        page.highlights.map {
-          highlight => HighlightForSearchNavigation.fromPageHighlight(page.page, highlight.index, highlight)
-        }
+      val highlights = for {
+        page <- pageData
+        highlight <- page.highlights
+      } yield {
+        HighlightForSearchNavigation.fromPageHighlight(page.page, highlight.index, highlight)
       }
-      Ok(Json.toJson(highlights.sortBy(h => (h.pageNumber, h.highlightNumber))))
+
+      val sortedHighlights = highlights.sortBy(h => (h.pageNumber, h.highlightNumber))
+      Ok(Json.toJson(sortedHighlights))
     }
   }
 }
--- a/backend/app/controllers/api/PagesController.scala
+++ b/backend/app/controllers/api/PagesController.scala
@@ -5,7 +5,7 @@ import commands.GetPages.PagePreviewMetadata
 import java.io.InputStream
 import commands.{GetPagePreview, GetPages, GetResource, ResourceFetchMode}
 import model.frontend.{Chips, HighlightableText, TextHighlight}
-import model.index.{FrontendPage, Page, PageHighlight, PageWithFind}
+import model.index.{FindHighlight, FrontendPage, HighlightForSearchNavigation, Page, PageHighlight, PageWithFind}
 import model.{Language, Languages, Uri}
 import org.apache.pdfbox.pdmodel.PDDocument
 import play.api.libs.json.Json
@@ -124,7 +124,12 @@ class PagesController(val controllerComponents: AuthControllerComponents, manife
       pagesWithHits <- pagesService.findInPages(uri, findQuery)
       pageData <- Attempt.sequence(pagesWithHits.map(pageData(uri, _, req.user.username, None, Some(findQuery))))
     } yield {
-      Ok(Json.toJson(pageData))
+      val highlights = pageData.flatMap { page =>
+        page.highlights.map {
+          highlight => HighlightForSearchNavigation.fromPageHighlight(page.page, highlight.index, highlight)
+        }
+      }
+      Ok(Json.toJson(highlights.sortBy(h => (h.pageNumber, h.highlightNumber))))
     }
   }
 }
--- a/backend/app/controllers/api/PagesController.scala
+++ b/backend/app/controllers/api/PagesController.scala
@@ -27,7 +27,7 @@ class PagesController(val controllerComponents: AuthControllerComponents, manife
   }
 
   // Get language and highlight data for a given page
-  private def pageData(uri: Uri, pageNumber: Int, username: String, sq: Option[String], fq: Option[String]): Attempt[FrontendPage] = {
+  private def frontendPageFromQuery(uri: Uri, pageNumber: Int, username: String, sq: Option[String], fq: Option[String]): Attempt[FrontendPage] = {
     // Across documents
     val searchQuery = sq.map(Chips.parseQueryString)
     // Within document
@@ -54,8 +54,11 @@ class PagesController(val controllerComponents: AuthControllerComponents, manife
 
   // Get language and highlight data for a given page
   def getPageData(uri: Uri, pageNumber: Int, sq: Option[String], fq: Option[String]) = ApiAction.attempt { req =>
-    val response = pageData(uri, pageNumber, req.user.username, sq, fq)
-    response.map { r => Ok(Json.toJson(r))}
+    for {
+      response <- frontendPageFromQuery(uri, pageNumber, req.user.username, sq, fq)
+    } yield {
+      Ok(Json.toJson(response))
+    }
   }
 
   case class HighlightGeometries(lang: Language, highlights: List[PageHighlight])
@@ -122,7 +125,9 @@ class PagesController(val controllerComponents: AuthControllerComponents, manife
 
     for {
       pagesWithHits <- pagesService.findInPages(uri, findQuery)
-      pageData <- Attempt.sequence(pagesWithHits.map(pageData(uri, _, req.user.username, None, Some(findQuery))))
+      pageData <- Attempt.sequence(
+        pagesWithHits.map(frontendPageFromQuery(uri, _, req.user.username, None, Some(findQuery)))
+      )
     } yield {
       val highlights = for {
         page <- pageData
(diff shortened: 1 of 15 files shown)


frontend/src/js/components/PageViewer/VirtualScroll.tsx
```diff
@@ -101,27 +103,39 @@ export const VirtualScroll: FC<VirtualScrollProps> = ({
...
     }
-  }, [pageHeight, jumpToPage]);
+  }, [pageHeight, jumpToPage, jumpToScrollPosition]);
+
+  useLayoutEffect(() => {
+    if (viewport?.current && focusedFindHighlight) {
+      const v = viewport.current;
+      const topOfHighlightPage = (pageHeight * (focusedFindHighlight.pageNumber - 1));
+      // const highlight = topOfHighlightPage + (focusedFindHighlight.firstSpan?.y ?? 0);
+      // v.scrollTop = highlight + (v.clientHeight / 2);
```
philmcmahon, line 129: tidy this up?  `v` also not an ideal variable name (though short lived!)

frontend/src/js/components/PageViewer/VirtualScroll.tsx
```diff
@@ -101,27 +103,39 @@ export const VirtualScroll: FC<VirtualScrollProps> = ({
...
+    if (viewport?.current) {
       const v = viewport.current;
-      const scrollTo = (jumpToPage - 1) * pageHeight;
-      v.scrollTop = scrollTo;
+      if (jumpToPage) {
+        // TODO: page jump should go to middle?
+        const scrollTo = (jumpToPage - 1) * pageHeight;
+        v.scrollTop = scrollTo;
+      }
+      if (jumpToScrollPosition) {
```
joelochlann, line 117: didn't use `jumpToScrollPosition` in the end but have left it in

frontend/src/js/components/PageViewer/FindInput.tsx
```diff
@@ -1,42 +1,47 @@
...
 type FindInputProps = {
   value: string;
   setValue: (v: string) => void;
   performFind: (query: string) => Promise<void>;
+  isPending: boolean;
   jumpToNextFindHit: () => void;
   jumpToPreviousFindHit: () => void;
-  hits: number[];
-  lastPageHit: number;
+  // TODO: could be null?
```
philmcmahon, line 22: is this simple enough to check before merging?

philmcmahon (changes requested):

frontend/src/js/components/PageViewer/VirtualScroll.tsx
joelochlann: yeah this was the code that was supposed to jump to a point where the highlight was in the middle of the page (instead of just to the top of that page), but it didn't work because I need to multiply by the page `scale` value which I don't have here

have deleted

unfortunately can't call `v` `viewport` because that name is already bound, but I can just keep referring to `viewport.current` if clearer

frontend/src/js/components/PageViewer/FindInput.tsx
joelochlann: it can't be null actually, it's state that initialised to `[]` in `PageViewer.tsx`. will delete comment

frontend/src/js/components/PageViewer/FindInput.tsx
```diff
@@ -1,42 +1,46 @@
...
-  useEffect,
+  useEffect, useMemo,
   useState,
-} from "react";
+} from 'react';
 import DownIcon from "react-icons/lib/md/arrow-downward";
 import UpIcon from "react-icons/lib/md/arrow-upward";
 import styles from "./FindInput.module.css";
+import { HighlightForSearchNavigation } from './model';
+import { Loader } from 'semantic-ui-react';
```
joelochlann, line 13: I don't love the dependency on Semantic UI but it was quickest way of getting a decent loader icon. We could switch to our own CSS in future PR. I've put this in checklist here: https://trello.com/c/ycfzJavQ/1986-page-viewer-navigate-between-individual-highlights

frontend/src/js/components/PageViewer/PageViewer.tsx
```diff
@@ -68,65 +65,77 @@ export const PageViewer: FC<PageViewerProps> = () => {
...
+      params.set("fq", removeLastUnmatchedQuote(query));
+
+      setIsFindPending(true);
+      return authFetch(`/api/pages2/${uri}/find?${params.toString()}`)
         .then((res) => res.json())
-        .then((searchHits) => {
-          setLastPageHit(middlePage);
-          setFindHits(searchHits);
+        .then((highlights) => {
+          setIsFindPending(false);
```
joelochlann, line 78: would be nice not to do a round-trip to server when emptying search box, but didn't seem super important for this PR. have added to https://trello.com/c/ycfzJavQ/1986-page-viewer-navigate-between-individual-highlights

backend/app/controllers/api/PagesController.scala
```diff
@@ -112,10 +120,24 @@ class PagesController(val controllerComponents: AuthControllerComponents, manife
     }
   }
 
-  def findInDocument(uri: Uri, fq: String) = ApiAction.attempt {
+  def findInDocument(uri: Uri, fq: String) = ApiAction.attempt { req =>
```
itsibitzi, line 123: As discussed IRL it would be neat if we could use [SSE](https://www.w3schools.com/html/html5_serversentevents.asp) here.

backend/app/model/index/Page.scala
```diff
@@ -18,16 +18,22 @@ object PageDimensions {
...
 
-sealed abstract class PageHighlight { def id: String }
-case class SearchHighlight(id: String, spans: List[HighlightSpan]) extends PageHighlight
-case class FindHighlight(id: String, spans: List[HighlightSpan]) extends PageHighlight
+sealed abstract class PageHighlight {
+  def id: String
+  def index: Int
+  def spans: List[HighlightSpan]
+}
+// TODO: these are identical... why do we need them?
```
itsibitzi, line 26: I think the idea was the `type` that gets inserted when serialized can be used to style them differently. This type existed before I did my work on `viewer2` so I'm not actually sure :)

frontend/src/js/components/PageViewer/FindInput.tsx
```diff
@@ -1,42 +1,46 @@
...
   performFind,
-  hits,
-  lastPageHit,
+  isPending,
+  highlights,
+  focusedFindHighlightIndex,
 }) => {
   const [showWarning, setShowWarning] = useState(false);
 
-  const debouncedPerformSearch = useCallback(_.debounce(performFind, 300), [
```
itsibitzi, line 39: Any reason this was changed to `useMemo`? The behavior should be identical but `useCallback` is more idiomatic, though personally I'm not very fussed about idiomatic details like this.

frontend/src/js/util/stringUtils.ts
```diff
@@ -1,3 +1,11 @@
 export function getLastPart(input: string, separator: string) {
-    return input.split(separator).slice(-1)[0];
+  return input.split(separator).slice(-1)[0];
+}
+
+export function removeLastUnmatchedQuote(input: string) {
```
itsibitzi, line 5: This is a good idea more generally. There are a few queries that can knock over our backend which should be pretty easy to detect. Anther example is `/` is considered an "begin regex" operator which is a bit mind blowning if you're a user I think.

frontend/src/js/components/PageViewer/VirtualScroll.tsx
itsibitzi: Shall we get rid? I think just viewing pages at a time is probably ok?

backend/app/controllers/api/PagesController.scala
joelochlann: Yeah this would be really cool! I think I'll press forward wiring up the main search highlights first and come back to this one, since page viewer work has dragged on so long

frontend/src/js/components/PageViewer/FindInput.tsx
joelochlann: Technically with the `useCallback` approach `_.debounce` gets called again each render because of the evaluation order: it has to evaluate the inner function call to `debounce` before it passes result to `useCallback`. Though `useCallback` won't use the result unless the deps have changed. 

Whereas with `useMemo` you can ensure that if you're calling a higher-order function you only re-call it when necessary.

This is a micro-optimisation though really as `debounce` is not an expensive operation. What do you prefer?

backend/app/controllers/api/PagesController.scala
joelochlann: added to this checklist: https://trello.com/c/ycfzJavQ/1986-page-viewer-navigate-between-individual-highlights

philmcmahon (approved):
Thanks for explaining/fixing the minor things I pointed out!

PR #57, opened 2022-06-21
Move highlight state to <Controls /> component

joelochlann:
There is a lot of state associated with highlights and navigating between them.

In order to not duplicate it all when I add "search in document" highlighting, I've moved as much as possible into `<Controls />` so that when I add the search controls I'll get a lot of functionality for free and properly encapsulated.

On refactor, the complex chain of state changes culminating in `triggerHighlightRefresh` was causing bugs, so I made the following changes:
- Remove `triggerHighlightRefresh` and refresh highlights when `findQuery` changes
- Group `pageCache.setFindQuery` and the subsequent re-rendering of pages together in the above effect to reduce chance that these end up out of sync
- Do not go to the server when `findQuery` is emptied, but rather wipe the highlights from cached pages. This has the effect that on initial page load (when `findQuery` is empty) we do not hit the server and cause an unnecessary re-render, which was causing some weird issues. (Previously the `triggerHighightRefresh > 0` check prevented this)

### DIFF
--- a/frontend/src/js/components/PageViewer/Controls.tsx
+++ b/frontend/src/js/components/PageViewer/Controls.tsx
@@ -42,7 +42,7 @@ export const Controls: FC<ControlsProps> = ({
 
 
   const performFind = useCallback((query: string) => {
-    const params = new URLSearchParams();
+   const params = new URLSearchParams();
     // The backend will respect quotes and do an exact search,
     // but if quotes are unbalanced elasticsearch will error
     // TODO: change to "q"
--- a/frontend/src/js/components/PageViewer/Controls.tsx
+++ b/frontend/src/js/components/PageViewer/Controls.tsx
@@ -66,7 +66,7 @@ export const Controls: FC<ControlsProps> = ({
           setFocusedFindHighlightIndex(null);
         }
       })
-  },   [uri]);
+  },   [uri, onQueryChange]);
 
 
   const jumpToNextFindHit = useCallback(() => {
--- a/frontend/src/js/components/PageViewer/Controls.tsx
+++ b/frontend/src/js/components/PageViewer/Controls.tsx
@@ -23,7 +23,7 @@ export const Controls: FC<ControlsProps> = ({
   rotateAnticlockwise,
   uri,
   onHighlightStateChange,
-    onQueryChange
+  onQueryChange
 }) => {
   const [focusedFindHighlightIndex, setFocusedFindHighlightIndex] = useState<number | null>(null);
   const [findHighlights, setFindHighlights] = useState<HighlightForSearchNavigation[]>([]);
@@ -45,6 +45,7 @@ export const Controls: FC<ControlsProps> = ({
     if (!query) {
       setFocusedFindHighlightIndex(null);
       setFindHighlights([]);
+      onQueryChange(query);
       return;
     }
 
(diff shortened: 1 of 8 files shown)


frontend/src/js/components/PageViewer/Controls.tsx
```diff
@@ -1,40 +1,117 @@
...
+
+    const params = new URLSearchParams();
+    // The backend will respect quotes and do an exact search,
+    // but if quotes are unbalanced elasticsearch will error
+    params.set("q", removeLastUnmatchedQuote(query));
+
+    // In order to use same debounce on communicating query change to parent
+    onQueryChange(query);
+    setIsFindPending(true);
+    return authFetch(`/api/pages2/${uri}/find?${params.toString()}`)
```
philmcmahon, line 58: what happens if this promise fails? do we care?

frontend/src/js/components/PageViewer/Controls.tsx
```diff
@@ -1,40 +1,117 @@
...
+      const previousHighlightIndex = (focusedFindHighlightIndex !== null && focusedFindHighlightIndex > 0)
+          ? (focusedFindHighlightIndex - 1)
+          : (findHighlights.length - 1);
+
+      setFocusedFindHighlightIndex(previousHighlightIndex);
+    }
+  }, [findHighlights, focusedFindHighlightIndex, setFocusedFindHighlightIndex]);
+
+  const handleUserKeyPress = useCallback((e) => {
+    if ((e.ctrlKey || e.metaKey) && e.keyCode === 70) {
```
philmcmahon, line 93: maybe a comment indicating what `70` is?

frontend/src/js/components/PageViewer/PageViewer.tsx
```diff
@@ -40,126 +46,52 @@ export const PageViewer: FC<PageViewerProps> = () => {
...
-          ? (focusedFindHighlightIndex + 1)
-          : 0;
-
-      preloadNextPreviousFindPages(nextHighlightIndex, findHighlights);
-      setFocusedFindHighlightIndex(nextHighlightIndex);
+    if (findHighlightsState.focusedIndex !== null && findHighlightsState.highlights.length) {
+      const length = findHighlightsState.highlights.length;
+
+      const indexesOfHighlightsToPreload = uniq(
+          range(-3, 3).map((offset) => {
```
philmcmahon, line 54: maybe comment explaining what -3 and 3 represent, it's pretty obvious though so not totally necessary (I assume the number of highlights in each direction to preload)

frontend/src/js/components/PageViewer/Controls.tsx
```diff
@@ -1,40 +1,117 @@
...
+
+    // In order to use same debounce on communicating query change to parent
+    onQueryChange(query);
+    setIsFindPending(true);
+    return authFetch(`/api/pages2/${uri}/find?${params.toString()}`)
+      .then((res) => res.json())
+      .then((highlights) => {
+        setIsFindPending(false);
+        setFindHighlights(highlights);
+        if (highlights.length) {
```
philmcmahon, line 63: maybe length !== 0 in case it ends up being undefined e.g. if json is not an array

philmcmahon (approved):
I haven't run this, but went through the typescript trying to nitpick stuff. Feel free to ignore!

frontend/src/js/components/PageViewer/Controls.tsx
joelochlann: we do care... we should probably display an error thingy in the search box. our approach to handling server errors in Giant is a bit inconsistent though - I'll put a TODO and for the time being we can let the console error bubble up etc. I'll address in next PR

PR #58, opened 2022-06-30
CLI delete ingestion improvements

philmcmahon:
## What does this change?
Prior to this PR, the delete-ingest CLI command only worked where the blobs in an ingestion didn't exist in any other ingestions, or you specified every single ingestion URI where the files included in an ingestion are included.

This was a problem as there are lots of files that are likely to exist all over the place, for example `.DS_Store` files created by macOS. 

This introduces a new `--conflictBehaviour` parameter allowing cli users to specify the behaviour in the scenario where a file exists in more than one ingestion. The options are:

 - stop: the previous behaviour - throw an exception explaining what's gone wrong, including sharing the names of the other places the file exists
 - skip: skip the file but carry on deleting everything else in the ingest
 - delete: delete the file anyway (this deletes the file from ALL locations in giant - so e.g. EVERY .DS_Store file)

## How to test

I've tested this locally by creating/deleting data from playground

Easier ingestion deletion in future!

### DIFF
--- a/cli/src/main/scala/com/gu/pfi/cli/service/CliIngestionService.scala
+++ b/cli/src/main/scala/com/gu/pfi/cli/service/CliIngestionService.scala
@@ -72,7 +72,8 @@ class CliIngestionService(http: CliHttpClient)(implicit ec: ExecutionContext) ex
   }
 
   def deleteBlob(id: String): Attempt[Unit] = {
-    http.delete(s"/api/blobs/$id?deleteFolders=false&checkChildren=false").map { r =>
+
+    http.delete(s"/api/blobs/${URLEncoder.encode(id, "UTF-8")}?deleteFolders=false&checkChildren=false").map { r =>
       if(r.code() == 204) {
         Attempt.Right(())
       } else {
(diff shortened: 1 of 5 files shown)


cli/src/main/scala/com/gu/pfi/cli/service/CliIngestionService.scala
```diff
@@ -72,7 +72,8 @@ class CliIngestionService(http: CliHttpClient)(implicit ec: ExecutionContext) ex
   }
 
   def deleteBlob(id: String): Attempt[Unit] = {
-    http.delete(s"/api/blobs/$id?deleteFolders=false&checkChildren=false").map { r =>
+
+    http.delete(s"/api/blobs/${URLEncoder.encode(id, "UTF-8")}?deleteFolders=false&checkChildren=false").map { r =>
```
joelochlann, line 76: Do we definitely want `deleteFolders=false`? If so could we have a comment explaining why, or tweak the naming so it's clearer what it means?

If there was also a way to clarify the `checkChildren` terminology in some way that would be a bonus... `skipIfParent`? `skipIfHasChildren`? Or, `orphanChildren=true`?

In fact the more I think about it the more confused I am... what's the behaviour if we delete at zip file with `checkChildren=false`? It deletes the zip blob, but leaves its contents intact in neo4j and elasticsearch? ditto email archives like .pst

At this point I start re-reading https://docs.google.com/document/d/1obN21_zUF0iNKyLO60XGfvBtx2B2zaFxX1C--uhp2bA/edit# and my head explodes a bit 🤯

PR #60, opened 2022-07-04
Wire up new page viewer

joelochlann:
This makes the new page viewer the default.

It will use the old viewer if the document has no pages (typically this means it was first uploaded before we had implemented the page extractor on the backend)

There is a "view as text" button which opens in a new tab:

Screenshot 2022-07-06 at 13 16 37

## Open questions
- Do we still need a feature switch? Should this be a user-level, or should it allow us to unilaterally switch page viewer on/off if we discover issues?
  - *Answer: just have an escape hatch button "view as text"*
- Do we care about comments? Or can we accept that the page viewer will disable commenting (and viewing comments) on docs with pages, until we implement comment logic for the page viewer? (See usage stats in [Trello](https://trello.com/c/wpIOExSS/2014-default-to-page-viewer-but-have-text-view-available-for-comments-and-non-page-viewable-docs))
  - *Answer: let's leave comments for now*

### DIFF
--- a/frontend/src/js/components/PageViewerOrFallback.tsx
+++ b/frontend/src/js/components/PageViewerOrFallback.tsx
@@ -5,9 +5,7 @@ import Viewer from './viewer/Viewer';
 import { PageViewer } from './PageViewer/PageViewer';
 import React from 'react';
 
-type PageViewerOrFallback = {}
-
-export const PageViewerOrFallback: FC<PageViewerOrFallback> = () => {
+export const PageViewerOrFallback: FC<{}> = () => {
     const { uri } = useParams<{ uri: string }>();
 
     const [totalPages, setTotalPages] = useState<number | null>(null);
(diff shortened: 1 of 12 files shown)


frontend/src/js/components/PageViewerOrFallback.tsx
```diff
@@ -0,0 +1,28 @@
+import { FC, useEffect, useState } from 'react';
+import authFetch from '../util/auth/authFetch';
+import { useParams } from 'react-router-dom';
+import Viewer from './viewer/Viewer';
+import { PageViewer } from './PageViewer/PageViewer';
+import React from 'react';
+
+type PageViewerOrFallback = {}
```
itsibitzi, line 8: Is this left over from something?

itsibitzi (approved):

frontend/src/js/components/PageViewerOrFallback.tsx
joelochlann: ah guess we don't need it, it doesn't accept any params in the end!

hoyla:
I think 

1. We should sacrifice comments in the short term because the benefits of page viewer outweigh the temporary inconvenience of no comments 
2. We should add a button for users "View as Text" which switches the URL to the old viewer mode, with the text view active.

PR #63, opened 2022-08-11
New brew dependency, upgrade node, add asdf file

joelochlann:
Some stuff that came out of trying to run Giant on my new laptop

- Adds `.tool-versions` for [asdf](https://asdf-vm.com/guide/getting-started.html) which is the new departmental recommendation
- Upgrades node 12 => 16, because node 12 does not have arm64 support
- Adds tesseract and libreoffice as brew deps
- npm and brew lockfile junk


joelochlann:
Current problem is the npm upgrade has exposed a peer dependency conflict between our version of React and the one specified by EUI. We hit this problem in Lurch and fixed by upgrading EUI. Would rather not do that here given that EUI is just for an experimental unused frontend...

Brewfile
```diff
@@ -0,0 +1,2 @@
+brew "tesseract"
+cask "libreoffice"
```
philmcmahon, line 2:
```suggestion
cask "libreoffice"
brew "wkhtmltopdf"
```

Following https://github.com/guardian/giant/pull/65 we'll need this too

joelochlann: nice spot

joelochlann:
Having a look at the peer dependency issue now... will see if I can align the EUI and React versions so it's OK (until we upgrade either of them)

joelochlann:
Think I've solved the peer dependency issue, though at the cost of a slight downgrade to TypeScript and React

joelochlann:
Seems to typecheck, just deploying to Playground to double-check

joelochlann:
Looks good. A better long-term solution would be to use npm workspaces to separate out the new UI, so its React version wouldn't be coupled to the main app. This would also mean no more "refresh to update styles" on the new UI, and no more increasing our bundle size to serve two separate UIs.

philmcmahon (approved):

philmcmahon:
Nice one Joe! Future giant devs owe you

PR #11, opened 2021-07-28
Always close s3 client in putObjectSync

philmcmahon:
## What does this change?
Following from https://github.com/guardian/pfi/pull/924 and a recent error on PROD when trying to upload a file to giant, this makes sure that post putObject the input stream for the object gets closed.

UPDATE: Sadly, this function is never actually called so can't be responsible for the issue. Probably worth fixing though in case we use it in future

Here's the stacktrace from earlier today

```
AWS Sdk Failure: com.amazonaws.SdkClientException: Unable to execute HTTP request: Timeout waiting for connection from pool
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.handleRetryableException(AmazonHttpClient.java:1175)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeHelper(AmazonHttpClient.java:1121)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.doExecute(AmazonHttpClient.java:770)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.executeWithTimer(AmazonHttpClient.java:744)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.execute(AmazonHttpClient.java:726)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutor.access$500(AmazonHttpClient.java:686)
	at com.amazonaws.http.AmazonHttpClient$RequestExecutionBuilderImpl.execute(AmazonHttpClient.java:668)
	at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:532)
	at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:512)
	at com.amazonaws.services.s3.AmazonS3Client.invoke(AmazonS3Client.java:4920)
	at com.amazonaws.services.s3.AmazonS3Client.invoke(AmazonS3Client.java:4866)
	at com.amazonaws.services.s3.AmazonS3Client.access$300(AmazonS3Client.java:389)
	at com.amazonaws.services.s3.AmazonS3Client$PutObjectStrategy.invokeServiceCall(AmazonS3Client.java:5800)
	at com.amazonaws.services.s3.AmazonS3Client.uploadObject(AmazonS3Client.java:1789)
	at com.amazonaws.services.s3.AmazonS3Client.putObject(AmazonS3Client.java:1749)
	at com.amazonaws.services.s3.transfer.internal.UploadCallable.uploadInOneChunk(UploadCallable.java:131)
[...]

PR #64, opened 2022-08-11
Send AWS SDK metrics to CloudWatch

joelochlann:
# The problem
In the past we've had Giant incidents caused by exhaustion of the S3 connection pool.

We've tried to mitigate by closing input streams (https://github.com/guardian/pfi/pull/924, #11) but they still crop up sometimes during large ingestions when the workers have lots of connections to S3.

The new page viewer (#32) has made this more common because it also makes lots of requests to S3 for individual pages.

# The solution
We want to get visibility on when we're in danger of hitting this issue. AWS provides extra metrics from its Java SDK, including around connection pools to AWS services such as S3, but you need to do some special stuff to enable it.

You can apparently do it by calling some methods on `AwsSdkMetrics`:
https://aws.amazon.com/blogs/developer/tuning-the-aws-sdk-for-java-to-improve-resiliency/

Or by setting some JVM options:
https://aws.amazon.com/blogs/developer/enabling-metrics-with-the-aws-sdk-for-java/

What neither of those tell you is **you need an extra dependency for it to work**:

```scala
"com.amazonaws" % "aws-java-sdk-cloudwatchmetrics" % awsVersion
```

# How they appear in CloudWatch
By default, they appear under custom namespaces formed from the stack/app/stage tags (no idea how AWS knows by default to use these... thought they were a Guardian-specific convention?)

If this is unhelpful I think we can customise. I haven't really played with CloudWatch metrics enough to know what makes the most sense

## top-level
Screzenshot 2022-08-11 at 12 50 59

## > pfi-playground-rex
Some useful JVM stuff in these other categories, e.g. heap space, number of threads, etc
Screenshot 2022-08-11 at 12 51 06

## > pfi-playground-rex > Metric Type
for some reason the connection pool metrics come under this generic "Metric Type" heading
Screenshot 2022-08-11 at 12 51 33

### DIFF
--- a/backend/app/AppComponents.scala
+++ b/backend/app/AppComponents.scala
@@ -2,10 +2,10 @@ import java.net.InetAddress
 import java.nio.file.Paths
 import java.security.Security
 import java.time.Clock
-
 import akka.actor.CoordinatedShutdown
 import akka.actor.CoordinatedShutdown.Reason
 import cats.syntax.either._
+import com.amazonaws.metrics.AwsSdkMetrics
 import com.gu.pandomainauth
 import com.gu.pandomainauth.PublicSettings
 import controllers.AssetsComponents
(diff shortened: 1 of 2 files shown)


joelochlann:
## todo
- [ ] tell AWS about the missing bit in their docs (the `cloudwatchmetrics` dependency)
- [ ] create a departmental recommendation

philmcmahon:
Nice one Joe!

With my eternal cost hat on, cloudwatch metrics are $3.60/year/metric - so this will work out around $500/year for playground, I imagine similar for prod. Definitely cheaper than spending hours guessing where we're exhausting our connection pool but worth thinking about.

philmcmahon:
> ## todo
> 
>     * [ ]  tell AWS about the missing bit in their docs (the `cloudwatchmetrics` dependency)
> 
>     * [ ]  create a departmental recommendation

I think the departmental recommendation should come with a cost warning

philmcmahon (approved):

joelochlann:
great points about cost!

joelochlann:
I'll wait to see if there's any feedback in Eng channel before posting recommendation

PR #65, opened 2022-08-11
Fix email preview, use wkhtmltopdf for html to pdf conversion

philmcmahon:
## What does this change?
Email preview is currently broken. This is due to some linux permissions issue when giant tries to call out to chromium-browser to generate the PDF preview. 

It would definitely be possible to fix the permissions issue (giant would need write access to /home/pfi/snap I think) but @itsibitzi suggested abandoning chromium-browser and instead using wkthmltopdf.

This works round the permissions issue, and should somewhat speed up the preview process now that we're not spinning up an entire browser to generate the pdf.

Separately, I noticed that we're not properly url encoding the filename in the url we pass to PDF.js - this PR also fixes that. 

## How to test
Open any .eml file in giant - here's an example on playground https://playground.pfi.gutools.co.uk/viewer/CAGHe3Pmc37CiMwbWks-CVOiM=uYhnpvq2=Zg51dkESL80On94g@mail.gmail.com?details=summary&view=preview

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,9 @@ There are two optional dependencies:
   - `brew install tesseract`
 - [Libre Office](https://www.libreoffice.org/)
   - To convert and preview Microsoft Office documents in the UI
+- [wkhtmltopdf](https://wkhtmltopdf.org/)
+  - To preview html files (such as emails)
+  - `brew install wkhtmltopdf`
 
 Elasticsearch requires Docker to have at least 4GB of memory from the preferences menu otherwise
 it will exit with no log output and error 137.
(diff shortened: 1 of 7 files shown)


joelochlann (approved):
Great work Phil! Do you have a screenshot of the new preview?

joelochlann:
Also I guess there's a corresponding Amigo change? Which I'm sure you've already done because you're an infra guru. Just always weirds me out that that stuff lives outside version control

philmcmahon:
Thanks Joe! Yep I've added whtmltopdf to amigo [here](https://amigo.gutools.co.uk/recipes/investigations-giant-app-java11) - not removed chrome yet as the same AMI is used by playground and prod.

Here's a screenshot. I'm not sure how much better/worse than the old chromium renders it is. 


Screenshot 2022-08-12 at 13 56 10

PR #66, opened 2022-08-15
Add DEV to giant allowed stages to support DEV db in aws

philmcmahon:
## What does this change?
This adds 'DEV' to the list of allowed stages for giant, in order to support AMI updates/deployment of the new remote Neo4j DEV instance we're setting up to work around M1 incompatibility with Neo4j 3. 

This should be a temporary thing - longer term we need to get onto Neo4j 4 as soon as possible


philmcmahon:
Note - we worked out how to run neo4j on an m1 chip via docker locally so no need for this

PR #67, opened 2022-08-17
Support local development on ARM64 laptops

philmcmahon:
## What does this change?
The current docker image we're using for neo4j doesn't support M1 arm64 macbooks. Neo4j don't support version 3.5 of neo4j anymore so aren't providing images but there is an experimental one we can use that seems to work: https://hub.docker.com/r/neo4j/neo4j-arm64-experimental/tags?page=1&name=3.

Elasticsearch also fails to start on a M1. @joelochlann  found that [this](https://stackoverflow.com/questions/50314586/elasticsearch-fails-to-start-config-seccomp-and-config-seccomp-filter-are-neede) solution worked - setting `bootstrap.system_call_filter=false` 

Note that people using M1 chips will need to either use the start-backend script or set `export NEO4J_IMAGE_OVERRIDE=neo4j/neo4j-arm64-experimental:3.5.30` prior to running docker compose in order to set docker to user the experimental image

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,3 +1,4 @@
 brew "tesseract"
 cask "libreoffice"
-brew "wkhtmltopdf"
+cask "wkhtmltopdf"
+brew "ocrmypdf"
--- a/Brewfile.lock.json
+++ b/Brewfile.lock.json
@@ -39,6 +39,45 @@
             }
           }
         }
+      },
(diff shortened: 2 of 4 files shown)


joelochlann (approved):
Nice one Phil! I've just road-tested it and was able to get Giant running on my M1 machine

PR #70, opened 2022-09-28
Delete page previews from S3

joelochlann:
## What's the issue?

When we do deletion, we haven't been deleting the individual page previews from S3.

## Where are these page previews stored?

They are stored under [`s3://pfi-giant-preview-rex/pages/${language}`](https://s3.console.aws.amazon.com/s3/buckets/pfi-giant-preview-rex?region=eu-west-1&prefix=pages/&showversions=false
)

Screenshot 2022-09-28 at 10 29 24

Note that `text` and `ocr.english` are legacy buckets from the time period between https://github.com/guardian/pfi/pull/884 and https://github.com/guardian/pfi/pull/929. We don't bother trying to delete from these, because we're going to do a one-off re-processing task of these items, followed by a one-off cleanup task deleting anything not in Giant.

## What's the fix?

When we delete a blob, delete objects in that bucket with prefix `pages/${language}/${blobPath}`, based on the languages on the `:PROCESSED` relation between the blob and the `OcrMyPdfExtractor`.

The individual objects correspond to pages, e.g.

```
s3://pfi-giant-preview-rex/pages/english/0/2/2/A/5/g/022A5gFy9WH_Q7dD3fkonxSfAf1iMCKzVzD5f9lNwkzrM19_FeKhtz8xqlgpy8zhNmnGV1im58Audz-CFdKyrQ/1.pdf
```

so in that case it should get everything starting

```
s3://pfi-giant-preview-rex/pages/english/0/2/2/A/5/g/022A5gFy9WH_Q7dD3fkonxSfAf1iMCKzVzD5f9lNwkzrM19_FeKhtz8xqlgpy8zhNmnGV1im58Audz-CFdKyrQ
```

## Do all `OcrMyPdfExtractor` `[:PROCESSED]` relations have languages set?
Yes

```
MATCH (r: Resource)<-[p:PROCESSED]-(e :Extractor {name: "OcrMyPdfExtractor"})
RETURN p.languages, count(distinct r)
```

| languages | count(distinct r) |
| ------------- | ------------- |
| ["english"] | 396,837 |
| ["english", "french", "russian"] | 2,325 |
| ["english", "french"] | 1,687 |


joelochlann:
Tests are passing now. I just need to test locally that it's doing what it should

joelochlann:
Seems to be working locally. I'll test on Playground too though

docker-compose.yml
```diff
@@ -35,7 +35,7 @@ services:
     image: lmenezes/cerebro:0.9.3
     container_name: pfi-cerebro
     ports:
-      - 9091:9000
+      - 9092:9000
```
philmcmahon, line 38: what's the motivation for changing this?

philmcmahon (approved):
Excellent work 

delete delete delete

docker-compose.yml
joelochlann: I needed to get minio port forwarding working so I could inspect local S3. For that I used 9091 since minio already exposes 9090.

This meant I had to bump this port number.

This change could've done with its own PR - normally I go the extra mile to prise apart the different concerns that have accumulated on a branch but this task is dragging on a bit so I didn't bother this time.

philmcmahon: 👍🏻 makes sense!

PR #72, opened 2022-09-30
getBlobs by collection

joelochlann:
Make the ingestion param optional so you can get all blobs under a collection without also specifying an ingestion.

This is so the CLI can have a "delete by collection" functionality. Every use case we've had for deletion via the CLI involves deleting an entire collection and it's annoying to have to find out all the ingestions first and pass them as an array to the CLI.

This could also form the basis of a (very dangerous! probably admin-only) "delete collection" button in the UI, if we wanted it.

### DIFF
--- a/backend/app/controllers/api/Blobs.scala
+++ b/backend/app/controllers/api/Blobs.scala
@@ -24,22 +24,11 @@ class Blobs(override val controllerComponents: AuthControllerComponents, manifes
     checkPermission(CanPerformAdminOperations, req) {
       (param("collection", req), param("ingestion", req)) match {
         case (Some(collection), maybeIngestion) =>
-          val parentExists = maybeIngestion match {
-            case Some(ingestion) => manifest.getIngestion(Uri(collection).chain(ingestion))
-            case _ => manifest.getCollection(Uri(collection))
-          }
-
           val size = param("size", req).map(_.toInt).getOrElse(500)
 
-          for {
-            _ <- parentExists
-            blobs <- index.getBlobs(collection, maybeIngestion, size)
-          } yield {
-            Ok(Json.obj(
-              "blobs" -> blobs
-            ))
-          }
-
+          index.getBlobs(collection, maybeIngestion, size).map(blobs =>
+            Ok(Json.obj("blobs" -> blobs))
+          )
         case _ =>
           Attempt.Right(BadRequest("Missing collection query parameter"))
       }
--- a/backend/app/controllers/api/Blobs.scala
+++ b/backend/app/controllers/api/Blobs.scala
@@ -23,13 +23,17 @@ class Blobs(override val controllerComponents: AuthControllerComponents, manifes
   def getBlobs = ApiAction.attempt { req =>
     checkPermission(CanPerformAdminOperations, req) {
       (param("collection", req), param("ingestion", req)) match {
-        case (Some(collection), Some(ingestion)) =>
-          val uri = Uri(collection).chain(ingestion)
+        case (Some(collection), maybeIngestion) =>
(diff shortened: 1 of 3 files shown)


backend/app/services/index/ElasticsearchResources.scala
```diff
@@ -442,13 +442,18 @@ class ElasticsearchResources(override val client: ElasticClient, indexName: Stri
 
   }
 
-  def getBlobs(collection: String, ingestion: String, size: Int): Attempt[Iterable[IndexedBlob]] = {
+  def getBlobs(collection: String, maybeIngestion: Option[String], size: Int): Attempt[Iterable[IndexedBlob]] = {
+    val query = maybeIngestion match {
+      case Some(ingestion) => matchQuery(IndexFields.ingestionRaw, s"$collection/$ingestion")
+      case _ => matchQuery(IndexFields.collectionRaw, collection)
+    }
+
```
philmcmahon, line 450: scala people always used to tell me not to pattern match on an option, an alternative would be:

```
val path = ingestion.map(i => "$collection/$i).getOrElse(collection)
val query = matchQuery(IndexFields.collectionRaw, path`
```

backend/app/controllers/api/Blobs.scala
```diff
@@ -23,21 +23,25 @@ class Blobs(override val controllerComponents: AuthControllerComponents, manifes
   def getBlobs = ApiAction.attempt { req =>
     checkPermission(CanPerformAdminOperations, req) {
       (param("collection", req), param("ingestion", req)) match {
-        case (Some(collection), Some(ingestion)) =>
-          val uri = Uri(collection).chain(ingestion)
+        case (Some(collection), maybeIngestion) =>
+          val parentExists = maybeIngestion match {
```
philmcmahon, line 27: could you add an explicit type here? getINgestion/getCollection sound like they might return an object but parentExists looks like a boolean

philmcmahon (approved):
Some suggestions but nothing crucial

backend/app/controllers/api/Blobs.scala
joelochlann: Yeah, fair point. I couldn't really understand why Michael did `manifest.getIngestion(uri)` and then threw away the result. My best interpretation was that he's checking whether the ingestion exists before counting its blobs. (Whether this is actually necessary, I haven't checked).

So I tried to rename to `parentExists` to make it clearer, but you're right, it sounds like it contains a boolean, when actually it's just being use as a boolean

backend/app/services/index/ElasticsearchResources.scala
philmcmahon: oops sorry, it's funny I did think I'd looked at the first argument but clearly not. 

here's the argument: https://www.originate.com/thinking/idiomatic-scala-your-options-do-not-match

I think pattern matching is probably more readable than the map getOrElse pattern though, certainly more readable than `fold`

backend/app/controllers/api/Blobs.scala
joelochlann: I think I might just delete that whole initial check for whether the collection/ingestion exists - don't see why it's necessary since the ES query should just return empty in such a case

backend/app/services/index/ElasticsearchResources.scala
joelochlann: Interesting. Though I totally reject their premise that concise = better. Why do we even bother naming our variables? It would be more concise to just call them `a`, `b`, etc.

The type safety point is the most compelling. I hadn't spotted that in my pattern match the type had become `Any` because one of the matches returned `Attempt[Ingestion]` and the other `Attempt[Collection]`

PR #73, opened 2022-10-03
If --redo-ocr fails with input file error, fall back to --skip-text

philmcmahon:
## What does this change?
We recently noticed that if `--redo-ocr` is set as an ocrmypdf option it fails for PDFs with fillable forms:

`This PDF has a user fillable form. --redo-ocr is not currently possible on such files.`

The error is an 'input file error' - in this PR we attempt to detect all input file errors, and re-run the ocr job with `--skip-text` set instead of `--redo-ocr`. 

@joelochlann gives a description of the differences between these settings here https://github.com/guardian/giant/pull/68

## How to test
I tested this on playground using this PDF: http://foersom.com/net/HowTo/data/OoPdfFormExample.pdf and it seemed to work ok. It fails on prod. 

Fewer OCR failures!

### DIFF
--- a/backend/app/utils/Ocr.scala
+++ b/backend/app/utils/Ocr.scala
@@ -67,9 +67,7 @@ object Ocr extends Logging {
         throw new OcrSubprocessCrashedException(exitCode, stderr.getOutput)
     }
   }
-
-
-
+  
   // TODO MRB: allow OcrMyPdf to read DPI if set in metadata
   // OCRmyPDF is a wrapper for Tesseract that we use to overlay the OCR as a text layer in the resulting PDF
   def invokeOcrMyPdf(lang: String, inputFilePath: Path, dpi: Option[Int], stderr: OcrStderrLogger, tmpDir: Path): Path = {
@@ -85,6 +83,9 @@ object Ocr extends Logging {
 
     val redoOcrExitCode = process(redoOcr=true)
     val exitCode = if (redoOcrExitCode == 2) {
+      // Exit code 2 from ocrmypdf is an input file error, we've noticed that this can be an error with --redo-ocr, and that
+      // running with --skip-text instead results in success. For example, if a PDF has a user fillable form then it can't
+      // be ocrd with --redo-ocr set. See https://github.com/guardian/giant/pull/68 for details of --skip-text vs --redo-ocr
       logger.info(s"Got input file error from ocrmypdf with --redo-ocr for ${inputFilePath.getFileName}, attempting with --skip-text")
       process(redoOcr=false)
     } else redoOcrExitCode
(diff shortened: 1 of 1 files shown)


backend/app/utils/Ocr.scala
```diff
@@ -67,16 +67,28 @@ object Ocr {
...
+    def process(redoOcr: Boolean): Int = {
+      val redoOcrOrSkipText = if(redoOcr) "--redo-ocr" else "--skip-text"
+      val cmd = s"ocrmypdf $redoOcrOrSkipText -l $lang ${dpi.map(dpi => s"--image-dpi $dpi").getOrElse("")} ${inputFilePath.toAbsolutePath} ${tempFile.toAbsolutePath}"
+      val process = Process(cmd, cwd = None, extraEnv = "TMPDIR" -> tmpDir.toAbsolutePath.toString)
+      process.!(ProcessLogger(stdout.append(_), stderr.append))
+    }
+
+    val redoOcrExitCode = process(redoOcr=true)
+    val exitCode = if (redoOcrExitCode == 2) {
+      // Exit code 2 from ocrmypdf is an input file error, we've noticed that this can be an error with --redo-ocr, and that
```
joelochlann, line 86: link to OcrMyPdf docs could be good here

backend/app/utils/Ocr.scala
```diff
@@ -67,16 +67,28 @@ object Ocr {
...
   // OCRmyPDF is a wrapper for Tesseract that we use to overlay the OCR as a text layer in the resulting PDF
   def invokeOcrMyPdf(lang: String, inputFilePath: Path, dpi: Option[Int], stderr: OcrStderrLogger, tmpDir: Path): Path = {
     val tempFile = tmpDir.resolve(s"${inputFilePath.getFileName}.ocr.pdf")
-    val cmd = s"ocrmypdf --redo-ocr -l $lang ${dpi.map(dpi => s"--image-dpi $dpi").getOrElse("")} ${inputFilePath.toAbsolutePath} ${tempFile.toAbsolutePath}"
-
     val stdout = mutable.Buffer.empty[String]
-    val process = Process(cmd, cwd = None, extraEnv = "TMPDIR" -> tmpDir.toAbsolutePath.toString)
-    val exitCode = process.!(ProcessLogger(stdout.append(_), stderr.append))
+
+    def process(redoOcr: Boolean): Int = {
```
joelochlann, line 77: An enum of the options would be a slightly clearer argument to accept here. It would also provide a place to put comments describing what they mean and linking to any docs

joelochlann: (Could include `ForceOcr` as well in case we need it for something in future... slight YAGNI violation but given the edge cases for ocr'ing PDFs I think it could help future devs to have the menu of options laid out in the codebase)

joelochlann (approved):
Amazing, thanks for fixing Phil!

One suggestion around an enum instead of a boolean, up to you whether you feel strongly about it though.

backend/app/utils/Ocr.scala
philmcmahon: Good shout - have added this

PR #74, opened 2022-10-08
Remove Guava from common

mbarton:
I've been playing around learning and implementing [Webauthn](https://webauthn.guide) (Yubikeys etc). WIP branch here: [mbarton/giant/webauthn](https://github.com/mbarton/giant/tree/mbarton/webauthn).

To use [webauthn4j](https://webauthn4j.github.io) I had to update to Jackson 2.13 and the latest Play 2.8. As part of this I was going through all the libraries and noticed that we don't really need Guava in common since we're just using it to hash files.

We do need it in the CLI to walk the file tree so I've just moved the dependency there. The CLI doesn't depend on the Play Framework at all (just play-json) so it doesn't matter what version of Guava we use. I haven't upgraded it though just to keep the changes minimal.

One less thing to worry about for the next Play upgrade :)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -83,9 +83,6 @@ lazy val common = (project in file("common"))
     libraryDependencies ++= Seq(
       "org.typelevel" %% "cats-core" % "2.2.0",
       "com.typesafe.play" %% "play-json" % "2.9.0",
-      // This version has been chosen the match the transitive dependency from Play.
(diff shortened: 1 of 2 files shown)


joelochlann:
Señor Barton you absolute hero!

philmcmahon (approved):
Tested on Playground 👍🏻  Thanks Michael!

PR #75, opened 2022-10-08
Scala 2.13

mbarton:
Have some free work on me :)

Upgrades Scala to 2.13 and fixes the MANY MANY warnings that come as a result. Also updates SBT to the latest too. Where I've had to upgrade dependencies to get 2.13 builds I have been as conservative as possible and gone for the latest version I can that doesn't involve additional code changes.

I've done individual commits per compile error so you'll probably want to squash and merge. I'll put in some inline comments for anything that is a wider change.

It's a lot of files changed but fundamentally it's all rewrites to avoid deprecation warnings. All the tests still pass (including integration tests) and I've spun up Giant, gone through the genesis flow, logged in as the admin and uploaded a file successfully.

### DIFF
--- a/backend/app/controllers/api/PagesController.scala
+++ b/backend/app/controllers/api/PagesController.scala
@@ -100,9 +100,9 @@ class PagesController(val controllerComponents: AuthControllerComponents, manife
       hlText.highlights.map(span => (lang, span))
     }.groupBy(_._2).toList.map { case (lang, commonSpans) =>
       commonSpans.head
-    }.groupBy(_._1).mapValues(_.map(_._2)).filter { case (k, v) =>
+    }.groupBy(_._1).view.mapValues(_.map(_._2)).filter { case (k, v) =>
       v.nonEmpty
-    }
+    }.toMap
   }
 
   def getPagePreview(uri: Uri, pageNumber: Int) = ApiAction.attempt { req =>
--- a/backend/app/ingestion/phase2/IngestStorePolling.scala
+++ b/backend/app/ingestion/phase2/IngestStorePolling.scala
@@ -88,7 +88,7 @@ class IngestStorePolling(
       )
       pollCompleteFuture.onComplete {
         case Success(pollDuration) => schedulePoll(pollDuration)
-        case SFailure(NonFatal(t)) =>
+        case SFailure(t) =>
           logger.error("Exception whilst processing ingestion batch", t)
           metricsService.updateMetric(Metrics.batchesFailed)
           schedulePoll(maximumWait)
--- a/backend/app/services/index/HitReaders.scala
+++ b/backend/app/services/index/HitReaders.scala
@@ -252,7 +252,7 @@ object HitReaders {
 
   private def readOcr(fields: FieldMap): Option[Map[String, String]] = {
     fields.optField[FieldMap](ocr).map { languages =>
-      languages.mapValues(_.asInstanceOf[String])
+      languages.view.mapValues(_.asInstanceOf[String]).toMap
     }
   }
 
--- a/backend/app/services/index/HitReaders.scala
+++ b/backend/app/services/index/HitReaders.scala
@@ -33,7 +33,7 @@ object HitReaders {
     def optLongField(name: String): Option[Long] = if(fields.contains(name)) { Some(longField(name)) } else { None }
     def listField[T](name: String): List[T] = optField[List[T]](name).getOrElse(Nil)
     def setField[T](name: String): Set[T] = optField[List[T]](name).map(_.toSet).getOrElse(Set.empty)
-    def optEnumField[T <: EnumEntry](name: String, enum: PlayEnum[T]): Option[T] = optField(name).flatMap(enum.withNameOption)
+    def optEnumField[T <: EnumEntry](name: String, playEnum: PlayEnum[T]): Option[T] = optField(name).flatMap(playEnum.withNameOption)
     def objectField(name: String): FieldMap = field[FieldMap](name)
     def nestedField(name: String): Map[String, Seq[String]] = {
       val base = Map.empty[String, Seq[String]]
--- a/backend/app/utils/Neo4jHelper.scala
+++ b/backend/app/utils/Neo4jHelper.scala
@@ -38,29 +38,29 @@ class Neo4jHelper(driver: Driver, executionContext: ExecutionContext, queryLoggi
     */
   class AttemptWrappedTransaction(underlying: StatementRunner, executionContext: ExecutionContext) {
     def run(statementTemplate: String, statementParameters: Record): Attempt[StatementResult] =
-      attemptNeo4J {
+      attemptNeo4J({
         underlying.run(statementTemplate, statementParameters)
-      }
+      })()
 
     def run(statementTemplate: String): Attempt[StatementResult] =
-      attemptNeo4J {
+      attemptNeo4J({
         underlying.run(statementTemplate)
-      }
+      })()
 
     def run(statementTemplate: String, parameters: Value): Attempt[StatementResult] =
-      attemptNeo4J {
+      attemptNeo4J({
         underlying.run(statementTemplate, parameters)
-      }
+      })()
 
     def run(statementTemplate: String, statementParameters: java.util.Map[String, AnyRef]): Attempt[StatementResult] =
-      attemptNeo4J {
+      attemptNeo4J({
         underlying.run(statementTemplate, statementParameters)
-      }
+      })()
 
     def run(statement: Statement): Attempt[StatementResult] =
-      attemptNeo4J {
+      attemptNeo4J({
         underlying.run(statement)
-      }
+      })()
 
     def run(statementTemplate: String, parameters: (String, AnyRef)*): Attempt[StatementResult] = {
       val flattenedParameters = parameters.flatten { case (k, v) => Vector(k,v) }
--- a/backend/app/utils/Neo4jHelper.scala
+++ b/backend/app/utils/Neo4jHelper.scala
@@ -9,9 +9,8 @@ import org.neo4j.driver.v1._
 import org.neo4j.driver.v1.types.TypeSystem
 import play.api.Logger
 import services.Neo4jQueryLoggingConfig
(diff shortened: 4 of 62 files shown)


backend/app/controllers/api/PagesController.scala
```diff
@@ -100,9 +100,9 @@ class PagesController(val controllerComponents: AuthControllerComponents, manife
       hlText.highlights.map(span => (lang, span))
     }.groupBy(_._2).toList.map { case (lang, commonSpans) =>
       commonSpans.head
-    }.groupBy(_._1).mapValues(_.map(_._2)).filter { case (k, v) =>
+    }.groupBy(_._1).view.mapValues(_.map(_._2)).filter { case (k, v) =>
```
mbarton, line 103: So `mapValues` is deprecated, so I added `.view` as suggested by the warning. We use enough of this call that I think the tests would catch if it's semantically different somehow but maybe worth additional testing?

backend/app/ingestion/phase2/IngestStorePolling.scala
```diff
@@ -88,7 +88,7 @@ class IngestStorePolling(
       )
       pollCompleteFuture.onComplete {
         case Success(pollDuration) => schedulePoll(pollDuration)
-        case SFailure(NonFatal(t)) =>
+        case SFailure(t) =>
```
mbarton, line 91: The 2.13 compiler is now smart enough to know this wasn't an exhaustive match as it would fail on a fatal error but tbh what chance do we have in that situation (out of memory etc) anyway. So I just fixed the warning by widening the match.

backend/app/services/index/HitReaders.scala
```diff
@@ -33,7 +33,7 @@ object HitReaders {
     def optLongField(name: String): Option[Long] = if(fields.contains(name)) { Some(longField(name)) } else { None }
     def listField[T](name: String): List[T] = optField[List[T]](name).getOrElse(Nil)
     def setField[T](name: String): Set[T] = optField[List[T]](name).map(_.toSet).getOrElse(Set.empty)
-    def optEnumField[T <: EnumEntry](name: String, enum: PlayEnum[T]): Option[T] = optField(name).flatMap(enum.withNameOption)
```
mbarton, line 36: Fixes a warning that `enum` is a reserved word in Scala 3

backend/test/controllers/api/UsersTest.scala
```diff
@@ -39,7 +39,7 @@ class UsersTest extends AnyFreeSpec with Matchers with Results with ScalaFutures
 
   "UsersController" - {
     "list partial user information to punters" in {
-      TestSetup(users + (paul -> (UserPermissions.bigBoss, List.empty), barry -> (UserPermissions.default, List.empty)), barry) { (controller, _) =>
```
mbarton, line 42: `+` with varargs is deprecated so it's easier just to change the test to include the default set of users automatically (see the bottom of the file)

build.sbt
```diff
@@ -12,8 +12,7 @@ val compilerFlags = Seq(
   "-unchecked",
   "-deprecation",
   "-feature",
-  "-Xfatal-warnings",
-  "-Ypartial-unification"
```
mbarton, line 16: on by default in 2.13

build.sbt
```diff
@@ -80,6 +79,7 @@ lazy val root = (project in file("."))
 lazy val common = (project in file("common"))
   .settings(
     name := "common",
+    scalacOptions := compilerFlags,
```
mbarton, line 82: this was missing by accident, adding it flagged up a few more deprecation warnings

build.sbt
```diff
@@ -112,7 +112,6 @@ lazy val backend = (project in file("backend"))
       "org.elasticsearch.client" % "elasticsearch-rest-client-sniffer" % "7.9.2",
       "com.typesafe.akka" %% "akka-cluster-typed" % "2.6.5", // should match what we get transitively from Play
       "org.neo4j.driver" % "neo4j-java-driver" % "1.6.3",
-      "org.scala-lang.modules" %% "scala-java8-compat" % "0.9.1",
```
mbarton, line 115: not needed in 2.13

cli/src/main/scala/com/gu/pfi/cli/Main.scala
```diff
@@ -17,14 +17,14 @@ import scala.concurrent.duration.Duration
...
   import scala.language.reflectiveCalls
 
-  val options = new Options(args)
+  val options = new Options(args.toIndexedSeq)
 
   options.subcommand match {
     case Some(options.hashCmd) =>
       HashFiles.run(options.hashCmd.files())
 
-    case Some(cmd @ options.loginCmd) =>
```
mbarton, line 26: I have no idea why this stopped working but it's arguably clear afterwards anyway

philmcmahon:
Omg! Wow Michael thanks so much this is awesome! We had ada lovelace day today so this slipped through but what a generous thing to do! Gonna get this reviewed and merged asap

joelochlann:
image

Thank you so much Michael! Single-handedly achieving health OKRs while no longer actually employed here. We owe you several rounds at the Star of Kings!

mbarton:
haha no worries, I didn't realise it was already in your plans!

mchv:
Legendary @mbarton 💎

paperboyo:
💎!

project/plugins.sbt
```diff
@@ -1,15 +1,10 @@
...
 
-libraryDependencies += "org.vafer" % "jdeb" % "1.6" artifacts (Artifact("jdeb", "jar", "jar"))
+libraryDependencies += "org.vafer" % "jdeb" % "1.10" artifacts (Artifact("jdeb", "jar", "jar"))
 
 addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.2")
 
-addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.1.0-M7")
+addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.18")
 
-addSbtPlugin("com.twitter" % "scrooge-sbt-plugin" % "18.4.0")
```
philmcmahon, line 11: is this just leftover from a time when we thought giant might want to eat some thrift?

mbarton: Yeah Giant did once have Thrift :)

Simon wrote some performance tracking on the ingestion pipeline that was stored as Thrift files in S3.

My guess is that we forgot to remove this when we removed the performance stuff.

common/src/main/scala/utils/attempt/Attempt.scala
```diff
@@ -106,7 +104,7 @@ object Attempt {
     * This implementation returns the first failure in the resulting list,
     * or the successful result.
     */
-  def traverse[A, B, M[X] <: TraversableOnce[X]](as: M[A])(f: A => Attempt[B])(implicit cbf: CanBuildFrom[M[A], B, M[B]], ec: ExecutionContext): Attempt[M[B]] = {
+  def traverse[A, B, M[X] <: TraversableOnce[X]](as: M[A])(f: A => Attempt[B])(implicit cbf: BuildFrom[M[A], B, M[B]], ec: ExecutionContext): Attempt[M[B]] = {
```
joelochlann: wow, special thanks due for this one as neither of these files are mentioned in the warning... think it would've taken me a while to unpick if I'd had to do it!

backend/app/utils/Neo4jHelper.scala
```diff
@@ -15,6 +15,7 @@ import scala.collection.mutable
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.control.NonFatal
 import scala.collection.JavaConverters._
+import utils.Logging
```
joelochlann, line 18: don't understand how this one even worked before! because they're both in a package called `utils`?

cli/src/main/scala/com/gu/pfi/cli/Main.scala
joelochlann: maybe worth us testing the CLI before merging, though if it type-checks it's probably fine

build.sbt
```diff
@@ -148,8 +148,8 @@ lazy val backend = (project in file("backend"))
 
       "org.scalacheck" %% "scalacheck" % "1.14.0" % Test,
       "org.scalatestplus.play" %% "scalatestplus-play" % "5.1.0" % Test,
-      "com.whisk" %% "docker-testkit-scalatest" % "0.11.0" % Test,
-      "com.whisk" %% "docker-testkit-impl-spotify" % "0.11.0-beta1" % Test,
+      "com.whisk" %% "docker-testkit-scalatest" % "0.9.9" % Test,
```
joelochlann, line 148: could you explain this change a little more?

joelochlann (approved):
Seems like @philmcmahon has done some thorough testing, so happy to merge

backend/app/utils/Neo4jHelper.scala
mbarton: hmm I got confused by this one too, turns out `import org.neo4j.driver.v1._` from above brought in `org.neo4j.driver.v1.Logging` which in 2.13 seems to take precedence over the one in the `utils` package. By importing `utils.Logging` I was then shadowing with the correct import again.

Digging in a bit more it looks like the way it worked before wasn't really supposed to work at all by the spec so they fixed it in 2.13: https://github.com/scala/scala/pull/6589. Nic hit the same thing in Riff-Raff: https://github.com/guardian/riff-raff/pull/666

backend/test/services/manifest/Neo4JManifestITest.scala
```diff
@@ -97,7 +97,7 @@ class Neo4JManifestITest extends AnyFreeSpec with Matchers with Neo4jTestService
       "Can insert emails" in {
         val result = manifest.insert(emails, Uri("test-collection/test-ingestion"))
         if (result.isLeft) {
-          val value = result.left.get
+          val value = result.swap.toOption.get
```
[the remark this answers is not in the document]
mbarton: I agree. I was following what the deprecation warning said though so in the interests of keeping this PR as a set of "straight forward" compile fixes I think it should be a separate PR?

mbarton:
Fixed another conflict with the routes file :)

PR #79, opened 2022-10-12
`inMultiple` flag to get only those blobs that are in multiple collections

joelochlann:
## `inMultiple` flag
When fetching blobs, you can now pass `inMultiple=true` to get only blobs that are also in other collections/ingestions than those supplied.

This is useful to do before deleting a collection, so you can check that blobs also exist in other collections are safe to delete.

### For instance
```url
?collection=c&inMultiple=true
```
 returns blobs in collection c and at least one other collection
 
```url
?collection=c&ingestion=i&inMultiple=true
```
returns blobs in ingestion c/i and at least one other ingestion

## Rename `collection` => `collections` and `ingestion` => `ingestions`
Because they're arrays

## Return `collections` property with blobs
Because it's in elasticsearch anyway and it's useful not to have to compute this from the ingestions when consuming the API.

```json
{
  "blobs": [
    {
      "uri": "MVGTRpihjQaCY3Rfn3RJIuXOwseGAPVo3ITre4Svc_oArVwofwBya0cGIJqYF8sdFOzP30Lb3u_4nS-qqVH6cQ",
      "collections": [
        "joe Documents"
      ],
      "ingestions": [
        "joe Documents/Upload 2022-10-12T14:54:20.113Z",
        "joe Documents/Upload 2022-10-12T15:17:03.146Z"
      ]
    }
  ]
}
```

### DIFF
--- a/backend/app/controllers/api/Blobs.scala
+++ b/backend/app/controllers/api/Blobs.scala
@@ -20,15 +20,21 @@ class Blobs(override val controllerComponents: AuthControllerComponents, manifes
 
   def param(name: String, req: Request[AnyContent]): Option[String] = req.queryString.get(name).flatMap(_.headOption)
 
-  def getBlobs = ApiAction.attempt { req =>
+  // inMultiple means only return blobs that are also in other collections/ingestions than those supplied.
+  // For instance:
+  //   ?collection=c&inMultiple=true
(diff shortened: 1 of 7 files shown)


philmcmahon (approved):
Sounds really useful, code LGTM

PR #80, opened 2022-10-13
Mbarton/remove guava from common 2

philmcmahon:
## What does this change?
Not sure this is the way to do it but merging this into a feature branch so we can test it - see https://github.com/guardian/giant/pull/74 for details


joelochlann:
does this mean you tested it successfully? what do we do with Michael's original branch then? just merge as well (presumably it's a no-op now)?

philmcmahon:
I merged Michael's change into a dev branch I created so that teamcity would build it (it doesn't build forks for security reasons). Now that it's tested I will delete the dev branch and have merged Michael's original PR

PR #82, opened 2022-10-17
Delete from elasticsearch last

joelochlann:
We use elasticsearch to determine what blobs are in a collection.

This means that once we've deleted from elasticsearch, the blob no longer exists from the perspective of a "delete collection" request.

So we should delete from elasticsearch last, so that if any of the other operations fail (delete from neo4j, delete from S3), we are still able to clear things up by restarting the delete collection operation.


philmcmahon:
Could you add an explanation of what `index` is in the PR description? I guess it's easy to work out from the code - is it the elastic document?

philmcmahon (approved):
:shipit:

PR #84, opened 2022-10-17
Add count blobs endpoint

joelochlann:
Currently the `getBlobs` endpoint can only return up to 500 blobs due to limitations of the elasticsearch search API. We could page through to get more, though it's not really suitable for returning hundreds of thousands of results (there is an inbuilt limit of [10,000](https://www.elastic.co/guide/en/elasticsearch/reference/7.8/paginate-search-results.html))

For that reason, it would be useful to have an endpoint for counting the number of blobs, so we can tell the user at the start of a deletion how much work there is to do, and track progress through it.


philmcmahon (approved):
counting

👍🏻 LGTM

PR #85, opened 2022-10-20
Deletion timing

joelochlann:
```
Getting langs from neo4j: 14 milliseconds
Delete page previews: 47 milliseconds
Preview S3 delete: 4 milliseconds
Ingest S3 delete: 4 milliseconds
Delete blob from neo4j: 16 milliseconds
Delete blob from elasticsearch: 11 milliseconds
Total to delete resource: 106 milliseconds
```

### DIFF
--- a/backend/app/commands/DeleteResource.scala
+++ b/backend/app/commands/DeleteResource.scala
@@ -7,25 +7,25 @@ import services.ObjectStorage
 import services.index.Index
 import services.manifest.Manifest
 import services.previewing.PreviewService
-import utils.Logging
+import utils.{Logging, Timing}
 import utils.attempt.{Attempt, DeleteFailure, IllegalStateFailure}
 
 import scala.concurrent.ExecutionContext
 
 
 class DeleteResource( manifest: Manifest, index: Index, previewStorage: ObjectStorage, objectStorage: ObjectStorage)  (implicit ec: ExecutionContext)
-   extends Logging {
+   extends Timing {
 
-     private def deletePagePreviewObjects(s3Objects: List[String]): Attempt[List[Unit]] = {
+     private def deletePagePreviewObjects(s3Objects: List[String]): Attempt[List[Unit]] = timeSync(s"Total to delete ${s3Objects.length} objects", {
        logger.info(s"Deleting objects: ${s3Objects.mkString(", ")}")
        val deletionAttempts = s3Objects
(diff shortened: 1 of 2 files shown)


backend/app/commands/DeleteResource.scala
```diff
@@ -32,7 +32,7 @@ class DeleteResource( manifest: Manifest, index: Index, previewStorage: ObjectSt
          // we introduced the OcrMyPdfExtractor.
          logger.info(s"No OCR languages found for blob ${uri.value}")
        } else {
-         logger.info(s"Deleting page previews for ${uri.value} in languages: ${ocrLanguages.map(_.key).mkString(", ")}")
+         // logger.info(s"Deleting page previews for ${uri.value} in languages: ${ocrLanguages.map(_.key).mkString(", ")}")
```
philmcmahon, line 35: remove?

common/src/main/scala/utils/Timing.scala
```diff
@@ -0,0 +1,35 @@
...
+import scala.concurrent.ExecutionContext
+
+trait Timing extends Logging {
+  private def log(name: String, totalElapsedNanoseconds: Long): Unit = {
+    // TODO: use Duration instead?
+    val elapsedNanoseconds = (totalElapsedNanoseconds % 1000000).toInt
+    val elapsedMilliseconds =(totalElapsedNanoseconds / 1000000).toInt
+    val elapsedSeconds = (elapsedMilliseconds / 1000).toInt
+    val elapsedMinutes = (elapsedSeconds / 60).toInt
+    logger.info(s"${name}: took ${elapsedMinutes} minutes ${elapsedSeconds} seconds ${elapsedMilliseconds} milliseconds ${elapsedNanoseconds} nanoseconds")
```
philmcmahon, line 14: do we care about nanoseconds?

philmcmahon (approved):
Some leftover logging but otherwise this look like a v helpful new functionality 👍🏻

common/src/main/scala/utils/Timing.scala
joelochlann: good point! just to millis now

backend/app/commands/DeleteResource.scala
joelochlann: actually I think re-instate, they're still useful in prod

PR #86, opened 2022-10-25
Batch S3 deletion requests

joelochlann:
After an excellent pairing session with @rtyley, we realised that we could get some performance gains by sending multiple keys to delete in a single S3 call, rather than making multiple calls.

This means we make at most two delete requests to S3: one to delete everything from the preview bucket (page previews and the whole-document preview), and one the delete the blob from the ingest bucket.

The biggest gains would be in documents with many pages since before we were making at least one call per page, e.g. 324 requests become 2 requests.

But even for a document with one page, we still reduce the S3 requests from 3 to 2.

# local perf stats
## before (for 322 page "Functional Programming in Scala" book)
**900ms**
```
Delete page previews: took 0 minutes 0 seconds 780 milliseconds
Preview S3 delete: took 0 minutes 0 seconds 5 milliseconds
Ingest S3 delete: took 0 minutes 0 seconds 5 milliseconds
Delete blob from neo4j: took 0 minutes 0 seconds 59 milliseconds
Delete blob from elasticsearch: took 0 minutes 0 seconds 18 milliseconds
Total to delete resource: took 0 minutes 0 seconds 900 milliseconds
```

## after
**145ms**
```
Get page preview S3 keys: took 0 minutes 0 seconds 54 milliseconds
Deleting 323 objects from preview storage
Preview storage S3 delete: took 0 minutes 0 seconds 61 milliseconds
Ingest storage S3 delete: took 0 minutes 0 seconds 4 milliseconds
Delete blob from neo4j: took 0 minutes 0 seconds 12 milliseconds
Delete blob from elasticsearch: took 0 minutes 0 seconds 5 milliseconds
Total to delete resource: took 0 minutes 0 seconds 145 milliseconds
```

# prod perf stats
## before
Note: these may be distorted by certain subsets of the documents in the collection I'm deleting having different properties. I've seen periods in which deletions went much quicker, presumably because it had a run of documents that had no page previews. I don't understand the order in which documents come back from the elasticsearch so don't know how much this clustering will tend to occur.

Averaged [500 deletions](http://localhost:5601/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'2022-10-25T07:36:58.871Z',to:'2022-10-25T09:36:27.579Z'))&_a=(columns:!(message),filters:!(),index:d2720940-0e30-11eb-a809-5706b391daea,interval:auto,query:(language:kuery,query:'%22Total%20to%20delete%20resource%22'),sort:!())): 263 ms total

### DIFF
--- a/backend/app/commands/DeleteResource.scala
+++ b/backend/app/commands/DeleteResource.scala
@@ -47,10 +47,9 @@ class DeleteResource( manifest: Manifest, index: Index, previewStorage: ObjectSt
        val prefixesToDelete = legacyPagePreviewPrefixes ::: pagePreviewPrefixes
         logger.info(s"Deleting prefixes: ${prefixesToDelete.mkString(", ")}")
 
-       val listOfPagePreviewObjectsAttempts = prefixesToDelete.map { prefix =>
+       Attempt.traverse(prefixesToDelete) { prefix =>
          Attempt.fromEither(previewStorage.list(prefix))
-       }
-       Attempt.sequence(listOfPagePreviewObjectsAttempts).map(_.flatten.toSet)
+       }.map(_.flatten.toSet)
      }
 
      private def deleteResource(uri: Uri): Attempt[Unit] = timeAsync("Total to delete resource", {
--- a/backend/app/commands/DeleteResource.scala
+++ b/backend/app/commands/DeleteResource.scala
@@ -16,12 +16,18 @@ import scala.concurrent.ExecutionContext
 class DeleteResource( manifest: Manifest, index: Index, previewStorage: ObjectStorage, objectStorage: ObjectStorage)  (implicit ec: ExecutionContext)
    extends Timing {
 
-     private def deletePagePreviewObjects(s3Objects: Set[String]): Attempt[Iterator[Unit]] = {
-       logger.info(s"Deleting ${s3Objects.size} objects")
-       Attempt.traverse(s3Objects.grouped(500))(x => Attempt.fromEither(previewStorage.deleteMultiple(x)))
+     private def deleteFromS3Preview(blobUri: Uri, pagePreviewKeys: Set[String]): Attempt[Iterator[Unit]] = {
+       // The full-document preview, as well as all the previews of individual pages
+       val keys = pagePreviewKeys + blobUri.toStoragePath
+       logger.info(s"Deleting ${keys.size} objects from preview storage")
+
+       // Group, just in case we have thousands of pages
+       Attempt.traverse(keys.grouped(500)) { batchOfS3Keys =>
+         Attempt.fromEither(previewStorage.deleteMultiple(batchOfS3Keys))
+       }
      }
 
-     private def deletePagePreviews(uri: Uri, ocrLanguages: List[Language]): Attempt[_] = {
+     private def getPagePreviewS3Keys(uri: Uri, ocrLanguages: List[Language]): Attempt[Set[String]] = {
        if (ocrLanguages.isEmpty) {
          // This typically means the blob was not processed by the OcrMyPdfExtractor,
          // either because it's not a PDF or because it was processed before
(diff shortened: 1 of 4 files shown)


backend/app/commands/DeleteResource.scala
```diff
@@ -16,16 +16,18 @@ import scala.concurrent.ExecutionContext
...
-         .map(Attempt.fromEither)
+     private def deleteFromS3Preview(blobUri: Uri, pagePreviewKeys: Set[String]): Attempt[Iterator[Unit]] = {
+       // The full-document preview, as well as all the previews of individual pages
+       val keys = pagePreviewKeys + blobUri.toStoragePath
+       logger.info(s"Deleting ${keys.size} objects from preview storage")
 
-       Attempt.sequence(deletionAttempts)
+       // Group, just in case we have thousands of pages
+       Attempt.traverse(keys.grouped(500)) { batchOfS3Keys =>
+         Attempt.fromEither(previewStorage.deleteMultiple(batchOfS3Keys))
```
rtyley, line 26: Lovely batching! The [AWS API docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html) say that 1000 object ids is the limit for a batch, I think we get a lot of benefit with a chunk size of 500.

backend/app/services/ObjectStorage.scala
```diff
@@ -38,8 +39,8 @@ class S3ObjectStorage private(client: S3Client, bucket: String) extends ObjectSt
     run(client.aws.deleteObject(bucket, key))
   }
 
-  def delete(keys: List[String]): Either[Failure, Unit] = {
-    val request = new DeleteObjectsRequest(bucket).withKeys(keys: _*)
+  def deleteMultiple(keys: Set[String]): Either[Failure, Unit] = {
+    val request = new DeleteObjectsRequest(bucket).withKeys(keys.toSeq: _*)
     run(client.aws.deleteObjects(request))
```
rtyley, line 44: This is probably not crucial, but I thought it was worth noting that a batch command like this can fail to delete some keys, but still return a response that is 'successful' in terms of HTTP response code. The 'things that failed' are [buried inside the response](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html#API_DeleteObjects_Examples:~:text=Object%3E%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%3C/Delete%3E-,Sample%20Response,-The%20response%20includes), eg:

```xml
HTTP/1.1 200 OK
            x-amz-id-2: 5h4FxSNCUS7wP5z92eGCWDshNpMnRuXvETa4HH3LvvH6VAIr0jU7tH9kM7X+njXx
            x-amz-request-id: A437B3B641629AEE
            Date: Fri, 02 Dec 2011 01:53:42 GMT
            Content-Type: application/xml
            Server: AmazonS3
            Content-Length: 251

            <?xml version="1.0" encoding="UTF-8"?>
            <DeleteResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
             <Deleted>
               <Key>sample1.txt</Key>
             </Deleted>
             <Error>
              <Key>sample2.txt</Key>
              <Code>AccessDenied</Code>
              <Message>Access Denied</Message>
             </Error>
            </DeleteResult>
```

...in this context, it might mean that the deleteMultiple() method should return failure if even one object fails to be deleted, and I'm not sure if the code would currently do this. However, as partial failure (eg due to access denied for just some objects, and not others) is unlikely to happen, this might not be a case that needs urgently handling.

rtyley (approved):

backend/app/services/ObjectStorage.scala
joelochlann: Great point. I think for the sake of our pseudo-transaction (do not delete the elasticsearch document until we're sure all other traces in S3 and neo4j are gone), it would probably be best to fail the whole thing if any individual deletion fails. However, we also don't want to consider "object does not exist" a failure because sometimes we try to delete previews where there is none, and that's fine. So that might complicate it...

joelochlann: Actually, the [Java SDK docs](https://docs.aws.amazon.com/zh_tw/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3.html#deleteObjects-com.amazonaws.services.s3.model.DeleteObjectsRequest-) say "If any object in the request cannot be deleted, this method throws a [MultiObjectDeleteException](https://docs.aws.amazon.com/zh_tw/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/MultiObjectDeleteException.html) with details of the error", so we should get a `Left` and fail as we would like.

joelochlann: I've also just quickly verified by testing that [this comment](https://github.com/guardian/giant/blob/main/backend/app/commands/DeleteResource.scala#L59) is still correct for multiple deletions. A multiple delete that includes one object that does not exist succeeds.

philmcmahon:
Nice!

backend/app/services/ObjectStorage.scala
rtyley: Ah fantastic, nice one!

PR #94, opened 2023-03-15
Bump junrar and okhttp libraries to fix vulnerability

philmcmahon:
## What does this change?
Moves these libraries to a later version. I'm going quite softly softly with giant dependencies - just one or two at a time - as it's in production and don't wanna break it etc


srbd (approved):
🔥

PR #105, opened 2023-05-02
Replace data-ui with elastic charts

philmcmahon:
## What does this change?
This change removes https://github.com/williaster/data-ui, replacing it with (a very old version of) [elastic charts](https://github.com/elastic/elastic-charts). data-ui is no longer being maintained and is was blocking our move to a more recent version of react-scripts (necessary to resolve a number of vulnerabilities).

### Why such an old version of elastic charts?
Any newer version of elastic charts uses typescript features incompatible with our current version. Changing our typescript version sets off an avalanche of required dependency upgrades (typescript, pdfjs-dist, elasticui, react-scripts) - which I've started in https://github.com/guardian/giant/tree/pm-histogram-pdfjsupgrade-mess (that code compiles but the page viewer is completely broken)

It still feels helpful to get us onto a library that is actually maintained, even if it's an ancient version. 

The new charts look somewhat similar, just a bit Elastic UIey. I'm open to styling them. 

Before:

Screenshot 2023-05-02 at 11 11 52

After:

Screenshot 2023-05-02 at 11 01 11

The tooltips used to be have a nicer date rather than mm/yyyy but then I had to downgrade about 30 versions of Elastic UI. Once we get back onto a more recent version the nice tooltip code is:

```
                    <Tooltip
                        headerFormatter={(tooltipData) => {
                            return renderTimeFancy(tooltipData.value);
                        }}
                    />
```

## How to test
Tested on playground - this change should only affect the date created histogram.

### DIFF
--- a/frontend/src/js/components/SearchResults/visualizations/TimeHistogram.js
+++ b/frontend/src/js/components/SearchResults/visualizations/TimeHistogram.js
@@ -7,19 +7,9 @@ import {
     Axis,
     Chart,
     Settings as EuiSettings,
-    ScaleType,  BarSeries
+    ScaleType, BarSeries
 } from "@elastic/charts";
 import '@elastic/charts/dist/theme_light.css';
-
-// const ResponsiveHistogram = withParentSize((props) => {
-//   return (
-//     <Histogram
-//         {...props}
-//         width={props.parentWidth}
-//         height={props.parentHeight}
-//       />
-//   );
-// });
 
 export default class TimeHistogram extends React.Component {
     static propTypes = {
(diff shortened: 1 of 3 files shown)


frontend/src/js/components/SearchResults/visualizations/TimeHistogram.js
```diff
@@ -84,31 +76,41 @@ export default class TimeHistogram extends React.Component {
             this.props.updateSearchText(query);
       }
 
+        const esData = binnedData.map(b => {
```
philmcmahon, line 79: elasticChartsData?

PR #106, opened 2023-05-02
Upgrade react-scripts, pdfjs-dist, elastic-ui

philmcmahon:
## What does this change?
This change drags Giant's frontend libraries kicking and screaming into some time in the not so recent past. The initial motivation for this work was to solve some vulnerabilities in `react-scripts`, which triggered a bit of an avalanche of other changes:

- Elastic Charts added (to replace data-ui, which is EOL -see https://github.com/guardian/giant/pull/105) 
- Update typescript to support recent a recent version of elastic charts
- Update elastic ui to find a version that matches our typescript version (ts is a peer dependency of elastic ui)

So far so good - no major changes, just some tsconfig tweaks. Unfortunately the react-scripts update broke pdfjs-dist, I think to do with conflicting babel/webpack configurations, resulting in errors like [this](https://github.com/mozilla/pdf.js/issues/13200) and [this](https://github.com/mozilla/pdf.js/issues/15083)

So I also had to update pdfjs-dist. Version 3 of pdfjs-dist seems like a good improvement - and tidies up our imports a bit. There were some breaking changes though:

 - `normalizeWhitespace` is [deprecated](https://github.com/mozilla/pdf.js/issues/14519)
 - `enhanceTextSelection` is [deprecated](https://github.com/mozilla/pdf.js/issues/9943#issuecomment-409369485) - apparently we should instead 'use textLayerMode = 2' - but I couldn't work out how to do that - hoping @joelochlann might have some idea as to how much we were relying on enhanceTextSelection

Other changes in this PR are related to tighter eslint rules around default exports needing to be named and redeclaration of e.g. `PropTypes`. You can see the relevant eslint warnings [here](https://teamcity.gutools.co.uk/buildConfiguration/InvestigationsReporting_GiantOpenSource/856122?expandBuildProblemsSection=true&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&showLog=856122_96_96&logFilter=debug&logView=flowAware)

For some reason the upgrades resulted in some style changes to the react-icons used in the page viewer controls. I added some styling to make things look more or less the same. Here's before/after:

Before

Screenshot 2023-05-02 at 16 16 42

After

Screenshot 2023-05-02 at 16 16 29

## How to test
I've tested this on playground and locally. The main area of concern is any impact on the page viewer.

We should see the number of vulnerabilities in Snyk go down.

This is quite a big change so may impact unknown areas of Giant's ui - sensible amounts of testing are required


philmcmahon:
Doh, just discovered this messes up the page viewer controls. Digging into this now. It's live on playground for testing otherwise 
Screenshot 2023-05-02 at 15 20 01

frontend/package.json
```diff
@@ -4,9 +4,10 @@
   "private": true,
   "dependencies": {
     "@elastic/charts": "15.0.5",
+    "@data-ui/histogram": "0.0.84",
```
philmcmahon, line 7: remove!

frontend/package.json
```diff
@@ -4,9 +4,10 @@
   "private": true,
   "dependencies": {
     "@elastic/charts": "15.0.5",
```
philmcmahon, line 6: remove!

frontend/src/js/components/SearchResults/visualizations/TimeHistogram.js
```diff
@@ -45,10 +46,21 @@ export default class TimeHistogram extends React.Component {
 
         binnedData = _.sortBy(binnedData, i => i.bin0);
 
+        const esData = binnedData.map(b => {
```
philmcmahon, line 49: check this isn't duplicated

frontend/tsconfig.json
```diff
@@ -17,7 +17,8 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react"
+    "jsx": "react-jsx",
+    "noFallthroughCasesInSwitch": true
```
philmcmahon, line 21: remove this?

frontend/src/js/components/SearchResults/visualizations/TimeHistogram.js
zekehuntergreen: ah did you just move it from line 79? 🤔

zekehuntergreen (commented):
I'm not able to install dependencies locally because I'm getting a 404 from github when trying to download [node-canvas](https://github.com/Automattic/node-canvas/releases/tag/v2.11.2)
Other than that this looks great - thanks for doing the painful upgrade!

PR #110, opened 2023-07-14
Authenticate against the downcased email coming in from panda cookie

srbd:
## What does this change?

The new standard for Guardian gsuite emails are lower case, although previously they might have been title case eg `Firstname.Lastname@theguardian.com`. We will switch to using lower case in Giant for all email usernames, which means authenticating against the downcased email from the panda cookie. We have a couple of users who are title case in the db which we will manually remove and re-add in their lowercase form. 

## How to test

We will need to test with a "title cased user" logging in successfully against their lowercase email as exists in Giant.


philmcmahon (approved):
going down

PR #112, opened 2023-07-17
Zoom in and zoom out pdf 

marjisound:
This PR is a collaboration with @philmcmahon and @joelochlann 

## What does this change?
This PR updates the pdf viewer with zoom-in/zoom-out functionality. It adds a `scale` state to the `PageViewer` which is used to control the container size and as a result controls how big or small the page should be rendered.

A new function `refreshPreview` is added in the  in PageCache class, that is responsible to update the existing preview in cache by rendering a new page and replacing the current preview in cache with the new preview. This way we avoid calling the server to retrieve preview for the page number.

In PdfHelpers, a new function `updatePreview` is implemented and when given the preview and a container size it renders a new page for that container size.

In the Page component, the `pageHeight` added to the component properties, is used to trigger updating the existing canvas with the canvas of the new preview. 

Because we always render only 3 pages at a time (for performance reasons), it was made sure that when the page range changes, the page in the new range is re-rendered if the scale (container size) has changed.

Below is demoing a long vertical page pdf and how the zooming and finding works.

multi-page-long

Below is demoing a landscape page pdf and how the zooming and finding works.
landscape-zoom

Below is demoing how the zooming works with search query. 
search-query

TODO in future PRs:
- ~~When using the zoom with search query avaiable, it's responding more slowly. Also, some pages could be left un-rendered!!~~ this was resolved
- ~~when clicking on zoom for several times e.g. 3 time, the page is rendered 3 times. For optimisation, we could maybe cancel the previous rendering tasks!?~~ this is now resolved by using debounce & useMemo 
- We could give user control on the scale of zooming e.g. 100%, 150%, etc. 
- Currently a new function is created in PageCache in order to refresh and re-render the relevant pdf page based on the container size. But this is not fully optimised, because it's replacing the existing cache, while it's best if we could just add to the cache in case the old preview size is needed again in the future. Since this optimisation needs change in the cache logic, and needs extra care, I will create a ticket to do this work separately.

### DIFF
--- a/frontend/src/js/components/PageViewer/Page.tsx
+++ b/frontend/src/js/components/PageViewer/Page.tsx
@@ -40,19 +40,22 @@ export const Page: FC<PageProps> = ({
     }
   };
 
-  useEffect(() => {
-    console.log('useEffect getPagePreview');
+  useEffect(() => {    
     if (containerRef.current) {
       getPagePreview
         .then((preview) => {
-          console.log('Render page preview into DOM');
           // Have to recheck here because the component may have dismounted
           const node = containerRef.current;
           if (node) {
             setScale(preview.scale);
-            // TODO DANGEROUS: this might completely mess up multi-page docs
-            node.innerHTML = '';
-            node.appendChild(preview.canvas);
+            if (node.hasChildNodes()) {
+              const oldCanvas = node.getElementsByTagName('canvas');
+              if (oldCanvas.length > 0) {
+                node.replaceChild(preview.canvas, oldCanvas[0]);
+              }                      
+            } else {
+              node.appendChild(preview.canvas);
+            }                        
           }
           setPreviewMounted(true);
           return preview;
@@ -67,7 +70,6 @@ export const Page: FC<PageProps> = ({
 
   useEffect(() => {
     getPageData.then((text) => {
-      console.log('Refresh page data');
       setPageText(text);
     }).catch(handleAbort);
   }, [containerRef, getPageData]);
--- a/frontend/src/js/components/PageViewer/Page.tsx
+++ b/frontend/src/js/components/PageViewer/Page.tsx
@@ -49,10 +49,10 @@ export const Page: FC<PageProps> = ({
           if (node) {
             setScale(preview.scale);
 
-            const oldCanvas = node.getElementsByTagName('canvas');
+            const oldCanvas = node.querySelector('canvas');
 
-            if (oldCanvas.length > 0) {
-              node.replaceChild(preview.canvas, oldCanvas[0]);
+            if (oldCanvas) {
+              node.replaceChild(preview.canvas, oldCanvas);
             } else {
               node.appendChild(preview.canvas);
             }
--- a/frontend/src/js/components/PageViewer/Page.tsx
+++ b/frontend/src/js/components/PageViewer/Page.tsx
@@ -48,14 +48,14 @@ export const Page: FC<PageProps> = ({
           const node = containerRef.current;
           if (node) {
             setScale(preview.scale);
-            if (node.hasChildNodes()) {
-              const oldCanvas = node.getElementsByTagName('canvas');
-              if (oldCanvas.length > 0) {
-                node.replaceChild(preview.canvas, oldCanvas[0]);
-              }                      
+
+            const oldCanvas = node.getElementsByTagName('canvas');
+
+            if (oldCanvas.length > 0) {
+              node.replaceChild(preview.canvas, oldCanvas[0]);
             } else {
               node.appendChild(preview.canvas);
-            }                        
+            }
           }
           setPreviewMounted(true);
           return preview;
--- a/frontend/src/js/components/PageViewer/Page.tsx
+++ b/frontend/src/js/components/PageViewer/Page.tsx
@@ -40,7 +40,7 @@ export const Page: FC<PageProps> = ({
     }
   };
 
-  useEffect(() => {    
+  useEffect(() => {
     if (containerRef.current) {
       getPagePreview
         .then((preview) => {
--- a/frontend/src/js/components/PageViewer/Page.tsx
+++ b/frontend/src/js/components/PageViewer/Page.tsx
@@ -11,6 +11,7 @@ type PageProps = {
   pageNumber: number;
   getPagePreview: Promise<CachedPreview>;
   getPageData: Promise<PageData>;
+  pageHeight: number;
 };
 
 export const Page: FC<PageProps> = ({
@@ -19,6 +20,7 @@ export const Page: FC<PageProps> = ({
   pageNumber,
   getPagePreview,
   getPageData,
+  pageHeight,
 }) => {
   const [pageText, setPageText] = useState<PageData | null>(null);
   const [scale, setScale] = useState<number | null>(null);
@@ -39,13 +41,17 @@ export const Page: FC<PageProps> = ({
   };
 
   useEffect(() => {
+    console.log('useEffect getPagePreview');
     if (containerRef.current) {
       getPagePreview
         .then((preview) => {
+          console.log('Render page preview into DOM');
           // Have to recheck here because the component may have dismounted
           const node = containerRef.current;
           if (node) {
             setScale(preview.scale);
+            // TODO DANGEROUS: this might completely mess up multi-page docs
+            node.innerHTML = '';
             node.appendChild(preview.canvas);
           }
           setPreviewMounted(true);
@@ -57,10 +63,13 @@ export const Page: FC<PageProps> = ({
         })
         .catch(handleAbort);
     }
-  }, [getPagePreview]);
+  }, [getPagePreview, pageHeight]);
 
   useEffect(() => {
-    getPageData.then(setPageText).catch(handleAbort);
+    getPageData.then((text) => {
+      console.log('Refresh page data');
+      setPageText(text);
+    }).catch(handleAbort);
   }, [containerRef, getPageData]);
 
   return (
--- a/frontend/src/js/components/PageViewer/PageCache.ts
+++ b/frontend/src/js/components/PageViewer/PageCache.ts
@@ -8,6 +8,7 @@ import {PDFWorker} from "pdfjs-dist";
 export type CachedPage = {
   previewAbortController: AbortController;
   preview: Promise<CachedPreview>;
+  previewContainerSize: number;
   dataAbortController: AbortController;
   data: Promise<PageData>;
 };
@@ -15,6 +16,7 @@ export type CachedPage = {
 export type CachedPreviewData = {
   previewAbortController: AbortController;
   preview: Promise<CachedPreview>;
+  previewContainerSize: number;
 };
 
 export type CachedPageData = {
@@ -80,6 +82,7 @@ export class PageCache {
     return {
       previewAbortController,
       preview,
+      previewContainerSize: this.containerSize,
     };
   };
 
@@ -128,12 +131,13 @@ export class PageCache {
   };
 
   refreshPreview = (pageNumber: number, preview: Promise<CachedPreview>, containerSize: number): CachedPage => {  
-    this.setContainerSize(containerSize);  
+    this.setContainerSize(containerSize);
     const originalPreviewData = this.previewCache.get(pageNumber);
     const newPreview = updatePreview(preview, containerSize);
     const newPreviewCache = {
       previewAbortController: originalPreviewData.previewAbortController,
       preview: newPreview,
+      previewContainerSize: this.containerSize    
     };  
 
     const data = this.dataCache.get(pageNumber);
--- a/frontend/src/js/components/PageViewer/PageCache.ts
+++ b/frontend/src/js/components/PageViewer/PageCache.ts
@@ -69,6 +69,8 @@ export class PageCache {
   };
 
   private onPreviewCacheMiss = (pageNumber: number): CachedPreviewData => {
+    console.log('onPreviewCacheMiss');
+
     const previewAbortController = new AbortController();
     const preview = authFetch(`/api/pages2/${this.uri}/${pageNumber}/preview`, {
       signal: previewAbortController.signal,
@@ -118,6 +120,18 @@ export class PageCache {
 
   getPage = (pageNumber: number): CachedPage => {
     const preview = this.previewCache.get(pageNumber);
+    const data = this.dataCache.get(pageNumber);
+
+    return {
+      ...preview,
+      ...data,
+    };
+  };
+
+  getPageAndRefreshPreview = (pageNumber: number): CachedPage => {
+    const preview = this.previewCache.getAndForceRefresh(pageNumber);
+
+    // TODO: we may need to refresh the data too, if we need a new server call to get new highlight positions
     const data = this.dataCache.get(pageNumber);
 
     return {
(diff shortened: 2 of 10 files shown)


frontend/src/js/components/PageViewer/VirtualScroll.tsx
```diff
@@ -180,23 +215,25 @@ export const VirtualScroll: FC<VirtualScrollProps> = ({
...
-            />
-          </div>
-        ))}
+            <div
+              key={page.pageNumber}
+              style={{
+                top: (page.pageNumber - 1) * pageHeight,
+                transform: `rotate(${rotation}deg)`,
+                left: `${scale > 1 ? '0' : ''}`,  
+              }}
```
marjisound, line 224: I had to set the left to 0 in here, because when the pages are landscape the page left side could go out of the container box
ezgif-5-80722b1ef1

frontend/src/js/components/PageViewer/VirtualScroll.tsx
```diff
@@ -45,23 +46,60 @@ export const VirtualScroll: FC<VirtualScrollProps> = ({
...
+        return newPages;
+      })
+      }, 500),
+    []
+  );
+
+  useEffect(() => {
+    debouncedRefreshPreview(pageCache, containerSize);
+  }, [pageCache, containerSize, debouncedRefreshPreview]);
+
```
marjisound, line 102: This useEffect is used to detect changes on container size. Container size changes when use zooms in or zooms out. Debounce is used to avoid multiple page re-rendering when user clicks zooms in/out multiple times

frontend/src/js/components/PageViewer/VirtualScroll.tsx
```diff
@@ -147,30 +185,47 @@ export const VirtualScroll: FC<VirtualScrollProps> = ({
...
+  }, [pageHeight, focusedSearchHighlight, scale]);
+
+  const getCachedPage = useCallback((pageNumber: number) => {
+    const cachedPage = pageCache.getPage(pageNumber);
+    if (cachedPage.previewContainerSize === pageCache.containerSize){
+      return cachedPage;
+    } else {
+      return pageCache.refreshPreview(pageNumber, cachedPage.preview, containerSize);
+    }
+  }, [pageCache, containerSize]);
```
marjisound, line 213: This function checks if the preview for a specific page number in cache is for the relevant container size or not. If not, a refresh preview is triggered

frontend/src/js/components/PageViewer/Page.tsx
```diff
@@ -46,7 +48,14 @@ export const Page: FC<PageProps> = ({
           const node = containerRef.current;
           if (node) {
             setScale(preview.scale);
-            node.appendChild(preview.canvas);
+
+            const oldCanvas = node.getElementsByTagName('canvas');
```
joelochlann, line 52: Since you're only expecting one element, you could maybe do `.querySelector` which will return `null` or the element

frontend/src/js/components/PageViewer/PageCache.ts
```diff
@@ -119,6 +130,40 @@ export class PageCache {
     };
   };
 
+  refreshPreview = (pageNumber: number, preview: Promise<CachedPreview>, containerSize: number): CachedPage => {  
+    this.setContainerSize(containerSize);
+    const originalPreviewData = this.previewCache.get(pageNumber);
+    const newPreview = updatePreview(preview, containerSize);
+    const newPreviewCache = {
+      previewAbortController: originalPreviewData.previewAbortController,
```
joelochlann, line 138: This makes me wonder what happens if you zoom while a page preview request is in flight... bit of an edge case but maybe worth testing

frontend/src/js/components/PageViewer/PageCache.ts
```diff
@@ -119,6 +130,40 @@ export class PageCache {
...
+
+    const newPreviewData = this.previewCache.get(pageNumber);
+
+    return {
+      ...newPreviewData,
+      ...data,
+    };
+  };
+
+  getPageAndRefreshPreview = (pageNumber: number): CachedPage => {
```
joelochlann, line 155: Seems like this one is unused now, can we get rid of it?

frontend/src/js/components/PageViewer/PageCache.ts
```diff
@@ -119,6 +130,40 @@ export class PageCache {
     };
   };
 
+  refreshPreview = (pageNumber: number, preview: Promise<CachedPreview>, containerSize: number): CachedPage => {
```
joelochlann, line 133: 👌 This is a much nicer approach than the re-requesting of the preview that we ended up with on Friday!

Is there a way - in the name of the function, or a comment - to make it clear that this does *not* trigger a new request, but only a re-render? Because it shares the "refresh" name with `getPageAndRefreshHighlights`, which does trigger a new request.

frontend/src/js/components/PageViewer/VirtualScroll.tsx
joelochlann: I'm not totally sure this is the right approach. Would need to pair on it but I feel like the right solution is keep the page centred, but be able to scroll leftwards if it's gone off the edge. There are lots of intermediate wrapping divs though so not quite sure what the right styling is to achieve that!

frontend/src/js/components/PageViewer/VirtualScroll.tsx
itsibitzi: Agreed, it's reasonable that a user might zoom in, then zoom out. So having the cache keyed off the number and the zoom level is reasonable!

I also think we should continue to display the page in a lower resolution until the new resolution has rendered, which will prevent the flicker when the page is very large.

joelochlann:
🎉 This is superb work @marjisound! The page viewer clientside code is notoriously complex so getting a working PR up so quickly is a huge accomplishment. Was also really fun pairing on it 🤓

I was going to say that most of the remaining things weren't blockers and could be fixed in subsequent PRs.

But... I just spotted one other thing. There is a noticeable flicker after loading, without touching any of the zoom functionality. Adding some logging in [here](https://github.com/guardian/giant/blob/ed00e9a0983a3068f0adc38d8351e2e8ab7dc3d0/frontend/src/js/components/PageViewer/Page.tsx#L48), it seems that we're now doing an unnecessary re-render.

Since this affects the render even when not using the zoom functionality, it probably needs to be resolved before merging. 😔

## before
Screenshot 2023-07-24 at 12 58 53

## after
Screenshot 2023-07-24 at 12 58 35

marjisound:
@joelochlann Thanks so much Jo for looking at the re-rendering issue. I had actually noticed that as well and was digging to find a fix for it. I'll definitely fix it and will let you know to look again when it's ready. I've also found a fix for the horizontal scrolling issue and will let you know when that's pushed as well (needs a bit of cleaning up). I'm off the rest of afternoon, so will be doing it tomorrow if that's OK?

frontend/src/js/components/PageViewer/VirtualScroll.tsx
marjisound: Thanks Jo, I created a function `setLeftScrollToMiddle` that would set the scroll to middle every-time the zooming is changed. I had to use a 1 second delay, because we need to calculate the scroll position after the new canvases have been created and we know the final scrollWidth. What do you think?

https://github.com/guardian/giant/pull/112/files#diff-5080e96f143544d1f02e141bb59b417dc6f40b6cc8c9693bdd75688c2c913b97R103

frontend/src/js/components/PageViewer/VirtualScroll.tsx
marjisound: Agree with both of you. This would be a must have optimisation. So 2 optimisations we need are:

1. update caching logic in order to keep hold of page previews for different scales in the cache rather than replacing the page preview 
2. Display the new preview after the pdf rendering has completed (For this we need to use the RenderTask.promise)

I had a chat with Jo and he agreed that the option 1, could be done in a separate PR. Would that be OK with you @itsibitzi ?

For option 2 @itsibitzi , I think it'd be better to focus on it in a separate PR, unless if you think this is necessary before we merge this PR?

frontend/src/js/components/PageViewer/PageCache.ts
marjisound: Great suggestion 👍

frontend/src/js/components/PageViewer/Page.tsx
marjisound: great suggestion 👍

frontend/src/js/components/PageViewer/PageCache.ts
marjisound: 👍

frontend/src/js/components/PageViewer/PageCache.ts
marjisound: Thanks Jo, I changed it to be reRenderPagePreview, is it good?

frontend/src/js/components/PageViewer/VirtualScroll.tsx
```diff
@@ -45,23 +46,59 @@ export const VirtualScroll: FC<VirtualScrollProps> = ({
...
+          ); 
+          return {
+            pageNumber: page.pageNumber,
+            getPagePreview: refreshedPage.preview,
+            getPageData: refreshedPage.data,
+          }
+        });
+
+        return newPages;
+      })
```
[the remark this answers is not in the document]
marjisound: Does it look better now?

frontend/src/js/components/PageViewer/PageCache.ts
joelochlann: Great!

frontend/src/js/components/PageViewer/VirtualScroll.tsx
joelochlann: This is a tricky one! I've had a bit more of a think and play around with this. Separating out the two issues:

### Overflow
`left: 0` isn't ideal because it means the page is no longer centred when at normal zoom, which is a regression from existing functionality.

Something was bothering me about the fact that you had to do this, because there should definitely be a way to make the `.scrollContainer` provide a horizontal scrollbar the respects the width of its children, e.g. https://jsfiddle.net/14fdgea0/

After playing around, I realised it's the flex layouts combined with absolute positioning that is [adding confusion](https://css-tricks.com/flexbox-and-absolute-positioning/). I've found that removing `display: flex` on the `.pages` element makes it behave correctly with respect to overflow.

The thing is, we're using flex to horizontally centre the pages so with that solution, we then end up back at the pages no longer being centred.

So the right solution will be to remove the absolute positioning and do everything with flex, or to remove flex and do the horizontal centre with a different technique (e.g. translate: https://medium.com/front-end-weekly/absolute-centering-in-css-ea3a9d0ad72e). I think the latter may be easier!

### Scroll
Using the 1 second delay is a nice way to prove out the logic (without getting bogged down in how to make it happen after render), but I don't think this would be a great user experience to move the page around under them. 

And thinking about it a bit more, centring the scrollbar is not actually desired behaviour while zooming: if you're trying to zoom to look at something in more detail, you want it to remain in view, and not lose your bearings by having the scrollbar yanked away to somewhere else.

Fortunately, I think the right solution to the overflow issue might make the scrollbar issue irrelevant: the browser's default behaviour should give us what we want, once we are able to get rid of the `left: 0`

frontend/src/js/components/PageViewer/VirtualScroll.tsx
joelochlann: I'm happy with both of those to be done in a separate PR

joelochlann:
Nice work fixing the double-render issue! 🎉

It seems to be working well when zooming into the first page of a document.

However, something is off with the vertical scroll position that becomes evident when you zoom into a high-numbered page on a multi-page doc.

Here, I zoom into page 28, but end up on page 16:

https://github.com/guardian/giant/assets/5122968/b99fc040-8e4e-4429-94b3-15e6d63d305b

joelochlann:
Oh, I guess it's just that the scrollbar needs to be adjusted because the total height of its container has changed, i.e. it was at pixel 1000 of 1000 pixel total height (end of doc), but then on zoom it's at pixel 1000 of a 2000 pixel total height (middle of doc)

frontend/src/js/components/PageViewer/VirtualScroll.tsx
joelochlann: **also...** I've realised that this issue isn't actually one that you created with the zoom functionality. It already existed due to the weird combo of flexbox and absolute positioning, but the zoom made it more obvious.

For instance, on PROD right now if you shrink the window, the left of the doc is cut off, even though the scrollbar is at the far left!

@itsibitzi do you have any memories of why you used absolute positioning with flex?

Screenshot 2023-07-25 at 12 42 38

frontend/src/js/util/LruCache.ts
```diff
@@ -85,6 +85,16 @@ export class LruCache<K extends string | number, V> {
   };
 
   replace = (k: K, v: V): void => {
-    this.addToCache(k, v);
+    this.entries[k] = v;
+    this.bumpRecency(k);
+
+    if (this.recencyQueue.length > this.maxEntries) {
```
joelochlann, line 91: theoretically we shouldn't need this check because replace doesn't increase the length of the queue

marjisound: Oh yes, so we shouldn't need to shift the queue at all, right?

frontend/src/js/util/LruCache.ts
```diff
@@ -85,6 +85,16 @@ export class LruCache<K extends string | number, V> {
   };
 
   replace = (k: K, v: V): void => {
-    this.addToCache(k, v);
+    this.entries[k] = v;
```
joelochlann, line 88: I think we should check before modifying the cache that the entry is indeed already in there. Otherwise I think `bumpRecency` might blow up because it assumes `k` is in the recency queue. (And if it doesn't, we'd end up with the recency queue and the cache out of sync again).

Alternatively, you could continue using `addToCache` but check whether `k` was already in the cache before adding it to the recency queue.

marjisound: Great point @joelochlann  I put a condition within the replace itself to check if the cache already exists and based on that it would either use `bumpRecency` or just go with adding the cache using `addToCache`.  What do you think?

joelochlann: Looks good! Maybe just worth a `console.warn` in the unexpected case

frontend/src/js/util/LruCache.ts
```diff
@@ -85,6 +85,11 @@ export class LruCache<K extends string | number, V> {
   };
 
   replace = (k: K, v: V): void => {
-    this.addToCache(k, v);
+    if (this.entries[k]) {
+      this.entries[k] = v;
+      this.bumpRecency(k);
+    } else {
+      this.addToCache(k, v);
```
joelochlann, line 92: let's `console.warn` a message here, since it might indicate a bug if someone is calling `replace` but the thing isn't actually in the cache already

joelochlann (approved):
Great work! Have had a play around and all looks good, excluding the bits that have been carded up in Trello already

philmcmahon (approved):
I've had a much less thorough review than Joe but have looked through the code changes and had a play on playground - looks great!

PR #114, opened 2023-08-07
Adding retry in client for upload failures

marjisound:
## What does this change?
This PR adds a retry (up to 3 retries) on file upload requests if they fail for any reason.  

### Delay between retries
There's a delay (gap) between the failure time and the retry. This delay starts at 1 second * random number between 0-1, and it increases on every retry. This increase is called **exponential backoff** which is a recommended strategy on retry mechanism because the more retries that fail, there's bigger chance that we're dealing with an outage that could take time, so we don't want to put too keep calling the server if it's experiencing an outage that could last longer. 

### Random time factor in delay between retries
There's also a randomness added to the backoff time in order to ensure that in the time of outage, if several clients are calling the server and encountering the error at the same time, they won't all re-send the request the server at the same time. 

Some client side logging have been added just for troubleshooting, they will be removed.

### DIFF
--- a/frontend/src/js/util/auth/authUploadWithProgress.ts
+++ b/frontend/src/js/util/auth/authUploadWithProgress.ts
@@ -18,9 +18,6 @@ export default function authUploadWithProgress(
     const retryInitialCount = 0;
 
     xhr.upload.onprogress = (e) => {
-      console.log(
-        `progress: loaded: ${e.loaded} from total: ${e.total} for url: ${url}`
-      );
       if (e.lengthComputable) {
         if (onProgress) {
           onProgress(e.loaded, e.total);
@@ -62,9 +59,6 @@ const processRequest = (
       resolve(xhr.response);
     } else {
       if (retryCount < RETRY_MAX_LIMIT) {
-        console.warn(
-          `retrying because request failed due to ${xhr.responseText} - status: ${xhr.status}, retry: ${retryCount} for url: ${url}`
-        );
         retryRequest(
           retryCount + 1,
           xhr,
@@ -85,7 +79,6 @@ const processRequest = (
     }
   };
 
-  console.log(`sending request: ${url} - retry: ${retryCount}`);
   sendRequest(xhr, url, file, path, uploadId, retryCount, workspace);
 };
 
@@ -102,7 +95,6 @@ const retryRequest = (
 ) => {
   const limit = retryCount ? Math.pow(2, retryCount - 1) * 1000 : 0;
   const pause = Math.random() * limit;
-  console.log(`delaying the next retry by ${pause / 1000} s - for url: ${url}`);
 
   setTimeout(() => {
     processRequest(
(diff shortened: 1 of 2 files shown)


frontend/src/js/util/auth/authUploadWithProgress.ts
```diff
@@ -1,57 +1,152 @@
...
+  xhr.setRequestHeader("X-PFI-Upload-Id", uploadId);
+  if (workspace) {
+    xhr.setRequestHeader("X-PFI-Workspace-Id", workspace.workspaceId);
+    xhr.setRequestHeader(
+      "X-PFI-Workspace-Parent-Node-Id",
+      workspace.parentNodeId
+    );
+    xhr.setRequestHeader("X-PFI-Workspace-Name", workspace.workspaceName);
+  }
+  xhr.setRequestHeader("X-PFI-Last-Modified", file.lastModified.toString());
```
philmcmahon, line 150: Would it be worth setting an extra request header here e.g. "X-PFI-Retry-Count"? I think we might be able to pick that up in the request logs from the server so if we get a user complaint we can see whether the retry mechanism was triggered

frontend/src/js/util/auth/authUploadWithProgress.ts
```diff
@@ -1,57 +1,152 @@
...
-            xhr.setRequestHeader('X-PFI-Workspace-Parent-Node-Id', workspace.parentNodeId);
-            xhr.setRequestHeader('X-PFI-Workspace-Name', workspace.workspaceName);
+  return new Promise((resolve, reject) => {
+    const xhr = new XMLHttpRequest();
+    const retryInitialCount = 0;
+
+    xhr.upload.onprogress = (e) => {
+      console.log(
+        `progress: loaded: ${e.loaded} from total: ${e.total} for url: ${url}`
+      );
```
philmcmahon, line 23: in my sample size of 1, there was some progress on the file (99%) before it appeared to stop, then there was a long pause until it hit the NS_ERROR_NET_INTERRUPT and failed before triggering the retry. If we wanted to trigger the retry sooner we could probably detect the stopped progress, though as this is a fairly intermittent bug I'm not sure it's worth the extra complexity - will just mean users have to wait the full minute before the retry kicks in

philmcmahon:
This looks great and appears to solve our problem! I suggest removing the logging that happens on the 'happy' path but keep the error logging. Then just one comment on adding an extra header above otherwise. I think this is ready to merge!

frontend/src/js/util/auth/authUploadWithProgress.ts
marjisound: good idea, done

frontend/src/js/util/auth/authUploadWithProgress.ts
marjisound: Interesting point. During the long pause, were you still getting progress logs? I'm just wondering if the progress handler periodically get triggered no matter if there is any progress or not OR if the handler only gets triggered while there is progress on the upload

philmcmahon: No I didn't get any progress logs during the pause - it just appeared to stop, I think it only gets triggered for actual progress. So if we wanted to detect 'no progress' I think we'd have to use some kind of setTimeout which checks what the progress is and alarms if it hasn't changed. I think too much complexity to add at this point though! It would speed up the retry but hopefully there won't be that many errors

philmcmahon (approved):
Looks great

shrek and donkey doing a retry

PR #115, opened 2023-08-10
Replace Akka library with Pekko equivalents wherever possible

philmcmahon:
## Background
In September 2022 Lightbend, the company who own and steward Akka, [announced](https://www.lightbend.com/blog/why-we-are-changing-the-license-for-akka) a licensing change from [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0.html) to the [Business Source License (BSL) v1.1](https://www.lightbend.com/akka/license). There is a limited grant made also for usage within the Play Framework.

BSL is a source-access licence, which allows non-production use of the software. After 3 years, the licence changes to Apache 2.0. Production use by 'early-stage' companies (less than $25m annual revenue) is also allowed.

Lightbend has provided a [FAQ page](https://www.lightbend.com/akka/license-faq) on the change with more detail. 

The new licence applies to Akka 2.7 on; version 2.6 remains under Apache 2.0 and Lightbend have promised to provide CVE fixes until September 2023.

Production-use licences can be purchased on an annual basis, and pricing is described [here](https://www.lightbend.com/akka#pricing). The pricing is very expensive and we (the Guardian) do not anticipate paying for usage.

## What does this change?
This change aims to remove all uses of akka from Giant, instead making use of [pekko](https://pekko.apache.org). I've largely followed the [migration guide](https://pekko.apache.org/docs/pekko/current/project/migration-guides.html), with some inspiration from [the same work being done in AMIgo](https://github.com/guardian/amigo/pull/1206)

The main changes are:
- swapping out the akka dependency
- updating a load of imports
- creating a new pekkoActorSystem. Play provides an actor system so we now have two, which isn't recommended by akka as actor systems are [heavyweight](https://doc.akka.io/docs/akka/current/general/actor-systems.html) but there's no real alternative here
- documenting all remaining uses of akka

Note that if https://github.com/playframework/playframework/pull/11847 became a thing we could further remove akka from Giant, but for now the bits integrating with play still need to use akka.

## How to test
Tested locally and on playground. This is quite a significant change to a critical part of giant so we should monitor carefully

We save mega $ on akka licenses 

pekko

### DIFF
--- a/backend/app/AppComponents.scala
+++ b/backend/app/AppComponents.scala
@@ -1,5 +1,7 @@
-import akka.actor.CoordinatedShutdown
-import akka.actor.CoordinatedShutdown.Reason
+import org.apache.pekko.actor.{ActorSystem => PekkoActorSystem, CoordinatedShutdown => PekkoCoordinatedShutdown}
+import akka.actor.{CoordinatedShutdown => AkkaCoordinatedShutdown}
+import org.apache.pekko.actor.CoordinatedShutdown.{Reason => PekkoReason}
+import akka.actor.CoordinatedShutdown.{Reason => AkkaReason}
 import cats.syntax.either._
 import com.gu.pandomainauth
 import com.gu.pandomainauth.PublicSettings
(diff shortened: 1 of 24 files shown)


backend/app/AppComponents.scala
```diff
@@ -57,20 +59,29 @@ class AppComponents(context: Context, config: Config)
...
+  applicationLifecycle.addStopHook(() => {
+    logger.info("Shutting down pekko")
+    pekkoActorSystem.terminate()
+  })
+
   override def httpFilters: Seq[EssentialFilter] = {
     super.httpFilters.filterNot(disabledFilters.contains) ++ Seq(
       new AllowFrameFilter,
+      //Note: still using akka (instead of pekko) materializer from Play as both filters extend Play's Filter, so the types need to match
       new RequestLoggingFilter(materializer),
```
philmcmahon, line 74: could we remove the materializer entirely? does play have a default?

marjisound: It seems that DevEx has also kept the the materializer for their HstsFilter class

marjisound (approved):
Looks great, I tested it locally as well and works all fine 👏

PR #117, opened 2023-08-15
Setup postgres and insert ingestion events to database

marjisound:
Co-Authored-By: Philip McMahon philip.mcmahon@guardian.co.uk
Co-Authored-By: Zeke Hunter-Green <zeke.huntergreen@guardian.co.uk>

## What does this change?
This PR adds the backend implementation for setting up the postgres client and inserting ingestion events and blob metadata to the postgres database. 

For now these are the event types that are inserted but this list could be expanded in the future if needed:
`HashComplete, BlobCopy, ManifestExists, MimeTypeDetected, IngestFile, InitialElasticIngest, RunExtractor`

The database credentials is retrieved from aws secret manager. If for whatever reason the secrets can not be retrieved there won't be any failure in the application and we just log the reason of it. 

Currently we are only letting the event logging  happen for the `pfi-giant` & `pfi-playground` stacks. Any other stack wouldn't do any event logging. 

### Database migration
We added the `migrate-db-local`, `migrate-db-code` & `migrate-db-prod` scripts to run the db migration locally. But we will probably enhance the infra to automate this in the future.

Relevant PRs that have already been merged:
- https://github.com/guardian/giant/pull/116 This PR adds the infra for the database (CDK)
- https://github.com/guardian/investigations-platform/pull/444 Adds database access security group to the frontend & worker launch configurations
- https://github.com/guardian/investigations-platform/pull/445 Adds a secret manager VPC endpoint and secret manager policy

## Why
Sometimes while ingesting data into giant, there is failure in some files but we don't have enough details on why the ingestion failed and in what stage. This changes helps to improve the observability of Giant and keep a record of everything that happens to a file throughout its ingestion journey. 

## How to test

Do a new ingestion and check the events in `ingestion_events` & `blob_metadata` tables in giant database.

PR #118, opened 2023-08-22
Add a UI to show ingestion events

philmcmahon:
## What does this change?
Following https://github.com/guardian/giant/pull/117, this PR adds an API endpoint and associated client side code to display a dashboard of ingestion events.

Currently there are two dashboards - 'my uploads' and 'all ingestion events', the latter of which can only be seen by admins. The former allows a user to explore uploads they have made to workspaces, whilst the latter is intended for admins to be able to explore all the ingestion events that have occurred to e.g. help debug a user who has reported a particular issue.

However, pending some work to auto-delete ingestion events after a period of time, at present the permissions are more locked down such that you can only see an ingestion dashboard if you already have permissions to view a particular collection. Admins will have to add themselves to particular collection before trying to debug issues with it.

The most complex part of this work is the new getEvents query in PostgresClient.scala. This was initially written by @zekehuntergreen - with the main complexity trying to get a one-blob-per-row view with all the relevant status updates for the different extractors run on a blob, taking account of the fact that a file can be ingested multiple times.

The frontend makes use of [Elastic UI](https://elastic.github.io/eui/) - there's probably an open question as to how heavily we want to start relying on EUI in giant due to the licensing concerns, but this was a quick way to get a nice table going and, as it's buried in the settings, I'm not too concerned about it looking a big jarring compared to the rest of giant's UI. An alternative  would be to use something like https://chakra-ui.com/getting-started/with-react-table

Right now I think the UI still needs a lot of work but has some basic useful elements. The aim of this PR is to provide a building block from which individual pieces of work can be done. For example, there's currently no rendering of errrors in the UI, which feels like it might be useful.

Missing UI features:
 - Paging
 - Search
 - Store current dashboard view in UI so refresh doesn't reset everything (right now refresh is the only way to get updates)
 - Much more - we should do an ingestion and work out what would be useful to see
 - auto update (particularly annoying given that when you refresh you lose the page you were on)

It looks like this:
Screenshot 2023-08-22 at 11 33 42
Screenshot 2023-08-22 at 11 33 13

## How to test
Tested on playground, seems to work ok - obviously with lots of missing data

As these dashboards are fairly hidden away the main concern is security - which is discussed above


backend/app/commands/IngestFile.scala
```diff
@@ -31,11 +31,14 @@ class IngestFile(collectionUri: Uri, ingestionUri: Uri, uploadId: String, worksp
       ingestion <- getIngestion()
 
       fileUri = FingerprintServices.createFingerprintFromFile(temporaryFilePath.toFile)
+      workspaceEvent = workspace.map(w =>
+        IngestionEvent.workspaceUploadEvent(fileUri, ingestionUri.value, w.workspaceName, EventStatus.Started))
+      _ = workspaceEvent.foreach(ingestionServices.recordIngestionEvent)
```
philmcmahon, line 36: document this

frontend/src/js/components/IngestionEvents/IngestionEvents.tsx
```diff
@@ -0,0 +1,221 @@
...
+    {collectionId, ingestId, workspaces, breakdownByWorkspace}: {
+        collectionId: string,
+        ingestId?: string,
+        workspaces: WorkspaceMetadata[],
+        breakdownByWorkspace: boolean
+    }) {
+    const [blobs, updateBlobs] = useState<BlobStatus[] | undefined>(undefined)
+
+    const ingestIdSuffix = ingestId && ingestId !== "all" ? `/${ingestId}` : ""
+    const [tableData, setTableData] = useState<IngestionTable[]>([])
```
philmcmahon, line 161: maybe put state together

frontend/src/js/components/IngestionEvents/IngestionEvents.tsx
```diff
@@ -0,0 +1,221 @@
...
+}
+
+function IngestionEvents(
+    {collectionId, ingestId, workspaces, breakdownByWorkspace}: {
+        collectionId: string,
+        ingestId?: string,
+        workspaces: WorkspaceMetadata[],
+        breakdownByWorkspace: boolean
+    }) {
+    const [blobs, updateBlobs] = useState<BlobStatus[] | undefined>(undefined)
```
philmcmahon, line 158: explicit | undefined isn't necessary

frontend/src/js/components/IngestionEvents/IngestionEvents.tsx
```diff
@@ -0,0 +1,221 @@
...
+    return {
+        workspacesMetadata: state.workspaces.workspacesMetadata,
+        currentUser: state.auth.token?.user,
+        collections: state.collections
+    };
+}
+
+function mapDispatchToProps(dispatch: GiantDispatch) {
+    return {
+        getCollections: bindActionCreators(getCollections, dispatch),
```
philmcmahon, line 217: maybe document why we need this

frontend/src/js/components/IngestionEvents/IngestionEvents.tsx
```diff
@@ -0,0 +1,221 @@
...
+        </>
+}
+
+
+function mapStateToProps(state: GiantState) {
+    return {
+        workspacesMetadata: state.workspaces.workspacesMetadata,
+        currentUser: state.auth.token?.user,
+        collections: state.collections
+    };
```
philmcmahon, line 212: this is all wrong!

philmcmahon (commented):
Comments from review session with @marjisound

frontend/src/js/components/IngestionEvents/IngestionEvents.tsx
philmcmahon: I tried removing | undefined but it resulted in red underlining in intellij

backend/app/services/observability/Models.scala
```diff
@@ -73,31 +75,79 @@ object EventDetails {
...
 
-case class BlobMetaData(ingestId: String, blobId: String, path: String, fileSize: Long)
 
-object BlobMetaData {
-  implicit val blobMetaDataFormat = Json.format[BlobMetaData]
+case class ExtractorStatus(extractorType: ExtractorType, statusUpdates: List[ExtractorStatusUpdate])
+object ExtractorStatus {
+  implicit val format = Json.format[ExtractorStatus]
+
+  def parseDbStatusEvents(extractors: Array[String], extractorEventTimes: Array[String], extractorStatuses: Array[String]): List[ExtractorStatus] = {
```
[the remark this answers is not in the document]
philmcmahon: I have written a single test. It took ages. I hate tests!

marjisound (approved):
All looks good, wonderful job 🎉 👏

PR #119, opened 2023-08-30
Fix observability dashboard when no extractor events

philmcmahon:
## What does this change?
There is currently a bug the observability dashboard where if a file has been hashed but no extractor events have started then the whole dashboard crashes. This resolves that by adding a few ternaries.

At the same time I discovered an issue where fileSize isn't defined - currently in our postgres query there is no default, it can just be NULL. So I made it optional.

This PR also adds a 'key' property to all our lists/repeated divs to avoid the [unique key prop warning](https://sentry.io/answers/unique-key-prop/).

At the same time I fixed our itemId in the EUI table to include the ingest ID (no user facing benefit here, just seems more accurate).

There's also silly pedantic change to rename metaData to metadata.

## How to test
Tested on playground, it appears to fix the main bug described above.

Fewer blank and screens and client side errors on the ingest dashboard.

### DIFF
--- a/frontend/src/js/components/IngestionEvents/IngestionEvents.tsx
+++ b/frontend/src/js/components/IngestionEvents/IngestionEvents.tsx
@@ -106,20 +106,30 @@ const columns: Array<EuiBasicTableColumn<BlobStatus>> = [
         field: 'extractorStatuses',
         name: 'Extractors',
         render: (statuses: ExtractorStatus[]) => {
-            return <ul>
+            return statuses.length > 0 ? (<ul>
                 {statuses.map(status => {
                     const mostRecent = status.statusUpdates.length > 0 ? status.statusUpdates[status.statusUpdates.length - 1] : undefined
-                    const allUpdatesTooltip = <p><b>All {status.extractorType} events</b> <br /> {status.statusUpdates.map(u => <>{`${moment(u.eventTime).format("DD MMM HH:mm:ss")} ${u.status}`}<br/></>)}</p>
-                    return <li><EuiFlexGroup>
+                    const statusUpdateStrings = status.statusUpdates.map(u => `${moment(u.eventTime).format("DD MMM HH:mm:ss")} ${u.status}`)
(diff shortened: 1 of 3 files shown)


marjisound (commented):
I tested locally and it all looks great. But I have encountered an issue that I can't yet get my head around. Sometimes, the workspace is returned undefined for some blobs that are supposed to be within a workspace. And as a result, those records don't show up in the dashboard. And then if I keep refreshing the page, they might show up again. 

The questions that I haven't yet figured out:
- why are we getting None for the records that are supposed to have a workspace name
- Why are we not showing the records that don't have any workspace?

frontend/src/js/components/IngestionEvents/IngestionEvents.tsx
```diff
@@ -63,14 +63,23 @@ const getBlobStatus = (statuses: ExtractorStatus[]) => {
     return failures.length > 0 ? blobStatusIcons.completeWithErrors : inProgress.length > 0 ? blobStatusIcons.inProgress : blobStatusIcons.complete
 }
```
marjisound, line 64: I know this is not related to this PR but just pointing it out to see if we need to fix this. There are sometimes failures on a blob extractor because of previous uploads. For example, I upload a file and it failes to complete OcrMyPdf, but the I upload the file again, and it succeeds. The statuses are showing the list of all statuses for all the uploads for this file. And while the file was processed successfully in the second time, dashboard is  showing this record as a failed record. Is that intentional?
image

marjisound: We discussed this in a mobbing session. This is a user experience concern and we will decide later if it needs to behave differently

marjisound:
We discussed the issue with the missing records in a mobbing session and found a fix for it. The issue was in the sql query for the workspace name and fix is in this commit https://github.com/guardian/giant/pull/119/commits/a1c55141f5dc90c778ffc74ebc4ef0d5ef1d25d9

marjisound (approved):
Looks great 👍

PR #120, opened 2023-08-30
Purge observability events older than 2 weeks

philmcmahon:
## What does this change?
This introduces a 'purge' job to delete the observability events initially introduced in https://github.com/guardian/giant/pull/117 after 2 weeks. These events are intended to help debug/understand what's happened with an ingest, but are of limited or no utility otherwise, so there's no reason to keep them long term.

## How to test
Events older than 14 days should be automatically deleted. We don't actually have any events older than 14 days at the moment though, so this is what I propose to check the feature is working:
 - deploy to playground
 - search the logs for the word 'purged' - should include the number of events purged **it should be 0**
 - Deploy to PROD if happy with the above
 - (in a few weeks) check again - some events should have been deleted at some point. Also check the DB - there should be no events older than two weeks, but events within the last 2 weeks should be unaffected

Not keeping unnecessary data

If I've done something wrong here we could accidentally delete all our observability data. In this scenario we would have to re-run ingests to get the data back.

### DIFF
--- a/backend/app/AppComponents.scala
+++ b/backend/app/AppComponents.scala
@@ -96,7 +96,7 @@ class AppComponents(context: Context, config: Config)
     val blobStorage = S3ObjectStorage(s3Client, config.s3.buckets.collections).valueOr(failure => throw new Exception(failure.msg))
 
     val postgresClient = config.postgres match {
-      case Some(postgresConfig) =>  new PostgresClientImpl(postgresConfig)
+      case Some(postgresConfig) =>  new PostgresClientImpl(postgresConfig, eventRetentionDays = 14)
       case None =>
         logger.warn("Postgres config not found, using dummy postgres client!")
         new PostgresClientDoNothing
--- a/backend/app/AppComponents.scala
+++ b/backend/app/AppComponents.scala
@@ -96,7 +96,7 @@ class AppComponents(context: Context, config: Config)
     val blobStorage = S3ObjectStorage(s3Client, config.s3.buckets.collections).valueOr(failure => throw new Exception(failure.msg))
 
     val postgresClient = config.postgres match {
-      case Some(postgresConfig) =>  new PostgresClientImpl(postgresConfig, eventRetentionDays = 14)
+      case Some(postgresConfig) =>  new PostgresClientImpl(postgresConfig, eventRetentionDays = 28)
       case None =>
         logger.warn("Postgres config not found, using dummy postgres client!")
         new PostgresClientDoNothing
--- a/backend/app/AppComponents.scala
+++ b/backend/app/AppComponents.scala
@@ -33,7 +33,7 @@ import services.events.ElasticsearchEvents
 import services.index.{ElasticsearchPages, ElasticsearchResources, Pages2}
 import services.ingestion.IngestionServices
 import services.manifest.Neo4jManifest
-import services.observability.{ObservabilityCleaner, PostgresClientDoNothing, PostgresClientImpl}
+import services.observability.{ObservabilityPurger, PostgresClientDoNothing, PostgresClientImpl}
 import services.previewing.PreviewService
 import services.table.ElasticsearchTable
 import services.users.Neo4jUserManagement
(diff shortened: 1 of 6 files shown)


philmcmahon:
> Not sure about expiring this data. I think it's reasonable for an investigation to go on for longer than 14 days and a journalist finds an interesting file that hasn't been extracted/ingested properly.

The assumption I'm making is that for a larger investigation the ingest will still have developer support, and that we'd aim to ingest 100% of files that look even remotely interesting within 2 weeks, so there would be no need for a journalist to use the observability info. For workspace uploads the dashboard is more useful, but given that the only user action that can really be taken if they see something has failed is to re-upload, I was thinking that would be easy to do within the 2 week windwo

zekehuntergreen:
My vote is that we postpone merging this. In the short term I think it would be useful to keep observability data around indefinitely for dev / testing purposes.
In the long term, it seems like even if we aim to ingest 100% of files successfully, in practice there will be ingests that we accept are incomplete and it will be useful to preserve information about failures. 
If we get to a point where for performance reasons we want to limit the size of `ingestion_events`, maybe we could think about horizontally partitioning by `event_time`?

itsibitzi:
To expand on what I was thinking, the kinds of failure modes we see around extractions are highly varied, and extraction failure isn't always the result of an error (e.g. a PDF being encrypted). I don't think it's reasonable for a developer to run and ingestion then go through the whole thing checking all the errors, so we need a way of keeping that data so it's useful if we need it.

A journalist landing on a document that has partial extraction failures would feel much more confident if we could leverage the observability data to give them a hint as to why this happened. An issue right now is you get a big red box and no other information - which requires a developer to look into what happened. It would be super cool if a journalist could copy-pasta the URL and send it over to one of us and we could debug the issue for them. Right now that's hugely annoying, and impossible if the logs have expired.

It's also worth bearing in mind that it's pretty common for us to do an ingestion while the project is still being planned from a resourcing perspective in editorial, it could take more than 2 weeks for a journalist to be assigned.

Finally, it would be very useful to have detailed information about these issues so that we can perform aggregations about what are common issues during ingestion. This would help guide development efforts in improving quality.

itsibitzi:
@zekehuntergreen If we're looking at horizontal scaling we might want to look at something like ULIDs or Snowflake IDs so we can spread out the data across multiple nodes. A basic implementation of event time partitioning could end up with all the hot data (e.g. the most recent stuff) being on one node.

I think there are bigger/lower effort performance wins we can do with the schema as it stands.

philmcmahon:
Thanks for the comments. My original motivation for this work - which I should have included in the PR description - was a sort of security/privacy concern. The way the dashboard works at the moment is that Giant Admins are able to see all events via the UI. In practice this means that they can see the filenames and upload time of files users are uploading to their private workspaces. Whilst admins could achieve that anyway by adding themselves to a particular users' collection, the dashboard is a way of seeing what users are doing without any kind of traceability. 

Deleting everything after 2 weeks is maybe a bit of an extreme solution though. We could instead not auto delete events but make sure that when a file/ingestion is deleted we also delete any associated observability events.

I think that there is some benefit in Giant admins being able to see e.g. errors users are encountering when uploading files, so I am reluctant to lock down the observability dashboards to only collections the user has the right to see. 

Maybe we should just trust admins? Or maybe only show filenames for files in a collection the user has the rights to access - that way if someone was complaining about a problem with an upload any giant admin could probably find the problematic file in question and see any events without knowing what it actually is.

I think I would argue that beyond a certain window - even if it's a year - we might as well delete this data - I can't see us going back and looking at it beyond that point.

philmcmahon:
> We had a long discussion about this in planning and reached a verdict that we would keep data for 28 days for now, and if we come across a situation where we regret not storing the data for longer we can address it then.

This was not a correct interpretation! The actual plan was to anonymise the data. In https://github.com/guardian/giant/pull/120/commits/56c04f7d447622cc53411610af4563cd92683d98 I've updated the cleaner so that it doesn't delete ingest events, instead just replacing the ingest_id with an md5 hash of the ingest id. The ingest_id is likely to be something like "Philip McMahon Documents/Upload 2023..." so likely to include personal information. We continue to drop blob_metadata events which include the filename.

zekehuntergreen:
In a discussion on 8/9/2023, we decided to close this PR and to instead have the API endpoints redact file and ingestion names for collections that the user doesn't have access to after a certain period of time.
As a separate piece of work we'll delete ingest event data when their associated files or collections are deleted.
This will let us keep stats on ingestion failures long term, and keep deletion behavior of observability data consistent with deletion of their associated files in the file system and records in elastic search and neo4j.

PR #122, opened 2023-09-05
only show statuses relevant to the ingest

marjisound:
## What does this change?
This PR fixes the issue where the observability dashboard was showing all the statuses from all ingestions for a specific file. This change, would only show the statuses that are relevant for that particular ingestion. 

I also added an order by, so that when the page is loaded for the first time, the data is ordered with most recent at the top.

### Before 
image

### After
image


philmcmahon (approved):
Looks good to me 👍🏻

PR #123, opened 2023-09-06
Store current workspace, collection, and ingestion in URL of observability dashboards

zekehuntergreen:
## What does this change?
Stores the current workspace in the url of the 'My uploads' observability dashboard and current collection and ingestion in the url of the 'All ingestion events' dashboard. These values are used to set the initial values of their respective dropdowns on the dashboard pages.

Screenshot 2023-09-06 at 14 21 06

## Why
- after refreshing the page, the user won't lose the ingestion events they were searching for
- the ingestion events from a specific ingestion can be sent between developers or from a user to a developer


philmcmahon:
One comment just FYI - not a blocker. We have the underscore library available which has https://underscorejs.org/#sortBy - can be useful for more complex sort operations

PR #124, opened 2023-09-06
Sort collections and ingestions in observability dashboard

zekehuntergreen:
## What does this change?
- order collections in dropdown alphabetically
- order ingestions alphabetically
  - "All ingestions" is the first rather than the last item

Screenshot 2023-09-06 at 14 49 49

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -44,7 +44,6 @@
         "semantic-ui-css": "^2.4.1",
         "semantic-ui-react": "^0.87.3",
         "typescript": "4.5.3",
-        "underscore": "^1.13.6",
         "uuid": "^3.4.0"
       },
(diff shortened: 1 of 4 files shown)


philmcmahon (approved):
sorting

philmcmahon:
Sorry, we actually already have a library called 'lodash' which seems has a [superset of the features of underscore](https://www.geeksforgeeks.org/difference-between-lodash-and-underscore/) so we don't need to add underscore as well. My bad for linking to the underscore docs! 

Here's an example of the import:
https://github.com/guardian/giant/blob/15cdfd94bafdc005f45384fcf6127361126f3ce7/frontend/src/js/components/IngestionEvents/IngestionEvents.tsx#L119

PR #127, opened 2023-09-08
Increase max size limit to 250MB

philmcmahon:
## What does this change?
In https://github.com/guardian/pfi/pull/839 we set the file upload limit to 100MB. This PR bumps that further. As  far as I can tell, there was no particular reason to limit workspace uploads to 100MB. We have had at least one user need to upload a file larger than 100MB, this change would have helped them. 

As a follow up piece of work, we should expose the file size limit in the UI and warn users when they try to upload a file above the limit.

## How to test
Tested on playground with a 190MB file

There could be performance implications of allowing such a large file to be uploaded via the UI. Hopefully the new dashboards should allow us to pick those up as and when they occur

### DIFF
--- a/backend/conf/application.conf
+++ b/backend/conf/application.conf
@@ -5,6 +5,7 @@ play {
     jwtResponseName = "X-Offer-Authorization"
   }
 
+  # When updating this value, also update MAX_FILE_UPLOAD_SIZE_MBYTES in UploadFiles.tsx
   http.parser.maxDiskBuffer = 250MB
 
   # Disable CSRF if there is an Authorization header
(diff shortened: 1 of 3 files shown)


philmcmahon:
Update: I have added some client side code to warn users about the limit:

Screenshot 2023-09-08 at 15 33 14

zekehuntergreen (approved):
tenor (1)

srbd (approved):
looks good to me

philmcmahon:
Outstanding gif game @zekehuntergreen @srbd

PR #128, opened 2023-09-13
delete resource if user is the only owner and if the file is deletable

marjisound:
## What does this change?
This PR adds a new endpoint to Giant api `POST /api/workspaces/:workspaceId/nodes/delete/:blobUri` which will
- delete the resource from neo4j (workspaceNode, file, blob & any descendants), S3 and elasticsearch 
   - if the only owner of the resource is the user who initiated the delete **AND** 
   - if the resource doesn't have any children (e.g. zip file that can have children files) 
- **Otherwise** return an error response

This PR also implements the UI functionality:
- show delete file in the context menu items when user right clicks on a file - only if the current user is the owner of the workspace
- when user clicks on delete file, show a modal window to ask the user if they are sure that they want to delete
- show the success or failure message on the modal after the deletion response is returned to the client

In this demo, user can only delete the files that are deletable and do not exist under any collection that's not owned by this user. 
ezgif-1-302b79d2cc

## How to test
This was tested localy and in Playground


backend/test/test/TestUserManagement.scala
```diff
@@ -172,6 +172,8 @@ class TestUserManagement(initialUsers: TestUserManagement.Storage) extends UserM
   override def setPermissions(username: String, permissions: UserPermissions): Attempt[Unit] =
     updateField(username, r => r.copy(permissions = permissions)).map(_ => ())
 
+  override def isOnlyOwnerOfBlob(blobUri: String, username: String): Attempt[Boolean] = ???
+
```
marjisound, line 176: This needs to get implemented

backend/test/controllers/api/WorkspaceSharingITest.scala
```diff
@@ -188,6 +188,17 @@ class WorkspaceSharingITest extends AnyFunSuite with Neo4jTestService with Elast
...
+  test("Barry cannot delete a file from Pauls workspace") {
+    val paulBlobAndNodeId = uploadAndAddFileToWorkspaceAsPaul()
+
+    asUser("barry") { implicit controllers =>
+      status(deleteFileFromWorkspace(
+        workspaceId = paulWorkspace.id,
+        blobUri = paulBlobAndNodeId.blobId
+      )) should be(500)
+    }
+  }
```
philmcmahon, line 200: nice!

frontend/src/js/actions/workspaces/deleteOrRemoveItem.ts
```diff
@@ -0,0 +1,31 @@
+import { deleteOrRemoveItem as deleteOrRemoveItemApi } from '../../services/WorkspaceApi';
+import { getWorkspace } from './getWorkspace';
+import { ThunkAction } from 'redux-thunk';
+import { AppAction, AppActionType, WorkspacesAction } from '../../types/redux/GiantActions';
+import { GiantState } from '../../types/redux/GiantState';
+
+export function deleteOrRemoveItem(
```
philmcmahon, line 7: should this have a different name as I think it is now just 'delete' which might fail with an error?

frontend/src/js/components/workspace/DeleteModal.tsx
```diff
@@ -0,0 +1,89 @@
...
+    
+    const modalTitle: Record<DeleteStatus, string> = {
+        unconfirmed: "Delete item?",
+        deleting: "Deleting item",
+        deleted: "Item deleted",
+        failed: "Failed to delete"
+    }
+
+    const modalMessage: Record<DeleteStatus, string> = {
+        unconfirmed: "Are you sure you want to delete this item? This action will permanently delete the resource.",
```
philmcmahon, line 23: maybe instead of 'resource' "...permanently delete the item from giant"

frontend/src/js/components/workspace/DeleteModal.tsx
```diff
@@ -0,0 +1,89 @@
...
+        deleting: "",
+        deleted: "This item has been successfully deleted.",
+        failed: "Failed to delete item. Please contact the admin to delete this item."
```
philmcmahon, line 26: maybe 'an administrator' rather than 'the admin'

philmcmahon (approved):
A few minor comments but this looks great to me! Event a test as well! Fantastic work.

delete bear

frontend/src/js/actions/workspaces/deleteOrRemoveItem.ts
marjisound: good spot, yes I just changed the name to `deleteResourceFromWorkspace` does it sound OK?

philmcmahon:
Updated changes look great 👍🏻

PR #130, opened 2023-09-19
Reprocess option added for files in the workspace view

srbd:
## What does this change?

Adds a reprocess option to the context menu in the workspace view - currently just for individual files. This hits the `/api/blobs/${itemId}/reprocess` endpoint which cleans up previous error states so the file is marked as processed if the reprocessing completes successfully.

## How to test

Tested on CODE

https://github.com/guardian/giant/assets/10672792/60026128-dd59-4e3c-88fe-048883ced12b


hoyla:
I'm very excited about this! Can we call it "Reprocess source file"? (I don't feel very strongly about this btw)

srbd:
> I'm very excited about this! Can we call it "Reprocess source file"? (I don't feel very strongly about this btw)

Done!

philmcmahon (approved):
Worth a test on playground but otherwise looks great! V useful button 

try again

PR #131, opened 2023-09-19
Move from JavaMail to JakartaMail

philmcmahon:
## What does this change?
Most of the work in this PR was done by @srbd 

JavaMail has reached [end of life](https://javaee.github.io/javamail/) and has been replaced by [jakarta mail](https://jakarta.ee/specifications/mail/2.0/jakarta-mail-spec-2.0). This PR introduces jakarta mail to Giant, including the Mbox provider JAR file which has to be manually compiled and checked in (see https://eclipse-ee4j.github.io/angus-mail/Mbox-Store)

Unfortunately, we've yet to succesfully ingest an mbox file using either jakarta mail or the latest version of java mail. Our assumption is that MBOX is a poorly defined format, so it's hard to get something that works for all MBOX files. The motivation for this PR isn't to actually fix anything :( but I have tested it with a normal .EML file which it appears to handle without any problems - so it shouldn't be making anything worse!

## How to test
Tested on playground, seems to exhibit the same behaviour as JavaMail..!

We'll benefit from future updates to jakarta mail


srbd (approved):
👍 LGTM thanks for finishing off

PR #133, opened 2023-09-19
Clean paths property of ingestion events in parser

philmcmahon:
## What does this change?
We have a bug at the moment where if there is no `path` available for a `BlobStatus` then the dashboard will crash. This PR fixes the problem at both ends - filtering out null paths both on the server and the client.

## How to test
Tested locally, think this is simple enough to go out as is!


marjisound (approved):
Looks great thank you

PR #136, opened 2023-09-25
Allow admins to view all observability data

philmcmahon:
## What does this change?
In future we hope that if we see an alarm or user error report with giant, we will be able to use the observability dashboard to help diagnose the issue. With the current implementation, it is only possible to view the observability data associated with an upload if the 'viewing' user has the rights to see the relevant collection. This PR changes the behaviour so that admin giant users are able to 'spy' on the observability data associated with any users' collections. 

Security wise, this doesn't represent a significant change as Giant Admins are already able to add themselves to any collection. We'd prefer they didn't do this though as it allows them to access all the files a user has uploaded.

The new logic is:
 - If you have view access to a collection, you can see all observability data
 - If you don't have view access but are a giant admin, then you can see all observability data, but everything older than 2 weeks is anonymised in the UI
 - Otherwise, you get a 404.

The motivation for the 2 week anonymisation is that, after 2 weeks, there's a very slim chance an admin user is looking for a specific file - instead it is more likely they are looking for e.g. what is the most common error to have occurred - in which case they shouldn't need to know specific filenames, workspace names etc.

There is an argument that we should just anonymise all data where the admin user doesn't have view rights over a collection - happy to change this depending what people think

## How to test
I've tested this on playground and locally by taking away access to my own collections, it looks like this:

Screenshot 2023-09-25 at 14 26 59

Admins are more quickly able to debug user upload errors

### DIFF
--- a/backend/app/controllers/api/IngestionEvents.scala
+++ b/backend/app/controllers/api/IngestionEvents.scala
@@ -4,7 +4,7 @@ import model.Uri
 import model.user.UserPermission.CanPerformAdminOperations
 import play.api.libs.json.Json
 import play.api.mvc.{Action, AnyContent}
-import services.observability.PostgresClient
+import services.observability.{BlobStatus, PostgresClient}
 import services.users.UserManagement
 import utils.attempt._
 import utils.controller.{AuthApiController, AuthControllerComponents}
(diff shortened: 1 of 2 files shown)


marjisound (approved):
Great work 👍

zekehuntergreen (approved):
redact

PR #138, opened 2023-09-25
fix: Exclude infinitely-ingesting blobs from observability query

zekehuntergreen:
## What does this change?
We've found a few cases of bugs where ingestion fails only to be retried minutes later, forever. These retries create thousands of rows in `ingestion_events` which makes the observability dashboard query timeout. This PR should fix the dashboard for ingests with problem files in an infinite retry cycle while we come up with a better retrying solution.

In a follow-up we plan to put blob ids which have more than 100 ingest events back into the dashboard table, with a helpful explanation as to what's going on.

## Accessibility

### DIFF
--- a/backend/app/services/observability/PostgresClient.scala
+++ b/backend/app/services/observability/PostgresClient.scala
@@ -141,6 +141,7 @@ class PostgresClientImpl(postgresConfig: PostgresConfig) extends PostgresClient 
               -- there is no index on extractorName but we aren't expecting too many events for the same blob_id/ingest_id
               AND blob_extractors.extractor = ingestion_events.details ->> 'extractorName'
               -- A file may be uploaded multiple times within different ingests - use group by to merge them together
+            WHERE ingestion_events.type = 'RunExtractor'
             GROUP BY 1,2,3
           )
           SELECT
--- a/backend/app/services/observability/PostgresClient.scala
+++ b/backend/app/services/observability/PostgresClient.scala
@@ -142,8 +142,8 @@ class PostgresClientImpl(postgresConfig: PostgresConfig) extends PostgresClient 
               AND blob_extractors.ingest_id = ingestion_events.ingest_id
               -- there is no index on extractorName but we aren't expecting too many events for the same blob_id/ingest_id
               AND blob_extractors.extractor = ingestion_events.details ->> 'extractorName'
-              -- A file may be uploaded multiple times within different ingests - use group by to merge them together
             WHERE ingestion_events.type = 'RunExtractor'
+            -- A file may be uploaded multiple times within different ingests - use group by to merge them together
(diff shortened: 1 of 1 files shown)


backend/app/services/observability/PostgresClient.scala
```diff
@@ -150,23 +157,24 @@ class PostgresClientImpl(postgresConfig: PostgresConfig) extends PostgresClient
...
+              blob_id,
+              ingest_id,
+              MIN(EXTRACT(EPOCH from event_time)) AS ingest_start,
+              MAX(EXTRACT(EPOCH from event_time)) AS most_recent_event,
+              ARRAY_AGG(details -> 'errors') as errors,
+              (ARRAY_AGG(details ->> 'workspaceName')  FILTER (WHERE details ->> 'workspaceName' IS NOT NULL))[1] as workspace_name,
+              (ARRAY_AGG(details ->> 'mimeTypes')  FILTER (WHERE details ->> 'mimeTypes' IS NOT NULL))[1] as mime_types
+            FROM ingestion_events
+            WHERE ingest_id LIKE ${if(ingestIdIsPrefix) LikeConditionEscapeUtil.beginsWith(ingestId) else ingestId}
+            AND blob_id NOT IN (SELECT blob_id FROM problem_blobs)
```
philmcmahon, line 171: Is postgres smart enough that if you put this filter here the CTE's are performant? I would have thought that you would need to repeat the `NOT IN (SELECT blob_id FROM problem_blobs)` in blob_extractors and extractor_statuses

backend/app/services/observability/PostgresClient.scala
```diff
@@ -110,30 +110,37 @@ class PostgresClientImpl(postgresConfig: PostgresConfig) extends PostgresClient
             *
             */
           val results = sql"""
-      WITH blob_extractors AS (
-      -- get all the extractors expected for a given blob
-       SELECT ingest_id, blob_id, jsonb_array_elements_text(details -> 'extractors') as extractor from ingestion_events
-       WHERE ingest_id LIKE ${if(ingestIdIsPrefix) LikeConditionEscapeUtil.beginsWith(ingestId) else ingestId} AND
-       type = ${IngestionEventType.MimeTypeDetected.toString}
+          WITH problem_blobs AS (
```
philmcmahon, line 113: maybe add a comment documenting this as it has that magic number 100 in it

philmcmahon (approved):
Just one speculative comment above, otherwise 👍🏻  

infinite journey

backend/app/services/observability/PostgresClient.scala
zekehuntergreen: good call! added to blob_extractors which will also exclude blobs in `problem_blobs` from `extractor_statuses`

PR #139, opened 2023-09-25
Add delete functionality in document viewer for non-admin users

marjisound:
## What does this change?
This PR adds delete functionality for all users within document viewer. However, the delete may only pass if 
- the file doesn't have children and
  - the file is within only 1 collection and the user has view access on the file
  - Or the file is within several collections that have been created by the requesting user

ezgif-5-abab464ff6

## How to test
This was tested both locally and in playground

### DIFF
--- a/backend/app/controllers/api/Blobs.scala
+++ b/backend/app/controllers/api/Blobs.scala
@@ -107,7 +107,7 @@ class Blobs(override val controllerComponents: AuthControllerComponents, manifes
     manifest.getCollectionsForBlob(blobUri).flatMap { collections =>
       // Here we check either of 2 followings:
       // if there's only 1 collection holding the blob and if the requesting user has view access to the collection
-      // OR if the requesting user is the only creator of the blob
+      // OR if the requesting user is the only creator of the blob e.g. if a user has uploaded the same file to multiple workspaces
       if ((collections.size == 1 && collections.head._2.contains(user.username)) || collections.forall(c => c._1.createdBy == Some(user.username))) {
         logAction(user, s"Deleting resource from Giant if no children. Resource uri: $blobUri")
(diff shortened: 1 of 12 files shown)


backend/app/controllers/api/Blobs.scala
```diff
@@ -88,13 +90,39 @@ class Blobs(override val controllerComponents: AuthControllerComponents, manifes
...
+    } else {
+      deleteForNoneAdmin(req.user, id).map(_ => NoContent)
     }
   }
+
+  private def deleteForNoneAdmin(user: User, blobUri: String): Attempt[Unit] = {
+    manifest.getCollectionsForBlob(blobUri).flatMap { collections =>
+      // Here we check either of 2 followings:
+      // if there's only 1 collection holding the blob and if the requesting user has view access to the collection
+      // OR if the requesting user is the only creator of the blob
```
philmcmahon, line 111: maybe extend this '...creator of the blob e.g. if a user has uploaded the same file to multiple workspaces'

backend/app/controllers/api/Blobs.scala
```diff
@@ -88,13 +89,39 @@ class Blobs(override val controllerComponents: AuthControllerComponents, manifes
...
+        val result = if (checkChildren) deleteResource.deleteBlobCheckChildren(id)
+        else deleteResource.deleteBlob(id)
+        result.map(_ => NoContent)
+      }
+    } else {
+      deleteForNoneAdmin(req.user, id).map(_ => NoContent)
     }
   }
+
+  private def deleteForNoneAdmin(user: User, blobUri: String): Attempt[Unit] = {
```
philmcmahon, line 106: minor - should be `deleteForNonAdmin`

philmcmahon (approved):
Looks great!

hide the evidence

PR #140, opened 2023-09-26
Remove redundant package-lock.json

philmcmahon:
All npm stuff lives in the frontend subdirectory, so we don't need a package-lock at the root. Also, having one here confuses the snyk GHA runner


marjisound (approved):
Lets go with yours  😆

PR #149, opened 2023-09-29
When a file fails to ingest, move to a 'dead letter' bucket

philmcmahon:
## What does this change?
We have seen an issue in giant where if a file fails the initial ('phase 1') ingestion step - i.e. any of the below processes fail:
 - hashing
 - copying to data bucket with hash as the key
 - mimetype detection with tikka
 - determining what extractors need running based off the mimetype
 - writing manifest to neo4j
 - writing to elasticsearch

....then the file will remain in the ingest bucket and be retried over and over again. This adds a lot of noise to the logs and observability dashboard and may result in increased AWS spend as workers try over and over again to ingest the same files.

This PR prevents this behaviour by removing ingest files from the ingest bucket and copying them to a 'dead letter' bucket. In order to get giant to have another go at ingesting those files, they will have to be uploaded again via the CLI or an API endpoint has been provided to copy the files back to the ingest bucket (this endpoint isn't very discoverable at the moment - arguably it should have an associated CLI option or a button somewhere, but it's reference in the logs and I think is better than nothing).

Note that this functionality only affects uploads via the CLI. User uploads will continue to fail immediately with an error message to the user if the initial ingest fails.

Depends on https://github.com/guardian/investigations-platform/pull/453 (already deployed)

## How to test
I tested this functionality on playground by stopping elasticsearch and then attempting a CLI ingest. At the same time I tested the 'retry dead letter' endpoint. 

At the moment, if an ingest fails because e.g. elasticsearch is down then it will be retried over and over again so that if elasticsearch recovers it will go through. In this new world this requires a manual step. Given that CLI uploads are already pretty manual this seems a worthwhile trade off

## Wikipedia on Dead Letter Mail
_Dead letter mail or undeliverable mail is [mail](https://en.wikipedia.org/wiki/Mail) that cannot be delivered to the addressee or returned to the sender. This is usually due to lack of compliance with postal regulations, an incomplete address and [return address](https://en.wikipedia.org/wiki/Return_address), or the inability to [forward the mail](https://en.wikipedia.org/wiki/Mail_forwarding) when both correspondents move before the letter can be delivered._
https://en.wikipedia.org/wiki/Dead_letter_mail

no post on sundays gif

### DIFF
--- a/backend/app/services/IngestStorage.scala
+++ b/backend/app/services/IngestStorage.scala
@@ -2,7 +2,6 @@ package services
 
 import java.io.InputStream
 import java.util.UUID
-
 import cats.syntax.either._
 import com.amazonaws.services.s3.model.S3ObjectSummary
 import model.{Languages, Uri}
@@ -13,6 +12,7 @@ import utils.attempt.{Failure, IllegalStateFailure, JsonParseFailure, UnknownFai
 import utils.aws.S3Client
 
 import scala.jdk.CollectionConverters._
+import scala.util.Try
 import scala.util.control.NonFatal
 
 trait IngestStorage {
@@ -81,13 +81,15 @@ class S3IngestStorage private(client: S3Client, ingestBucket: String, deadLetter
     }
   }
 
-  override def retryDeadLetters() = {
+  override def retryDeadLetters(): Either[UnknownFailure, Unit] = {
     Either.catchNonFatal {
       val result = client.aws.listObjects(deadLetterBucket)
       if (!result.isTruncated) {
         logger.info(s"Sending ${result.getObjectSummaries.size()} files from dead letter bucket to ingest bucket.")
-        result.getObjectSummaries.forEach{summary =>
-          client.aws.copyObject(deadLetterBucket, summary.getKey, ingestBucket, summary.getKey)
+        result.getObjectSummaries.forEach{ summary =>
+          Try(client.aws.copyObject(deadLetterBucket, summary.getKey, ingestBucket, summary.getKey))
+            // on success, clean up the file from the dead letter bucket
+            .map(_ => client.aws.deleteObject(deadLetterBucket, summary.getKey))
         }
       } else {
         val msg = "Too many dead letter files to resync via API. Try using aws cli e.g aws s3 sync s3://deadletterbucket s3://ingestbucket"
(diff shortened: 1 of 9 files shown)


backend/app/services/IngestStorage.scala
```diff
@@ -63,19 +65,51 @@ class S3IngestStorage private(client: S3Client, bucket: String) extends IngestSt
...
+            // on success, clean up the file from the dead letter bucket
+            .map(_ => client.aws.deleteObject(deadLetterBucket, summary.getKey))
+        }
+      } else {
+        val msg = "Too many dead letter files to resync via API. Try using aws cli e.g aws s3 sync s3://deadletterbucket s3://ingestbucket"
+        logger.error(msg)
+        throw new Error(msg)
+      }
     }.leftMap(UnknownFailure.apply)
   }
```
marjisound, line 100: I know this task should be very quick. But just wondering, do we not care if the metadata is copied before the actual file? Because if the file is copied first, and if it quickly be picked up by the worker before the metadata is copied, the worker may fail

marjisound (approved):
Looks great 👏 👏

backend/app/services/IngestStorage.scala
philmcmahon: Ohh good point, I had thought about this a little but just hoped s3 would be fast 🤦🏻 

I've done a bit more research - it looks like files are ingested only when the 'data' bit has been uploaded ([getNextBatch](https://github.com/guardian/giant/blob/main/backend/app/ingestion/phase2/IngestStorePolling.scala#L166) calls [list](https://github.com/guardian/giant/blob/4f61f5543df91057e9dec6c91e9ca7de966d040a/backend/app/services/IngestStorage.scala#L36) which only fetches the data prefix) 

So I've refactored this to do the metadata events first, which should mean that the reingestion only starts once the data bit has finished copying

philmcmahon: (changes are in https://github.com/guardian/giant/pull/149/commits/5a144231655409b4bf0861cfb7115e1a7e345003, I'll test on playground before merging)

marjisound: Amazing, really like the usage of the partition 👌

PR #151, opened 2023-09-29
Bump postgres to version 15

philmcmahon:
## What does this change?
There is currently a mismatch between the version of postgres we are running locally via docker and the version we are running in RDS. Rather than downgrade our local version I decided to upgrade to 15.3 in RDS. As we're using 15.3 locally already, we have a good amount of confidence that we won't have any compatability issues.

## How to test
This has already been deployed on CODE and doesn't appear to cause any issues. 

Deploy steps for PROD:
 - merge this
 - deploy via riffraff 

Queries that work locally should also work in AWS 🎉

### DIFF
--- a/postgres/cdk/lib/giant.ts
+++ b/postgres/cdk/lib/giant.ts
@@ -1,9 +1,6 @@
 import type { GuStackProps } from '@guardian/cdk/lib/constructs/core';
 import {
-	AppIdentity,
-	GuParameter,
 	GuStack,
-	GuStringParameter,
 } from '@guardian/cdk/lib/constructs/core';
 import { GuVpc, SubnetType } from '@guardian/cdk/lib/constructs/ec2/vpc';
 import type { App } from 'aws-cdk-lib';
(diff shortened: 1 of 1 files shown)


postgres/cdk/lib/giant.ts
```diff
@@ -53,8 +50,9 @@ export class Giant extends GuStack {
 				}),
 			},
 			engine: DatabaseInstanceEngine.postgres({
-				version: PostgresEngineVersion.VER_13,
+				version: PostgresEngineVersion.VER_15
 			}),
+            allowMajorVersionUpgrade: true,
```
marjisound, line 55: Just for my own understanding, what would happen if we don't use this field? `allowMajorVersionUpgrade`

philmcmahon: The upgrade would be blocked by RDS - it's like an extra safety feature. Once this is merged and the DB has been upgraded on PROD i think we should remove this line again so we don't accidentally upgrade!

PR #152, opened 2023-09-29
Add extra information/context to observability dashboard

philmcmahon:
## What does this change?
Initially the observability dashboard focussd on errors occuring during the extraction process. Errors can occur at other times during ingestion. This PR aims to expose those situations by adding some extra fields to the events API and rendering some new data. 

There is a new table showing pre-extraction ingestion events:
Screenshot 2023-09-29 at 15 04 12

Ingestion errors are now broken down based off the associated ingestion event:

Screenshot 2023-09-29 at 15 05 51

To make this work I've had to add a new 'IngestionEventStatus' field to the BlobStatus model. I think this makes sense. Extractor events are excluded for now as they are accounted for elsewhere (I actually think it would be useful to show them in this table, but it would involve quite a bit of extra work to get the extractorName included in ingestionEventStatus so one for a future bit of work)

The motivation for this work came off the back of work to deal with pre-extractor errors:
 -  https://github.com/guardian/giant/pull/146
 -  https://github.com/guardian/giant/pull/149 

With the dead letter queue we will no longer get millions of errors/events for files that fail the initial ingest. This PR helps make it easier to spot and debug them in the observability dashboard

## How to test
Tested on playground and locally. This PR depends on https://github.com/guardian/giant/pull/151 (the data type that postgres returns for EXTRACT EPOCH changes between versions 13 and 15)


frontend/src/js/components/IngestionEvents/IngestionEvents.tsx
```diff
@@ -172,10 +199,22 @@ const renderExpandedRow = (blobStatus: BlobStatus) => {
                 </p></>
             })}
         </div>
+        <h4>All ingestion events prior to extraction</h4>
```
zekehuntergreen, line 202: Might be worth putting a similar title above the first section e.g. "Extraction events"?

zekehuntergreen (approved):

frontend/src/js/components/IngestionEvents/IngestionEvents.tsx
philmcmahon: good shout, I've also reordered it so that the different sections are in order rather than having extractors first

PR #153, opened 2023-10-02
Migrate giant to github actions

philmcmahon:
## What does this change?
Adds a github actions workflow to build giant. 

The only change I had to make was to comment out this bit where we manually copy the compiled pdf.worker.min.js file as github actions appears to not have the same issue that teamcity did where it didn't follow the symbolic link associated with the file. 

Otherwise this is a fairly standard node/scala gha build

## How to test
I've tested this on playground by deploying a version generated by github actions, it seems to work (including the pdf worker!)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,5 +50,5 @@ jobs:
       - name: Run script/teamcity
         run: |
           LAST_TEAMCITY_BUILD=5000
-          export BUILD_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
-           ./scripts/teamcity.sh
+          export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $LAST_TEAMCITY_BUILD ))
+          ./scripts/teamcity.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,8 +8,7 @@ on:
   workflow_dispatch:
 
   push:
-    branches:
-      - "main"
+
 jobs:
   CI:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


scripts/teamcity.sh
```diff
@@ -21,7 +21,10 @@ popd
 cp -r frontend/build/* backend/public
 # Replace the symbolic link we use in dev with the actual file.
 # On Teamcity the JDeb build doesn't seem to follow the symbolic link while packaging, weirdly
-cp frontend/node_modules/pdfjs-dist/build/pdf.worker.min.js backend/public/third-party/pdf.worker.min.js
+# NOTE: On Github actions this seems to work, and in fact it will complain if you try and run the below line
```
marsavar, line 24: I can't comment on parts of this file that weren't modified in this PR, but  lines 4-11 in this script are made redundant by `actions/setup-node@v3` in the workflow so they can be removed I think

https://github.com/guardian/giant/blob/958316e75653ba701c2d2da746acb2c245a93c4a/scripts/teamcity.sh#L4-L11

scripts/teamcity.sh
marsavar, line 1: This should probably be renamed to `ci.sh` since it no longer runs in TeamCity.

.github/workflows/build.yml
```diff
@@ -0,0 +1,49 @@
...
+      # Configuring caching is also recommended.
+      # See https://github.com/guardian/actions-setup-node
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: .nvmrc
+
+      # Configuring caching is also recommended.
+      # See https://github.com/actions/setup-java
+      - name: Setup Java 8
```
marsavar, line 38:
```suggestion
      - name: Setup Java 11
```

marsavar (approved):
Looks good!
I think it's worth separating out the frontend and the backend into separate steps rather than having them in the same script; that would allow you to do some interesting stuff with caching (Happy to work on it during 10% 😄). But as a first step this seems to work fine.

PR #155, opened 2023-10-03
Show infinitely looping files in observability dashboard

zekehuntergreen:
Some file types fail to be ingested and ingestion is retried over and over in an infinite loop, inserting new rows into the observability database every few minutes. in #149 we stopped this from happening by moving failed files to a dead-letter bucket. 
Any files that have previously been in an infinite loop, and any files that cause this failure mode for other reasons that we haven't thought of in the future should be shown in the observability dashboard.

## Before
- files with more than 100 ingestion events are assumed to be stuck in an infinitely loop and, in order to improve performance, are excluded from the query that selects the data for the dashboard.

## After
- These problematic files are added back into the observability dashboard. In order to avoid slowing down the query, we don't array aggregate the ingestion events' errors (of which there could be thousands).
- Most of the dropdown in the dashboard for these files is replaced with a message explaining that there's an infinite loop and providing the blob id so that the problem can be investigated

Screenshot 2023-10-04 at 17 49 45

## Accessibility

### DIFF
--- a/backend/app/services/observability/Models.scala
+++ b/backend/app/services/observability/Models.scala
@@ -197,7 +197,8 @@ case class BlobStatus(
                        eventStatuses: List[IngestionEventStatus],
                        extractorStatuses: List[ExtractorStatus],
                        errors: List[IngestionErrorsWithEventType],
-                       mimeTypes: Option[String])
+                       mimeTypes: Option[String],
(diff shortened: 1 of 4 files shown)


marjisound (approved):
Looks good 👍

PR #156, opened 2023-10-04
Fix excel extractor failure on invalid dates

marjisound:
## What does this change?
This PR fixes a bug where an excel date field has an invalid date value,, and therefore can not be converted to a Java date. `DateUtil.getJavaDate(dateAsDouble)` returns null if the provided argument is not a valid date. So the ingestion of the file was failing due to calling `toString` on a null and was getting a `NullPointerException` exception. 

## How to test
Tested locally and in playground


zekehuntergreen (approved):
👍 looks good

PR #157, opened 2023-10-06
fix bug where todo attempt was not updated after fetchWork

marjisound:
## What does this change?
When worker calls fetchWork, the resources that have `TODO` relationships will be returned so the worker can execute the relevant extractors. As part of the fetchWork, the `attempts` property of the TODO relationship is supposed to get updated. For a strange reason that I didn't fully understand, the query was not updating the TODO attempts. Therefore, everytime the worker is triggered, the task that has already been attempted is returned by the query meaning it's stuck in an infinite loop. 

The fix in this PR does the 2 followings:
- Fix the query so that the todo attempts is updated 
- Add a `Extraction_Failure` relationship between the relevant resource and the extractor, so that it is known to us that this extractor is not supported. This way the UI also shows that the extractor for the file failed. 

 

## How to test
Tested locally

### DIFF
--- a/backend/app/extraction/Worker.scala
+++ b/backend/app/extraction/Worker.scala
@@ -56,8 +56,10 @@ class Worker(
               Attempt.Right((extractor, blob, ExtractionParams(ingestion, languages, parentBlobs, workspace)))
 
             case _ =>
-              logger.error(s"Unknown extractor $extractorName")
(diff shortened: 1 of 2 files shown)


itsibitzi (approved):
Nice one! 

Cypher is a bit of a mystery to me... Slightly disinclined to (re)learn it since I think we ultimately would like to move to PostgreSQL anyway

PR #160, opened 2023-10-16
handle unhandled exceptions in worker

marjisound:
## What does this change?
This PR is wrapping the worker logic in a try catch in order to handle the errors that couldn't be handled through the Attempt failure case. 

E.g. we were experiencing error from the neo4j driver `org.neo4j.driver.v1.exceptions.ServiceUnavailableException`, which put the worker schedular in a failed state and stopped triggering the worker on that instance. 

In Giant we are using the Attempt wrapper that can contain specific errors in the form of Failure. But if there's any error thrown that's not handled and wrapped in a Failure, it won't be recovered by the `recoverWith` in the WorkerScheduler and thus, the actorSystem schedular will stop working and triggering the worker. 

It might be good if this implementation is enhanced in the future and we use a proper messaging system such as kinesis stream to pick up tasks and process them. But for now, a quick fix would be to use a try and catch to handle the exceptions in the WorkerScheduler. 

## How to test
This was tested locally by throwing a neo4j ServiceUnavailableException, the error was logged and the schedular triggered the worker after the application interval time.

### DIFF
--- a/backend/app/services/WorkerScheduler.scala
+++ b/backend/app/services/WorkerScheduler.scala
@@ -20,15 +20,21 @@ class WorkerScheduler(actorSystem: ActorSystem, worker: Worker, interval: Finite
   }
 
   def go(): Unit = {
-    worker.pollAndExecute().map {
-      case completed if completed > 0 => go()
-      case _ => tryAgainIn(interval)
-    }.recoverWith {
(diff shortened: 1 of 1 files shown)


zekehuntergreen (approved):
Great investigation and fix 👏

PR #161, opened 2023-10-17
add timeout to ocrMyPdf process

marjisound:
## What does this change?
This PR is adding a timeout to `ocrMyPdf` process. This is because there's been occurrences where ocrMyPdf gets stuck in processing a file and doesn't return any exit code. As a result, the worker waits forever until the file processing is completed or errored. 

The timeout duration is chosen based on the number of pages within a file and is giving at most 1 minute for every page plus an overall 5 minutes to mitigate the risk of un-necessary timeout for files with low number of pages, e.g. for a pdf page with 10 pages, 15 (10 + 5) minutes is allocated for timeout. 

Thanks to @zekehuntergreen who helped with looking through the ingestion_events table to confirm that (1 minute * numberOfPages) + 5 minutes should be more than enough time to allocate for the timeout duration of ocrMyPdf.

## How to test
Tested locally and in playground once by manually setting a shorter timeout and handling the error. And also, by setting the timeout based on page numbers and upload a few files.

### DIFF
--- a/backend/app/utils/Ocr.scala
+++ b/backend/app/utils/Ocr.scala
@@ -60,6 +60,7 @@ object Ocr extends Logging {
 
   val TIMEOUT_DURATION_PER_PAGE = 1
   val DEFAULT_TIMEOUT_IN_MINUTES = 300
+  val TIMEOUT_DURATION_OVERHEAD = 5
 
   def invokeTesseractDirectly(lang: String, imageFileName: String, config: TesseractOcrConfig, stderr: OcrStderrLogger): String = {
     val cmd = s"tesseract $imageFileName stdout -l $lang --oem ${config.engineMode} --psm ${config.pageSegmentationMode}"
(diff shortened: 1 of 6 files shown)


backend/app/utils/Ocr.scala
```diff
@@ -102,15 +106,36 @@ object Ocr extends Logging {
...
+
+    // Timeout is added because ocrMyPdf may get stuck processing a file
+    // and we don't want to block the worker forever
+    def runProcessWithTimeout(processBuilder: ProcessBuilder): Int = {
+      val process = processBuilder.run(ProcessLogger(stdout.append(_), stderr.append))
+      val future = Future(blocking(process.exitValue()))
+      // timeout duration gives at most 1 minute to ocr every page
+      // default 300 minutes is used in case the number of pages failed to
+      // get retrieved from the file and the value was None at this stage
+      val timeoutDuration = numberOfPages.getOrElse(defaultTimeoutInMinutes)
```
itsibitzi, line 123: Pedantic but if you're working with untyped durations it can be helpful to have the unit in the name, e.g. `timeoutMinutes`

You could also define a constant `TIMEOUT_DURATION_PER_PAGE = 1`, which might make the code scan more easily.

I appreciate the comment addresses a few of these concerns, but might be worth putting into the code too.

itsibitzi (approved):
lgtm! 

Will this result in an extraction failure being assigned?

backend/app/utils/Ocr.scala
marjisound: good point. I made the changes you suggested. I also added a 5 minute to the overall timeout value, to reduce the risk of un-necessary timeout for files with low number of pages, in case there is some warming up time when ocrMyPdf starts processing a file. Let me know if you think this is not needed

zekehuntergreen (approved):
Looks good! For posterity, we looked through the `ingestion_events` table to confirm that (1 minute * # pages) + 5 minutes should be more than enough time to finish ocrmypdf.

backend/app/utils/Ocr.scala
```diff
@@ -102,15 +109,36 @@ object Ocr extends Logging {
...
+    // Timeout is added because ocrMyPdf may get stuck processing a file
+    // and we don't want to block the worker forever
+    def runProcessWithTimeout(processBuilder: ProcessBuilder): Int = {
+      val process = processBuilder.run(ProcessLogger(stdout.append(_), stderr.append))
+      val future = Future(blocking(process.exitValue()))
+      // timeout duration gives at most 1 minute to ocr every page plus an overall 5 minutes
+      // to mitigate the risk of un-necessary timeout for files with low number of pages.
+      // default 300 minutes is used in case the number of pages failed to
+      // get retrieved from the file and the value was None at this stage
+      val timeoutMinutes = numberOfPages.map(p => (p * TIMEOUT_DURATION_PER_PAGE) + 5).getOrElse(DEFAULT_TIMEOUT_IN_MINUTES)
```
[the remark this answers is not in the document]
marjisound: Good idea, added a warn log for when numberOfPages is None

PR #162, opened 2023-10-19
Fix `NullPointerException` in `EmlEmailExtractor`

zekehuntergreen:
## What does this change?
`MimeBodyPart.getEncoding` can return null if the header is unavailable, leading to a null pointer exception when we call toLowerCase on the return value. Wrap the return value in an Option to handle this case.


marjisound (approved):
LGTM

PR #163, opened 2023-10-19
Fix incorrect status in observability dashboard for in-progress extractions 

zekehuntergreen:
## Before
At the point in the ingestion / extraction process where extractors have been detected but not yet run are mislabelled as 'Complete' in 'status' column. The cell in the 'extractors' column is empty.
Screenshot 2023-10-19 at 12 18 14

## After
The tables accurately reflect the status of these extractors - the status is 'in progress', extractors that need to be run are listed.
Screenshot 2023-10-19 at 12 18 44


marjisound (approved):
Nice one 👍

PR #164, opened 2023-10-23
Enable caching for github actions workflow

philmcmahon:
## What does this change?
Following advice from @adamnfish and in an attempt to speed up giant builds (and generally be less wasteful), let's enable caching for the relevant [setup npm](https://github.com/actions/setup-node/tree/main) and [setup java](https://github.com/actions/setup-java#caching-sbt-dependencies) steps. 

We can also look at caching the docker containers we use to run the tests but I'm guessing that might be slightly more complex, so possibly for a future PR.

## How to test
If this build completes https://github.com/guardian/giant/actions/runs/6614917759/job/17965921924 that will be encouraging. We'd then need to run a few more builds to get some updated build time data. 

Faster builds, less energy wastage, happy days

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,6 +33,7 @@ jobs:
         with:
           node-version-file: .nvmrc
           cache: npm
+          cache-dependency-path: frontend/package-lock.json
 
       # TODO: configure caching
       # See https://github.com/actions/setup-java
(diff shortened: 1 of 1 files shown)


philmcmahon:
Update: we could consider using https://github.com/marketplace/actions/docker-cache (third party action) to cache the elasticsearch and neo4j docker containers that are fetched in order to run the integration tests.

philmcmahon:
Update - it looks like the caching is working based off the output from recent builds. Still not sure what the build time impact will be but 🤞🏻

marjisound (approved):
LGTM

PR #165, opened 2023-10-25
Add ffmpeg pre-process in transcription extractor

marjisound:
## What does this change?
This PR is on the pm-transcription branch and adds the `ffmpeg` conversion step within the transcription extractor. This step is mandatory before process the file through whisper for transcription. Therefore, if this step errors, the extraction fails and the ffmpeg error message is recorded in the Extraction_failure neo4j relationship and also in postgres database. 

The full range of audio and video formats are too many, so in this PR, only the formats that have already been uploaded in Giant or the most common formats have been added. Most of these formats were tested but for some of them, I couldn't find any samples. Some of the tested ones, also had problem in either whisper step, or some were successfully ingested, but preview didn't work for them. 

The mentioned issues could be resolved in future PRs.


backend/app/extraction/TranscriptionExtractor.scala
```diff
@@ -43,7 +59,8 @@ class TranscriptionExtractor(index: Index, scratchSpace: ScratchSpace, ingestion
     println(params.languages)
 
     val result = Either.catchNonFatal{
-      val transcriptResult = Whisper.invokeWhisper(file.toPath, tmpDir, translate =false)
+      val convertedFile = FfMpeg.convertToWav(file.toPath, ffMpegTmpDir)
```
zekehuntergreen, line 62: It sounds like conversion is faster than transcription but would it be worth skipping this if the file's already a wav?

backend/app/extraction/TranscriptionExtractor.scala
```diff
@@ -1,26 +1,41 @@
...
+import java.io.File
 import scala.io.Source
-import scala.reflect.runtime.universe.Try
-import scala.util.control.NonFatal
 
 class TranscriptionExtractor(index: Index, scratchSpace: ScratchSpace, ingestionServices: IngestionServices) extends FileExtractor(scratchSpace) with Logging {
   val mimeTypes: Set[String] = Set(
     "audio/wav",
     "audio/vnd.wave",
+    "audio/x-aiff", // converted and transcribed. But preview doesn't work
```
zekehuntergreen, line 21: Thanks for making a note of the results of all of these different file types!

backend/app/utils/FfMpeg.scala
```diff
@@ -0,0 +1,24 @@
...
+import scala.sys.process.{Process, ProcessLogger, stdout}
+
+object FfMpeg extends Logging {
+
+  case class FfMpegSubprocessCrashedException(exitCode: Int, stderr: String) extends Exception(s"Exit code: $exitCode: ${stderr}")
+
+  def convertToWav(originalFile: Path, tmpDir: Path): Path = {
+    val ffMpegStdErrLogger = new BasicStdErrLogger()
+    val tempFile = tmpDir.resolve(s"${originalFile.getFileName}.wav")
+    val cmd = s"ffmpeg -i $originalFile -ar 16000 -ac 1 -c:a pcm_s16le ${tempFile}"
```
zekehuntergreen, line 13: Do we need to think about how to install `ffmpeg` in dev and prod environments? Or did we say it comes with whisper 🤔

philmcmahon: I think it would be worth adding here https://github.com/guardian/giant/blob/main/Brewfile, I have already added it here https://amigo.gutools.co.uk/recipes/investigations-giant-app-java11

backend/app/utils/Whisper.scala
```diff
@@ -1,31 +1,16 @@
...
-    acc.mkString("\n")
-  }
-}
 object Whisper extends Logging {
   private class WhisperSubprocessCrashedException(exitCode: Int, stderr: String) extends Exception(s"Exit code: $exitCode: ${stderr}")
 
 
   def invokeWhisper(audioFilePath: Path, tmpDir: Path, translate: Boolean): TranscriptionResult = {
-    val whisperLogger = new WhisperStdErrLogger()
+    val whisperLogger = new BasicStdErrLogger()
```
philmcmahon, line 13: 👍🏻 thanks for making this generic!

backend/app/utils/FfMpeg.scala
```diff
@@ -0,0 +1,24 @@
+package utils
+
+import java.nio.file.Path
+import scala.sys.process.{Process, ProcessLogger, stdout}
+
+object FfMpeg extends Logging {
+
+  case class FfMpegSubprocessCrashedException(exitCode: Int, stderr: String) extends Exception(s"Exit code: $exitCode: ${stderr}")
+
+  def convertToWav(originalFile: Path, tmpDir: Path): Path = {
```
philmcmahon, line 10: maybe worth adding a comment explaining that whisper.cpp only works on 16 bit wavs and linking to the docs https://github.com/ggerganov/whisper.cpp#quick-start

philmcmahon (approved):
Great work 👍🏻  

convert

backend/app/utils/FfMpeg.scala
marjisound: Thanks for the comments, updated the brew file 👍

backend/app/utils/FfMpeg.scala
marjisound: Good suggestion, added the comment 👍

PR #168, opened 2023-10-30
Add transcript.<language> query param on transcribed resource url

marjisound:
## What does this change?
This PR adds a check if resource has transcript, and returns the transcript.<language> for view query parameter. 

It also removes the `Text` tab on the viewer if resource has transcript.

### DIFF
--- a/frontend/src/js/util/resourceUtils.ts
+++ b/frontend/src/js/util/resourceUtils.ts
@@ -72,6 +72,10 @@ export function getDefaultView(resource: Resource): string | undefined {
         return undefined;
     }
 
+    if (resource.transcript) {
+        return "transcript." + Object.keys(resource.transcript)[0];
+    }
+
     // We removed the isBasic check in Viewer during conversion to Typescript because it meant changing the definitions
     // of Resource and BasicResource in ways that would ripple across the codebase.
     // This is effectively the same thing, but without TypeScript really knowing what's going on.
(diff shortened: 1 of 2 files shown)


philmcmahon (approved):
Nice one

frontend/src/js/util/resourceUtils.ts
```diff
@@ -72,6 +72,10 @@ export function getDefaultView(resource: Resource): string | undefined {
         return undefined;
     }
 
+    if (resource.transcript) {
+        return "transcript." + Object.keys(resource.transcript)[0];
```
marsavar, line 76: Two comments:

1) I think the custom for string interpolation (following what's already in this file - and modern practice in general since ES6) would be `transcript.${Object.keys(...)}`` rather than the `+` operator.
2) Question (this is not a blocker, it's just for my understanding): is there ever a chance where a transcript could be more than one language? Just asking because since you're always indexing into the first element of the array - and also looking at the current modeling of `transcript`, which is 
```typescript
    transcript?: {
        [lang: string]: HighlightableText
    }
```
it doesn't look like that's the case. So I wonder why it wasn't modeled like so:
```
    transcript?: {
        lang: string,
        text: HighlightableText
    }
```
That would allow you to do something like
```typescript
return `transcript.${resource.transcript.lang}`
```
which feels cleaner.
Again, not a blocker - just curious :)

marjisound: Thanks @marsavar for looking. I made a change for your 1st comment :)

Regarding the 2nd comment, yes there are cases where we have transcription on several languages. The reason I'm choosing the first index, is that this would return the view query parameter (e.g. view=transcript.arabic) which is used when the viewer page opens. The user would still be able to change the tabs to go to other transcriptions if they exist but the first one is in the view at first.

PR #169, opened 2023-10-30
Moving react-scripts to devDependencies as it's only used in build

srbd:
## What does this change?
`react-scripts` has a dependency of `babel/traverse@7.21.5` which has a critical vulnerability.

However as it's used as a build tool as part of CRA we shall move it to `devDependencies`. 

See https://github.com/facebook/create-react-app/issues/11174

## How to test

tested on CODE, and this should remove the critical vulnerability warning from Snyk once merged as Synk doesn't look at `devDependencies`.


philmcmahon (approved):
vaguely relevant script gif

PR #171, opened 2023-11-02
Make giant logging less noisy

philmcmahon:
## What does this change?
This removes request logging for succesful requests to '/healthcheck' from giant logs, which makes up a lot of the noise. 

It also removes 'authentication succeeded' logging, I think it's ok to only log in the 'failure' path. 

## How to test
Tested locally.

##success?
More readable logs, less filtering required


marjisound (approved):
This is great. What a pain to filter out these logs every-time we were looking at the logs

PR #172, opened 2023-11-04
feat: full docker containerization

komachi:
## What does this change?

I wanted to compare giant with aleph and datashare and found it's pretty hard to run giant, so I made a Dockerfile for it to run everything with single command.

## How to test

Run `docker compose up` in root, then visit http://localhost:9001

I also fixed package-lock in frontend (it missed hash for one dep so it was not installable with npm ci, also I run npm audit fix and updated caniuse-lite db).

For postgres/migrate-db, I added postgrator-cli to run migration easily and without custom script.

Also I made docker image with java17 because it's default java in debian bookworm (current stable), this forced me to remove some deprecated cli opts not supported anymore in java17. If you still need it to run on java11 with this options, should they be moved somewhere else?

Some config options also became configurable via ENV and default endpoints changed from localhost to docker-compose hosts.

Deps in docker-compose were updated to latest non-breaking change.


philmcmahon:
Thanks for this @komachi! We'll go through this PR this week. At a quick glance looks like a whole bunch of very helpful changes, we'll just need to review to check for compatibility with our production cluster.

PR #174, opened 2023-11-06
allow admin user to delete a public workspace

marjisound:
## What does this change?
This PR updates the query for deleting workspace in order to allow the admins to delete a public workspace. 
Also on the UI, the delete button is only enabled if the user is admin and the workspace is public, or if the user is the owner of the workspace. 

## How to test
This was tested locally and in playground. In playground, I asked my colleague to create a workspace and make it public. I then deleted that workspace as I am an admin user.


philmcmahon (approved):
Great work on this, the queries look good to me 
delete

itsibitzi:
The vertical line on the right is a few degrees off axis at the end of the animation.

marjisound:
> The vertical line on the right is a few degrees off axis at the end of the animation.

Nice catch, it is indeed :)

PR #176, opened 2023-11-09
Set whisper model via configuration parameter

philmcmahon:
## What does this change?
This pulls out the model we are using for whisper into a configuration parameter, which will allow us to change the model we are using by updating a value in parameter store and then redeploying giant. It also allows us to use different models on playground/dev/prod - which may be useful given that playground runs on less performant hardware to PROD, and often when testing stuff we care less about the quality of the transcription.

In future, we might want to make this even more configurable by using a feature switch

## How to test
Tested on playground - I switched to using the `small` model and had a succesful trasnsription. 

To test in production edit the transcribe/modelFilename param in SSM

Based on some recent experiments, we might be able to get away with bumping the model to `small` from `base` - this would make that an easier process.

### DIFF
--- a/backend/app/AppComponents.scala
+++ b/backend/app/AppComponents.scala
@@ -161,7 +161,7 @@ class AppComponents(context: Context, config: Config)
     val imageOcrExtractor = new ImageOcrExtractor(config.ocr, scratchSpace, esResources, ingestionServices)
     val ocrMyPdfImageExtractor = new OcrMyPdfImageExtractor(config.ocr, scratchSpace, esResources, previewStorage, ingestionServices)
 
-    val transcriptionExtractor = new TranscriptionExtractor(esResources, scratchSpace)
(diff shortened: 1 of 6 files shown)


komachi:
Check out [distil-whisper](https://github.com/huggingface/distil-whisper) and [this issue](https://github.com/ggerganov/whisper.cpp/issues/1423) regarding distill-whisper support in whisper.cpp. This should run much faster without much error penalty.

philmcmahon:
> Check out [distil-whisper](https://github.com/huggingface/distil-whisper) and [this issue](https://github.com/ggerganov/whisper.cpp/issues/1423) regarding distill-whisper support in whisper.cpp. This should run much faster without much error penalty.

Fantastic, we'd seen distil-whisper but I hadn't realised it was already in whisper.cpp! It's english-only at the moment, whichis probably 90% of what we're transcribing but we'll need some kind of language detection before deciding whether we can use distill-whisper (or wait for them to do other languages)

marjisound (approved):
This is great 👏

PR #177, opened 2023-11-09
Fix transcription bug where a translation overwrites a transcription

philmcmahon:
## What does this change?
We had a bug in the transcript code - addDocumentTranscript is an async function but we were calling it once for the transcription then again for the translation without waiting for the first call to finish, resulting in unpredictable behaviour (often the transcription would overwrite the translation or vice versa).

This fixes that by writing both transcript and translated transcript in a single elasticsearch query

## How to test
I've tested this locally with a file in german. Will test on playground just as soon as github actions finishes...


marjisound (approved):
Looks great 🥳

PR #178, opened 2023-11-21
Create API token page for whistleflow

zekehuntergreen:
## What does this change?
Creates a page with a logged-in user's jwt token on it so that it can be scraped by [whistleflow-giant-login](https://github.com/guardian/whistleflow/pull/43).

## Why
To enable export from SecureDrop Workstation to Giant, we need SDW users to be able to log in to Giant in a dedicated whistleflow-giant-login VM, and scrape their jwt token so that it can be used to authenticate against the Giant API to retrieve a list of their workspaces and to upload files.

This token is currently available on the [settings/about](https://giant.pfi.gutools.co.uk/settings/about) page but creating a dedicated page will reduce the risk that future changes to the about page break whistleflow.

### DIFF
--- a/frontend/src/js/App.js
+++ b/frontend/src/js/App.js
@@ -32,6 +32,7 @@ import SettingsSidebar from './components/Settings/SettingsSidebar';
 import ExtractionFailures from './components/Settings/ExtractionFailuresComponent';
 import Users from './components/Settings/Users';
 import About from './components/Settings/About';
+import Token from './components/UtilComponents/Token';
 import FeatureSwitches from './components/Settings/FeatureSwitches';
 import { WeeklyUploadsFeed } from './components/Uploads/Uploads';
 
@@ -93,6 +94,10 @@ class App extends React.Component {
 
                     <Route path = '/settings/my-uploads' component={MyUploads} />
                     <Route path = '/settings/all-ingestion-events' component={AllIngestionEvents} />
+
+                    {/* This page is used by securedrop workstation / whistleflow to retrieve a jwt token
+                        that will be used to export SDW submissions to giant */}
+                    <Route path='/token' component={Token} />
                 </div>
             </div>
             <Route exact path='/' render={() => <Redirect to='/search' />} />
@@ -113,7 +118,7 @@ class App extends React.Component {
         </div>;
     }
 
-    renderAnonymouse() {
+    renderAnonymous() {
         return <Switch>
             <Route path='/login' component={Login} />
             <Route path='/register' component={Register} />
@@ -139,7 +144,7 @@ class App extends React.Component {
             <ConnectedRouter history={this.props.history}>
                 <div className='app'>
                     <Header user={maybeUser} config={this.props.config} preferences={this.props.preferences}/>
-                    {loggedIn ? this.renderLoggedIn() : this.renderAnonymouse()}
+                    {loggedIn ? this.renderLoggedIn() : this.renderAnonymous()}
                 </div>
             </ConnectedRouter>
         );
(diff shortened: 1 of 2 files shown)


frontend/src/js/App.js
```diff
@@ -113,7 +118,7 @@ class App extends React.Component {
         </div>;
     }
 
-    renderAnonymouse() {
```
zekehuntergreen, line 116: 🐁

itsibitzi: Noooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo

frontend/src/js/components/UtilComponents/Token.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+type Props = ReturnType<typeof mapStateToProps>
+
+class Token extends React.Component<Props> {
+    render() {
+        return (
+            <div className="app__main-content">
+                {/* Whistleflow relies on this id */}
+                <div id="giant-api-token">{this.props.auth.jwtToken}</div>
+            </div>
```
philmcmahon, line 14: I was wondering whether it might be worth adding a bit of explainer text in case something goes wrong and the user gets stuck on this page - something like "If you see this page for more than a few seconds, something has gone wrong with securedrop login. Please close this window and try again."

philmcmahon (approved):
Nice 👍🏻 

I was originally thinking this could just be a JSON endpoint, but I think we need all the React to do the login dance and pass the cookie.

give me tokens

frontend/src/js/App.js
philmcmahon: 🐀

frontend/src/js/App.js
```diff
@@ -118,7 +118,7 @@ class App extends React.Component {
         </div>;
     }
 
-    renderAnonymous() {
+    renderAnonymouse() {
```
zekehuntergreen, line 121: anonymouse lives to squeak another day

frontend/src/js/components/UtilComponents/Token.tsx
zekehuntergreen: Good idea

zekehuntergreen:
@itsibitzi pointed out that authed giant users will have a `pfiAuthHeader` in local storage which puppeteer could just access directly rather than needing to scrape a token from a dedicated token page.
But we need some way to know whether or not the user has gone through the auth flow and landed on Giant. It seems like the accepted way to do this is using `page.waitForSelector` passing a selector for an element that exists on the Giant page we're redirected to after successful auth (and which hopefully doesn't exist in current or future google login pages). We _could_ close this PR and instead use a Giant url + selector combination that already exists but we then risk breaking whistleflow auth if that selector ever changes, which might not be unlikely especially if we make loads of front end changes as part of the switch to elastic ui.
So even though local storage is an option, I think it makes sense to create this dedicated whistleflow login landing page.

philmcmahon:
> @itsibitzi pointed out that authed giant users will have a `pfiAuthHeader` in local storage which puppeteer could just access directly rather than needing to scrape a token from a dedicated token page. But we need some way to know whether or not the user has gone through the auth flow and landed on Giant. It seems like the accepted way to do this is using `page.waitForSelector` passing a selector for an element that exists on the Giant page we're redirected to after successful auth (and which hopefully doesn't exist in current or future google login pages). We _could_ close this PR and instead use a Giant url + selector combination that already exists but we then risk breaking whistleflow auth if that selector ever changes, which might not be unlikely especially if we make loads of front end changes as part of the switch to elastic ui. So even though local storage is an option, I think it makes sense to create this dedicated whistleflow login landing page.

That seems reasonable to me, at least this way it is very explicit what is happening and what this page is for

PR #180, opened 2023-11-27
Upgrade to version 3 of the Play Framework. Remove akka completely

philmcmahon:
## What does this change?
Play 3.0 is the latest version of the play framework. There are two main benefits of upgrading:
 - Play 3 [uses pekko instead of akka](https://www.playframework.com/documentation/3.0.x/General#How-Play-Deals-with-Akkas-License-Change), so we can remove the last traces of akka from our codebase and not risk angering the license gods
 - Play 3 supports Scala 3

This was a pretty easy upgrade - just a few dependency bumps. The only 'code' change was I had to increase the length of the [play application secret ](https://www.playframework.com/documentation/3.0.x/ApplicationSecret)

We can't actually move to Scala 3 until we upgrade to elasticsearch 8 as elastic4s doesn't support elasticsearch 7 at the moment. 

## How to test
Tested on playground.

### DIFF
--- a/backend/test/utils/auth/AuthActionBuilderTest.scala
+++ b/backend/test/utils/auth/AuthActionBuilderTest.scala
@@ -30,7 +30,7 @@ class AuthActionBuilderTest extends AnyFreeSpec with Matchers with BaseOneAppPer
   override def fakeApplication(): Application = {
     val env = Environment.simple(new File("."))
     val initialSettings: Map[String, AnyRef] = Map(
-      "play.http.secret.key" -> "TestKey",
+      "play.http.secret.key" -> "MySecretKeyShhDontTellAnyoneOrIWillBeVerySad",
       "play.http.session.maxAge" -> Int.box(900000),
       "pekko.actor.provider" -> "local"
     )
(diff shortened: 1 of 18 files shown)


backend/test/utils/auth/AuthActionBuilderTest.scala
```diff
@@ -30,7 +30,7 @@ class AuthActionBuilderTest extends AnyFreeSpec with Matchers with BaseOneAppPer
   override def fakeApplication(): Application = {
     val env = Environment.simple(new File("."))
     val initialSettings: Map[String, AnyRef] = Map(
-      "play.http.secret.key" -> "TestKey",
+      "play.http.secret.key" -> "MySecretKeyShhDontTellAnyoneOrIWillBeVerySad",
```
zekehuntergreen, line 33: 🤫

build.sbt
```diff
@@ -81,10 +80,11 @@ lazy val common = (project in file("common"))
   .settings(
     name := "common",
     scalacOptions := compilerFlags,
+//    dependencyOverrides += "org.scala-lang.modules" %% "scala-xml" % "2.2.0",
```
zekehuntergreen, line 83: should we remove?

zekehuntergreen (approved):
Thanks for upgrading 🎉

build.sbt
philmcmahon: yes!

PR #181, opened 2023-12-07
Upgrade postgres and scalike libraries

philmcmahon:
## What does this change?
Resolves https://security.snyk.io/vuln/SNYK-JAVA-ORGPOSTGRESQL-2401816

The scalike upgrade wasn't really necessary as part of this but I went for it in the hope it would be painless as it is a requirement for Scala 3. In the end there was only a small code change required - doing '.apply()' doesn't appear necessary anymore.

## How to test
Tested on playground - giant still appears to be happily chatting with postgres.


zekehuntergreen (approved):
I can't seem to find any release notes or documentation on upgrading to scalikejdbc 4 🤔  
Happy with this based on your test in CODE though

PR #187, opened 2023-12-18
Upgrade elasticsearch SDK to version 8.11

philmcmahon:
## What does this change?
(Draft PR pending upgrade to ES8)

Following from https://github.com/guardian/giant/pull/190 This PR drags giant's elasticsearch code into the future, ready to support elasticsearch 8.

Key changes:
 - version number bumps in build.sbt, docker-compose, elasticsearch docker service
 - Replacing [deprecated updateByQuery](https://github.com/sksamuel/elastic4s/blob/83162ceb97be3e023f7644fce637e2da0678896c/elastic4s-core/src/main/scala/com/sksamuel/elastic4s/api/UpdateApi.scala#L11)

## How to test
Tested locally and on playground. Make sure you have run https://github.com/guardian/giant/pull/190 (now merged into main) locally before trying this as you need your local DB to already be on 7.17 to work

Deploying on PROD:
 - ideally it would be good to use rest compatability mode https://www.elastic.co/guide/en/elasticsearch/reference/8.11/rest-api-compatibility.html so that we could upgrade the DB without breaking giant, and then apply this PR afterwards. Unfortunately I can't find any support for compatability mode anywhere in the elastic4s documentation - opened an issue https://github.com/sksamuel/elastic4s/issues/2978 in case anyone has a suggestion. 

HOWEVER, we appear to be succesfully running sdk 7.17 against elasticsearch 8 on playground, so maybe we don't need compatability mode

### DIFF
--- a/backend/app/services/index/ElasticsearchPages.scala
+++ b/backend/app/services/index/ElasticsearchPages.scala
@@ -9,6 +9,7 @@ import model.index.{Page, PageResult, PagesSummary}
 import model.{Language, Languages, Uri}
 import services.ElasticsearchSyntax
 import services.index.HitReaders.{PageHitReader, RichFieldMap}
+import services.table.TableRowFields
 import utils.Logging
 import utils.attempt.{Attempt, ElasticSearchQueryFailure, MultipleFailures, NotFoundFailure}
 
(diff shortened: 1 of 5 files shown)


marjisound (approved):
Looks good

PR #189, opened 2024-01-02
fix transcript and extractText download

marjisound:
## What does this change?
This PR does the following:
- Fixes the extractedText download. the actual text is within the contents field within the text object. The existing code was trying to get the text string from the text itself. 
- Fixes the transcript download. The transcript download was added as a new download type so that depending on which transcript language the user picks, it would download the relevant language text file. 

ezgif-1-522aa3e19e


hoyla:
Awesome!

philmcmahon (approved):
Nice

PR #191, opened 2024-01-04
Support copying files between workspaces

philmcmahon:
## What does this change?
Giant currently supports moving files between workspaces via drag/drop. There has been a user request to support copying files rather than moving them. This PR adds that functionality, by popping open a modal to ask the user whether they want to move or copy when a drag/drop operation takes place between workspaces.

Copying is a bit harder than moving as for moving it's just an update operation to change the parent ID of the item in question. 

The main complexity in the new feature is the new 'copyItem' endpoint in the backend - which relies on a recursive `copyTree` function in `Workspaces.scala` to perform the actual copy.

## How to test
Tested locally and on playground - looks like this:

https://github.com/guardian/giant/assets/3606555/28be0178-d589-4db4-8c29-e13981781452

## Security
An obvious concern here is that we don't want to allow users to copy files from a workspace they don't have access to into one that they do have access to.

Ensuring user has access to source files: This is done by the call to `getWorkspaceContents`
Ensuring user has access to destination: This is done by the calls to `addFolder` and `addResourceToWorkspaceFolder` in `insertItem`

### DIFF
--- a/backend/app/controllers/api/Workspaces.scala
+++ b/backend/app/controllers/api/Workspaces.scala
@@ -2,8 +2,8 @@ package controllers.api
 
 import commands.DeleteResource
 import model.Uri
-import model.annotations.Workspace
-import model.frontend.TreeEntry
+import model.annotations.{Workspace, WorkspaceEntry, WorkspaceLeaf}
+import model.frontend.{TreeEntry, TreeLeaf, TreeNode}
 import net.logstash.logback.marker.LogstashMarker
 import net.logstash.logback.marker.Markers.append
 import play.api.libs.json._
@@ -201,6 +201,27 @@ class Workspaces(override val controllerComponents: AuthControllerComponents, an
     }
   }
 
+  private def copyTree(workspaceId: String, destinationParentId: String, tree: TreeEntry[WorkspaceEntry], user: String): Attempt[List[String]] = {
+    println("copy tree")
+    val newId = UUID.randomUUID().toString
+    tree match {
+      case TreeLeaf(_, name, data, _) =>
+        data match {
+          case WorkspaceLeaf(_, _, _, _, uri, mimeType, size) =>
+            val addItemData = AddItemData(name, destinationParentId, "file", Some("document"), AddItemParameters(Some(uri), size, Some(mimeType)))
+            insertItem(user, workspaceId, newId, addItemData).map(i => List(i))
+          case _ => Attempt.Left(WorkspaceCopyFailure("Unexpected data type of TreeLeaf"))
+        }
+
+      case TreeNode(_, name, data, children) =>
+        val addItemData = AddItemData(name, destinationParentId, "folder", None, AddItemParameters(None, None, None))
+        val newChildIds = insertItem(user, workspaceId, newId, addItemData).flatMap{_ =>
+          Attempt.traverse(children)(child =>  copyTree(workspaceId, newId, child, user))
+        }
+        newChildIds.map(ids => newId +: ids.flatten )
+    }
+  }
+
   def addItemToWorkspace(workspaceId: String) = ApiAction.attempt(parse.json) { req =>
     for {
       data <- req.body.validate[AddItemData].toAttempt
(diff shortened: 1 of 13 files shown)


frontend/src/js/actions/workspaces/copyItem.ts
```diff
@@ -0,0 +1,46 @@
...
+import {ThunkAction} from "redux-thunk";
+import {GiantState} from "../../types/redux/GiantState";
+import {AppAction, AppActionType, WorkspacesAction} from "../../types/redux/GiantActions";
+import {getWorkspace} from "./getWorkspace";
+
+export function copyItems(
+    workspaceId: string,
+    itemIds: string[],
+    newWorkspaceId?: string,
+    newParentId?: string
```
zekehuntergreen, line 11: Is there a call site where `newParentId` is passed? If not we could simplify by removing the parameter here and on the backend.

zekehuntergreen: maybe it's not necessary because we only copy to the root of the new workspace, whereas we can move to a different folder in the same workspace

frontend/src/js/components/workspace/WorkspacesSidebarItem.tsx
```diff
@@ -17,19 +20,33 @@ type PropTypes = ReturnType<typeof mapDispatchToProps>
...
-            const entryIds = getIdsOfEntriesToMove(selectedEntries, idOfDraggedEntry);
-            moveItems(selectedWorkspaceId, entryIds, linkedToWorkspaceId);
-        }}
-        key={linkedToWorkspaceId}
-        to={`/workspaces/${linkedToWorkspaceId}`}
-    >
-        <div className='sidebar__item__text'>{linkedToWorkspaceName}</div>
-    </SidebarSearchLink>
+
+const WorkspacesSidebarItem: FC<PropTypes> = ({selectedEntries, moveItems, copyItems, selectedWorkspaceId, linkedToWorkspaceId, linkedToWorkspaceName}) => {
```
zekehuntergreen, line 24: There's an edge case that already exists where `selectedWorkspaceId` == `linkedToWorkspaceId` which probably means that the user messed up and didn't actually want to copy or move items to the workspace where they're already located. Could be worth either doing nothing or showing an error modal in this case?

backend/app/controllers/api/Workspaces.scala
```diff
@@ -256,6 +283,22 @@ class Workspaces(override val controllerComponents: AuthControllerComponents, an
     }
   }
 
+  def copyItem(workspaceId: String, itemId: String) = ApiAction.attempt(parse.json) { req =>
+    for {
+      data <- req.body.validate[MoveItemData].toAttempt
```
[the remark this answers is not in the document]
philmcmahon: good shout, maybe `MoveCopyDestination`?

frontend/src/js/actions/workspaces/copyItem.ts
philmcmahon: Ah yeah good spot! I was originally designing this with the intention that you would be able to copy to a folder within the same workspace, but when 'hold down ctrl' didn't work I gave up. I think it might be worth keeping this in though so if there is a use case in future the wiring is already there maybe?

frontend/src/js/actions/workspaces/copyItem.ts
```diff
@@ -0,0 +1,46 @@
...
+): ThunkAction<void, GiantState, null, WorkspacesAction | AppAction> {
+    console.log("copying")
+    return dispatch => {
+        return copyItemApi(workspaceId, itemId, newWorkspaceId, newParentId)
+            .then(() => {
+                dispatch(getWorkspace(workspaceId));
+            })
+            .catch(error => dispatch(() => errorCopyingItem(error)));
+    };
+}
```
marjisound, line 37: I may be missing something here, but the consumer of this copyItems function is not providing newParentId argument. Could this argument be needed for some future work?

frontend/src/js/actions/workspaces/copyItem.ts
```diff
@@ -0,0 +1,46 @@
...
+export function copyItems(
+    workspaceId: string,
+    itemIds: string[],
+    newWorkspaceId?: string,
+    newParentId?: string
+): ThunkAction<void, GiantState, null, WorkspacesAction | AppAction> {
+    return dispatch => {
+        for (const itemId of itemIds) {
+            if (itemId !== newParentId) {
+                console.log("dispatching")
```
marjisound, line 16: Do we need this log?

frontend/src/js/actions/workspaces/copyItem.ts
philmcmahon: You're right - it's not currently used but it could be used in future to support 'copy to folder within the same workspace' functionality. I have added a comment explaining this

frontend/src/js/actions/workspaces/copyItem.ts
philmcmahon: I've added a comment to explain this

philmcmahon:
Screenshot 2024-01-09 at 17 54 44

frontend/src/js/components/workspace/WorkspacesSidebarItem.tsx
philmcmahon: Great shout - here's the new modal 
Screenshot 2024-01-09 at 17 54 44

PR #193, opened 2024-01-15
Remove deprecated bootstrap.system_call_filter setting

philmcmahon:
## What does this change?
Following the upgrade to ES8 #187  this setting is no longer valid


marjisound (approved):
Great thanks for doing it

PR #194, opened 2024-01-16
reprocess instead of upload if uploaded file already exist

marjisound:
## What does this change?
This PR adds a new api endpoint to upload files so that it would not create new nodes if the file already exists.
- expects a new header `X-PFI-Ingestion-Name`
- using the file fingerprint, checks if same file already exists in the same workspace path
- reprocess the file if already exists
- if file doesn't already exist, it checks if there's already an ingestion with same name as retrieved from header `X-PFI-Ingestion-Name`
- If ingestion doesn't already exist, it creates ingestion, otherwise it uses the existing ingestion
- It processes the ingest file 
image

As can be seen in the video when the same file is uploaded in a workspace path that already has that file, file is reprocessed. But if that file is uploaded into a different path of the same workspace, the file upload does occur. 

https://github.com/guardian/giant/assets/15894063/5104db8e-a425-4c03-a2ac-2d29f6c74d91

## How to test
The old upload endpoint is untouched and can be used as it could previously. This was tested by undoing the UI changes and do the file uploads with the old UI which is still making request to the old endpoint and it worked fine. 

The new endpoint was also tested in these scenario:
- upload same file in the same workspace path
- upload same file in a different workspace path e.g. different folder
- upload multiple files when one or some of them already exists in the same workspace path but some of them are new
- upload a folder with files that already exist in the workspace (in this scenario file within the folder are all uploaded)
- upload a folder with duplicate files (in this scenario duplicate files are not uploaded)

### DIFF
--- a/backend/app/commands/CreateIngestion.scala
+++ b/backend/app/commands/CreateIngestion.scala
@@ -23,11 +23,9 @@ case class CreateIngestion(data: CreateIngestionRequest, collectionUri: Uri, man
     val current = ingestionUri.flatMap(uri => manifest.getIngestion(uri))
 
     current.map { c =>
-      println(s"ingestion ${c.uri} already exists")
       Uri(c.uri)
     }.recoverWith {
       case f: NotFoundFailure =>
-        println(s"error getting ingestion - ${f.msg}")
         createIngestion(ingestionId, ingestionUri)
     }
   }
--- a/backend/app/commands/CreateIngestion.scala
+++ b/backend/app/commands/CreateIngestion.scala
@@ -2,7 +2,7 @@ package commands
 
 import java.nio.file.Paths
 
-import model.{CreateIngestionRequest, Language, Languages, Uri}
+import model.{CreateIngestionRequest, Languages, Uri}
 import services.index.{Index, Pages}
 import services.manifest.Manifest
 import utils.{Logging, UriCleaner}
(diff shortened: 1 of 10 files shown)


backend/app/commands/CreateIngestion.scala
```diff
@@ -13,22 +13,47 @@ import scala.concurrent.ExecutionContext
...
+    current.map { c =>
+      Uri(c.uri)
+    }.recoverWith {
+      case f: NotFoundFailure =>
+        createIngestion(ingestionId, ingestionUri)
     }
+  }
 
-    val langs = data.languages.flatMap(Languages.getByKey)
+  private def provideIngestionUri(): (Attempt[String], Attempt[Uri]) = {
```
philmcmahon, line 33: I think in general it might be a good idea to remove the Attempt from stuff before you pass it into a function so that functions don't immediately have to deal with the fact something might be a failure (and to avoid having to construct an Attempt object when e.g. testing the function). I had a go at rewriting this file with that in mind - I think it is a slight improvement - see what you think https://github.com/guardian/giant/commit/aa42028cf1c112631a854cad65c4a1c58a066cf1

backend/app/commands/CreateIngestion.scala
```diff
@@ -13,22 +13,47 @@ import scala.concurrent.ExecutionContext
...
+    val providedDisplayName = getProvidedDisplayName()
 
+    val ingestionId = providedDisplayName.map(d => UriCleaner.clean(d))
+    val ingestionUri = ingestionId.map(id => collectionUri.chain(id))
+    (ingestionId, ingestionUri)
+  }
+
+  private def createIngestion(ingestionId: Attempt[String], ingestionUri: Attempt[Uri]): Attempt[Uri] = {
```
philmcmahon, line 41: I think in general it might be a good idea to remove the Attempt from stuff before you pass it into a function so that functions don't immediately have to deal with the fact something might be a failure (and to avoid having to construct an Attempt object when e.g. testing the function). I had a go at rewriting this file with that in mind - I think it is a slight improvement - see what you think https://github.com/guardian/giant/commit/aa42028cf1c112631a854cad65c4a1c58a066cf1

philmcmahon (approved):
I have just the one scala style suggestion in CreateIngestion.scala which you can see whether you agree with. Otherwise looks great, fantastic work

copy

backend/app/commands/CreateIngestion.scala
marjisound: This is a great suggestion, thanks for pointing it out and also for doing it in a branch. I made the changes :)

frontend/src/js/components/Uploads/UploadTarget.ts
```diff
@@ -21,18 +20,12 @@ export function getDefaultCollection(username: string, collections: Collection[]
     });
 }
 
-async function createUploadIngestion(collection: Collection): Promise<string> {
-    const ingestionName = `${new Date().toISOString()}`;
-    const newIngestion = await createNewIngestion(collection.uri, `Upload ${ingestionName}`);
```
zekehuntergreen, line 26: should we remove `uploadFileToIngestion` and `createNewIngestion` as they're no longer used?

frontend/src/js/components/Uploads/UploadTarget.ts
```diff
@@ -44,11 +37,10 @@ export async function getUploadTarget(
         throw new Error(`No workspace entry in either focused entry ${focusedEntry} or root of workspace ${workspace.id}, cannot upload to workspace`);
     }
 
-    const ingestionUri = await createUploadIngestion(defaultCollection);
-
+    const ingestionName = `Upload ${new Date().toISOString()}`;
```
zekehuntergreen, line 40: non-blocking: while we're creating this new endpoint, would it be worth simplifying things by have the api create the ingestion name based on the current datetime?
Seems like the advantage might be that we simplify client-side code and reduce the risk of misnamed ingestions due to the client's clock being incorrect, and instead use the more reliable server time.

zekehuntergreen (approved):
Looks good! 👏  Just a few non-blocking comments

frontend/src/js/components/Uploads/UploadTarget.ts
marjisound: Yes this is a great idea. To do this there needs to be a bit more changes in the backend, and I think it might be better if that is not included part of this PR to reduce the risk of introducing bug. if that's OK with you, I created this health ticket for it https://trello.com/c/tRaxMQI3/2881-giant-simply-client-by-creating-the-ingestion-name-in-backend

PR #198, opened 2024-02-01
Enable RDS backups with AWS Backup

srbd:
## What does this change?

We'll use [DevX managed RDS backups](https://github.com/guardian/aws-backup) which have tighter access control rather than automated backups (todo: remove the old backup config).

## How to test

Tested on CODE - tags have been added.


philmcmahon (approved):
backup has arrived

PR #199, opened 2024-02-02
Yield control of backups to aws backup

srbd:
## What does this change?

Removing legacy automated backup settings to prevent cloudformation drift, as they are now handled by aws-backups


philmcmahon (approved):
yield

PR #204, opened 2024-03-14
Search in workspace folder

zekehuntergreen:
## What does this change?
- Introduces a new "workspace folder" Chip (documented here [Search in Giant](https://docs.google.com/document/d/1uejYdU7w5GJpmg9R2u-Wi6qr1YfSg7bWHRRazDwHjUU/edit)) which allows a user to search inside a specific folder in a workspace.
  - at the moment, a "workspace folder" chip can be removed but not created or edited
- Adds a "search in folder" item to the context menu when a user right clicks on a folder in the workspace view. This links to the search view with a "workspace folder" chip pre-populated.
- Changes to the /search endpoint to handle this new chip type. Most of the logic to filter by the blobs in a folder was introduced in https://github.com/guardian/pfi/pull/811 and is repurposed here.

## How to test
- [x] tested in CODE

Users will be able to search inside a folder inside a workspace.

## Potential future changes
- We could add the ability to create and edit workspace folder chips directly from the search page with a typeahead of workspace and folder names. 

## Example
https://github.com/guardian/giant/assets/17057932/5ae08934-3080-4b0f-ae9c-8f93f3db68fd

### DIFF
--- a/backend/app/controllers/api/Resource.scala
+++ b/backend/app/controllers/api/Resource.scala
@@ -38,7 +38,7 @@ class Resource(val controllerComponents: AuthControllerComponents, manifest: Man
 
     val resourceFetchMode = (basic, parsedChips) match {
       case (true, _) => ResourceFetchMode.Basic
-      case (false, pt) => ResourceFetchMode.WithData(pt.map(_.query))
+      case (false, pc) => ResourceFetchMode.WithData(pc.map(_.query))
     }
 
     val decodedUri = Uri(UriEncoding.decodePath(uri.value, StandardCharsets.UTF_8))
--- a/backend/app/controllers/api/Resource.scala
+++ b/backend/app/controllers/api/Resource.scala
@@ -34,11 +34,11 @@ class Resource(val controllerComponents: AuthControllerComponents, manifest: Man
                index: Index, pages: Pages, annotations: Annotations, previewStorage: ObjectStorage) extends AuthApiController {
 
   def getResource(uri: Uri, basic: Boolean, q: Option[String]) = ApiAction.attempt  { implicit req =>
-    val query = q.map(Chips.parseQueryString)
+    val parsedChips = q.map(Chips.parseQueryString)
 
-    val resourceFetchMode = (basic, query) match {
+    val resourceFetchMode = (basic, parsedChips) match {
       case (true, _) => ResourceFetchMode.Basic
-      case (false, text) => ResourceFetchMode.WithData(text)
+      case (false, pt) => ResourceFetchMode.WithData(pt.map(_.query))
     }
 
     val decodedUri = Uri(UriEncoding.decodePath(uri.value, StandardCharsets.UTF_8))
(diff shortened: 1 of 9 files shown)


frontend/src/js/components/workspace/Workspaces.tsx
```diff
@@ -594,6 +598,27 @@ class WorkspacesUnconnected extends React.Component<Props, State> {
...
+                                    "",
+                                    {
+                                        n: "Workspace Folder",
+                                        v: entry.name,
+                                        op: "+",
+                                        t: "workspace_folder",
+                                        workspaceId: workspace.id,
+                                        folderId: entry.id,
+                                    },
+                                    "*"
```
philmcmahon, line 615: I was thinking this might have been a mistake but actually a good shout to start with * so that there are documents visible straight away before the user adds a search

backend/app/controllers/api/Resource.scala
```diff
@@ -34,11 +34,11 @@ class Resource(val controllerComponents: AuthControllerComponents, manifest: Man
...
 
   def getResource(uri: Uri, basic: Boolean, q: Option[String]) = ApiAction.attempt  { implicit req =>
-    val query = q.map(Chips.parseQueryString)
+    val parsedChips = q.map(Chips.parseQueryString)
 
-    val resourceFetchMode = (basic, query) match {
+    val resourceFetchMode = (basic, parsedChips) match {
       case (true, _) => ResourceFetchMode.Basic
-      case (false, text) => ResourceFetchMode.WithData(text)
+      case (false, pt) => ResourceFetchMode.WithData(pt.map(_.query))
```
philmcmahon, line 41: maybe `pc` instead of `pt`?

philmcmahon:
This is looking great, would be good to do a playground test just to check for any url weirdness on a proper domain but otherwise code is looking good to me

frontend/src/js/components/workspace/Workspaces.tsx
zekehuntergreen: yep that's what I was thinking. Still want to test this in code with a workspace with loads of documents to see if it slows down the initial folder search.

backend/app/controllers/api/Resource.scala
zekehuntergreen: parsed tchip

backend/app/model/frontend/Chip.scala
```diff
@@ -46,14 +53,29 @@ object Chip {
...
+case class ParsedChips (query: String, workspace: Option[WorkspaceSearchContextParams])
+
 object Chips {
   // TODO - when the index supports attempts we can uncomment these lines to make this attempty
   //def parseQueryString(q: String): Attempt[String] = {
-  def parseQueryString(q: String): String = {
+  def parseQueryString(q: String): ParsedChips = {
     //Attempt.catchNonFatal {
+    val parsedQ = Json.parse(q)
+    // find WorkspaceSearchContextParams
```
philmcmahon, line 64: it might be worth adding a slightly more detailed comment here explaning what's happening - i.e extracting the workspace parameters, then ignoring them in the next bit

philmcmahon (approved):
cat looking in a folder

marjisound (approved):
This looks great, thanks for walking me through the PR :)

hoyla:
Woohoo!

PR #206, opened 2024-03-21
Create cloudwatch metric when user does workspace search

zekehuntergreen:
## What does this change?
Creates a cloudwatch metric when a user does a workspace search. This will allow us to make a grafana dashboard to track user adoption of the workspace search feature.


marjisound (approved):
Looks great 🎉

PR #209, opened 2024-05-15
Upgrading giantUI to node 18

srbd:
## What does this change?

Upgrading because we need as a dependency for a [snyk vulnerability in pdfjsdist](https://security.snyk.io/vuln/SNYK-JS-PDFJSDIST-6810403) and also we should upgrade anyway

## How to test

Tested on CODE


itsibitzi (approved):
1 character PRs ftw 😎

PR #220, opened 2024-06-21
Fix integration test

marjisound:
## Accessibility


.github/workflows/build.yml
```diff
@@ -27,22 +27,33 @@ jobs:
...
-        with:
-          java-version: '11'
-          distribution: 'adopt'
-          cache: sbt
+#      - name: Setup Java
+#        uses: actions/setup-java@v4
+#        with:
+#          java-version: '11'
+#          distribution: 'adopt'
+#          cache: sbt
```
philmcmahon, line 56: given that the build passed does this mean that we don't need to run this? I would have thought we would still need this step

backend/test/controllers/api/WorkspacesITest.scala
```diff
@@ -41,21 +58,41 @@ class WorkspacesITest extends AnyFunSuite with Neo4jTestService with Elasticsear
...
+//    // would work to isolate mutable variables between tests.
+//    // Whereas with instance vars inside the Helpers library, I suspect they
+//    // would've ended up shared between test instances (since it's a static object, not
+//    // instantiated each time WorkspaceITest is instantiated).
+//    userControllers = setupUserControllers(
+//      usernames = Set("paul", "barry", "jimmy"),
+//      neo4jDriver,
+//      elasticsearch = this
+//    )
+//  }
```
philmcmahon, line 85: remove?

backend/test/test/integration/Neo4jTestContainer.scala
```diff
@@ -0,0 +1,26 @@
...
+trait Neo4jTestContainer extends AttemptValues {
+  implicit def patience = PatienceConfig(Span(30, Seconds), Span(250, Millis))
+
+  def getNeo4jContainer() = {
+    val neo4jContainerDef = Neo4jContainer.Def(
+      dockerImageName = DockerImageName.parse("neo4j/neo4j-arm64-experimental:3.5.30").asCompatibleSubstituteFor("neo4j")
+    )
+//    val neo4jContainerDef = Neo4jContainer.Def(
+//      dockerImageName = DockerImageName.parse("neo4j:3.3.1")
+//    )
```
philmcmahon, line 17: remove

.github/workflows/build.yml
marjisound: As we discussed in the call earlier, I added the setup-java job before the sbt installation. Inspired from this https://eed3si9n.com/setup-sbt/

philmcmahon (approved):
Fantastic work on this, great to have the integration tests running more reliably and to try out the new ARM runners

backend/test/services/index/ElasticsearchEventsITest.scala
```diff
@@ -1,54 +1,73 @@
...
+  var elasticsearchTestService: ElasticsearchTestService = _
+
+  override def startContainers(): Containers = {
+    val elasticContainer = getElasticSearchContainer()
+    val url = s"http://${elasticContainer.container.getHttpHostAddress}"
+
+    elasticsearchTestService = new ElasticsearchTestService(url)
+
+    elasticContainer
+  }
```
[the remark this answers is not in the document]
marjisound: This could end up time consuming to find a good way of reducing the repetition, so for now lets leave it and if we get a chance we can improve this in an upcoming PR

PR #225, opened 2024-07-12
upgrade d3 from 4.13.0 to v7.9.0

marjisound:
## What does this change?

This PR upgrades d3 library. The only usage of this library is in https://github.com/guardian/giant/blob/main/frontend/src/js/components/EmailBrowser/Timeline.js#L5 

I tested this locally and the email threads are rendering correctly. 

## Why 
We need to upgrade d3-colour to 3.1.0 due to vulnerability https://github.com/guardian/giant/security/dependabot/2
But d3 library depends on d3-colour and to be able to update d3-colour we had to upgrade d3 from v4 to v7.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -14,7 +14,7 @@
         "@emotion/css": "^11.11.2",
         "@emotion/react": "^11.11.1",
         "connected-react-router": "^6.8.0",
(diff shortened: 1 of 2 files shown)


dominickendrick (approved):
Tested locally and in CODE ! Great Job

PR #226, opened 2024-07-17
upgrade elastic/charts from v57 to v66

marjisound:
## What does this change?

Upgrade elastic/charts from `v57.0.0` to `v66.1.0`. The only usage of this dependancy is in `TimeHistogram.js` and the upgrade does not have any breaking changes in our code base apart from the chart colour which I fixed. 

I updated this as an attempt to fix the `d3-color` vulnerability in https://github.com/guardian/giant/security/dependabot/2 that's suggesting the vulnerable version is introduced by elastic/charts. But it seems that this vulnerebility is also introduced by `d3-scale` which is also a sub dependancy of the latest version of elastic/charts. So this change doesn't seem to be fixing the main issue  of d3-color vulnerability. But it's still good to have the most recent version 

## How to test

tested locally

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -8,7 +8,7 @@
       "name": "giant-ui",
       "version": "0.1.0",
       "dependencies": {
-        "@elastic/charts": "57.0.0",
+        "@elastic/charts": "66.1.0",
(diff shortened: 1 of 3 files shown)


philmcmahon (approved):
👍🏻

PR #236, opened 2024-09-24
Add confirmation modal for removing from workspace

marjisound:
## What does this change?

This PR adds a confirmation modal for removing items from workspace. If it's a single file to be removed, the confirmation message displayed to user is: 
"**This will remove the file [filename] from the current workspace. It cannot be undone. Are you sure you want to proceed?**" 

If it’s any other type (folder, zip, email database that can be “expanded” etc), the message is:
"**This will remove the selection [item name] and everything nested inside it from your workspace. It cannot be undone. Are you sure you want to proceed?**"

The PR also updated the delete modal message to be consistent with the remove modal. 
ezgif-7-ef995e90b0

## How to test

Tested locally


philmcmahon (approved):
Nice work

PR #237, opened 2024-09-27
Make the arm64 neo4j container the default

philmcmahon:
## What does this change?

All developers on the team are using apple silicon.This change makes the arm version of the neo4j image the default, which means we'll be able to run `docker-compose up` without having to set the NEO4J_OVERRIDE environment variable 

## How to test
Tested on my local machine. This should have no impact on playground/prod

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   neo4j:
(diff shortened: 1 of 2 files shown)


marjisound (approved):
Looks good 👍

PR #247, opened 2024-11-08
Upgrade to Panda v7 - support key rotation

rtyley:
This upgrades Panda from v3 to v7, increasing security by allowing us to use key-rotation as introduced with guardian/pan-domain-authentication#150.

I'm very happy to do a walkthrough of the changes with the reviewer if that helps!

### Changes to Panda that affected code in this project

* Panda v5
  * https://github.com/guardian/pan-domain-authentication/pull/147 removed the old `PublicKey` & `PrivateKey` classes in our `com.gu.pandomainauth` package, in favour of using the existing `java.security` classes. To create instances of those classes, we can use the `SettingsReader.{privateKeyFor, publicKeyFor}` methods.
* Panda v6:
  * https://github.com/guardian/pan-domain-authentication/pull/152 means the `CookieUtils.generateCookieData()` method now communicates errors with `CookieResult` values containing `CookieIntegrityFailure`, rather than exceptions.
* Panda v7:
  * https://github.com/guardian/pan-domain-authentication/pull/150 means that code shouldn't directly reference private or public keys anymore (eg do not reference `settings.signingKeyPair`). Instead, use `settings.signingAndVerification` or `publicSettings.verification`. Note also that `publicSettings.publicKey` was previously optional, and `publicSettings.verification` is not.

### DIFF
--- a/backend/app/AppComponents.scala
+++ b/backend/app/AppComponents.scala
@@ -124,8 +124,7 @@ class AppComponents(context: Context, config: Config)
         val publicSettings = new PublicSettings(config.publicSettingsKey, config.bucketName, pandaS3Client)
         // start polling of S3 bucket for public key
         publicSettings.start()
-        val getCurrentPublicKey: () => Option[pandomainauth.PublicKey] = () => publicSettings.publicKey
-        new PanDomainUserProvider(config, getCurrentPublicKey, users, metricsService)
+        new PanDomainUserProvider(config, () => publicSettings.verification, users, metricsService)
     }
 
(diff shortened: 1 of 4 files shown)


philmcmahon:
Wow thank you Roberto! I'm just deploying this to our test environment now...

rtyley:
> Wow thank you Roberto! I'm just deploying this to our test environment now...

Ah brilliant! Let me know how it goes...

philmcmahon (approved):
This is working fine on our playground environment, let's go! 

Thanks again @rtyley 

panda gif

PR #258, opened 2025-04-11
Hotfix type error in download modal

zekehuntergreen:
## What does this change?
Fixes type error in DownloadModal component when props.resource.text is undefined

## How to test
Testing in prod


srbd (approved):
nice catch

PR #270, opened 2025-08-07
Tidy up highlighting for giant plain text view

philmcmahon:
## What does this change?
I need to do a bit more history on this because it's possible that there was a very good reason highlighting was working the way it was before. BUT this pr fixes an issue where if you have a multi line highlight then the highlight messes up the flow of the text.

Heavily inspired by https://stackoverflow.com/questions/70008968/css-custom-text-highlight-that-works-across-multiple-lines

Before:
Screenshot 2025-08-07 at 14 49 49

After:
Screenshot 2025-08-07 at 14 49 30

This pr also removes some duplicated code - result-highlight was repeated twice, looks like maybe an accident when it was moved from typography.scss to document.scss here https://github.com/guardian/pfi/pull/818

## How to test
I've tested this locally, will try out on playground as well

### DIFF
--- a/frontend/src/stylesheets/components/_document.scss
+++ b/frontend/src/stylesheets/components/_document.scss
@@ -195,21 +195,12 @@
 
 result-highlight {
   position: relative;
-  display: inline-block;
   font-weight: bold;
   font-style: italic;
   z-index: 1;
-
-  &:before {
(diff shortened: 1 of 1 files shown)


joelochlann (approved):
Neat!

PR #239, opened 2024-10-02
WIP - Use transcription service

marjisound:
## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #274, opened 2025-08-07
Fix 'reprocess this file' for externally processed files.

philmcmahon:
## What does this change?
When we completed the external transcription service work in https://github.com/guardian/giant/pull/239 we introduced the concept of a 'PROCESSING_EXTERNALLY' relation - which is similar to a TODO in that it has an 'attempts' property - which could in theory allow for giant to repeatedly send a job to the external transcription service until it gets a success response. 

However, we neglected to update the bits of Neo4j code that increment the 'attempts' property so that they work with PROCESSING_EXTERNALLY relations as well as TODOs. We also didn't update the job that resets the attempts count to 0 when the user sends a file to be reprocessed.

From a user side, this means that currently in giant if you try and reprocess a file that has been through the external transcription service then it will fail (you can test this by trying to reprocess any audio file in giant).

This PR resolves that, by updating the relevant neo4j queries to work with PROCESSING_EXTERNALLY as well as TODOs

## How to test
I've tested this on playground - reprocessing is now working for externally transcribed stuff and the PROCESSING_EXTERNALLY relation has 'attempts' set to 1

### DIFF
--- a/backend/app/services/manifest/Neo4jManifest.scala
+++ b/backend/app/services/manifest/Neo4jManifest.scala
@@ -445,23 +445,23 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
         |  WITH
         |    worker
         |
-        |MATCH (e: Extractor)-[todo: TODO|:PROCESSING_EXTERNALLY]->(b: Blob:Resource)
+        |MATCH (e: Extractor)-[todo_or_pe: TODO|:PROCESSING_EXTERNALLY]->(b: Blob:Resource)
         |  WHERE
-        |    NOT (b)-[:LOCKED_BY]->(:Worker) AND todo.attempts < {maxExtractionAttempts}
+        |    NOT (b)-[:LOCKED_BY]->(:Worker) AND todo_or_pe.attempts < {maxExtractionAttempts}
         |
-        |  WITH worker, todo, e, b
+        |  WITH worker, todo_or_pe, e, b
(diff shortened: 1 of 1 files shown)


joelochlann:
I'm sure you thought about this when implementing `:PROCESSING_EXTERNALLY`, but just enumerating all the functions that still refer to `:TODO` but not `:PROCESSING_EXTERNALLY`, just in case any of them actually need to take account of the other one too:

- `insertBlob`
- `releaseLocks`
- `releaseLocksForTerminatedWorker`
- `markAsComplete`
- `getMimeTypesCoverage`
- `getWorkCounts`
- `setProgressNote`

and these ones to do with re-processing:
- `rerunSuccessfulExtractorsForBlob`
- `rerunFailedExtractorsForBlob`

philmcmahon:
Just had a follow up discussion with @joelochlann and agreed that this isn't fit for purpose - we don't want to mix up TODOs and PROCESSING_EXTERNALLY. I'll have another look at this and see if there is a way we can get reprocessing working for external transcription jobs whilst keeping them separate from the way giant handles normal work

joelochlann (changes requested):
As discussed IRL - the danger with this is that it will start registering work that should be processed externally (via the queue/container system) as work that should be processed by spinning up worker instances.

This could cause issues.

One solution we discussed verbally would be: set attempts to 1 at the point where we create the PROCESSING_EXTERNALLY relation when handling the task queue, instead of mingling this in with code that is geared towards TODOs and worker instances.

PR #275, opened 2025-08-12
Support reprocessing for failed external extraction jobs

philmcmahon:
## What does this change?
In https://github.com/guardian/giant/pull/239 when we introduced 'external extractors' so that we could run transcription via the transcription-service rather than directly on giant workers, we didn't think about reprocessing. This PR implements that. It's a bit complicated how it works, I've attempted to explain below.

In my testing, I found that the neo4j manifest for a blob can end up in different **failure** states based on what goes wrong.

1. Failure to send a message to the transcription service (e.g. if the queue doesn't exist - easy to reproduce locally)
2. Transcription service runs but returns an error

In state 1, you end up with relations that look like this:
Screenshot 2025-08-12 at 12 02 42

In state 2, they look like this (3 failure events! This is because the way I forced the transcription service to fail meant it kept retrying 3 times - see https://trello.com/c/N9v5DcMo/3320-external-transcription-failures-currently-result-in-3-extractionfailed-relations-being-created for details):

Screenshot 2025-08-12 at 12 03 40

To 'reset' the blob's relations so that it gets resent to the transcription service we need to:

In failure state 1 - delete the EXTRACTION_FAILURE relation, reset the 'attempts' of the TODO to 0
In failure state 2 - delete the EXTRACTION FAILURE relation(s), create a TODO with the _same_ properties as the PROCESSING_EXTERNALLY relation, except with 'attempts' set to 0, delete the PROCESSING_EXTERNALLY relation

That is what this PR does. Note that it does not deal with 'reprocess' for succesful transcription jobs - only failures - that's another bit of work to do. It also doesn't deal with jobs that don't receive a response from the transcription service (these would have a PROCESSING_EXTERNALLY relation indefinitely). I think that's ok though, as it is still a step forwards from where we are at the moment as it allows us to send failed transcription jobs for reprocessing

## How to test
I've tested this on playground. To do so I deployed the branch `worker-always-error` to the transcription service (https://riffraff.gutools.co.uk/deployment/view/cc45cb57-6608-48ab-a917-bef8790cd039) then ran some jobs, then redeployed main and sent some of the failed jobs in https://playground.pfi.gutools.co.uk/workspaces/146ace05-76b0-4a7a-8a17-78742074b20d for reprocessing

### DIFF
--- a/backend/app/services/manifest/Neo4jManifest.scala
+++ b/backend/app/services/manifest/Neo4jManifest.scala
@@ -1023,7 +1023,7 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
   }
 
   override def rerunFailedExternalExtractorsForBlob(uri: Uri): Attempt[Unit] = attemptTransaction { tx =>
-    // Re-run extractors that failed.
+    // Re-run external extractors that failed.
     // this function can deal with multiple failure scenarios:
     //  - Giant fails to send a message to the transcription service = 1 TODO to reset, 1 EXTRACTION_FAILED to delete
     //  - the transcription service fails to transcribe the file = 1 TODO to create, 1 PROCESSING_EXTERNALLY to delete, 1 or more EXTRACTION_FAILED to delete


backend/app/services/manifest/Neo4jManifest.scala
```diff
@@ -1022,6 +1022,54 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
     })
   }
 
+  override def rerunFailedExternalExtractorsForBlob(uri: Uri): Attempt[Unit] = attemptTransaction { tx =>
+    // Re-run extractors that failed.
```
joelochlann, line 1026:
```suggestion
    // Re-run external extractors that failed.
```

joelochlann (approved):
Having squinted at it for a while, I think this looks good! Though my understanding of the relationship between the TODOs and the task queue is a little hazy

The explanation and diagrams in the PR description are very helpful

philmcmahon:
> I guess it's confusing to have a `PROCESSING_EXTERNALLY` relation on something that is not actually processing, it could've failed a long time ago.
> 
> But then I think that's the same model we have with `TODO`s right? We leave them in place on failure - I think so they can be used to re-process

Agreed, I think we need a 'timeout' job that will throw away something that has been PROCESSING_EXTERNALLY for more than X hours and mark it as a failure

PR #276, opened 2025-08-13
Don't fail external reprocess job if there are no errors to reprocess.

philmcmahon:
## What does this change?
https://github.com/guardian/giant/pull/275 introduced a bug where if you send a file (blob) for reprocessing that doesn't have an EXTRACTION_FAILURE relation with an external extractor then an error is returned every time. This resolves that by changing the error condition - if we haven't deleted any relations then we don't have to worry about how many counters were reset.

## How to test
Reprocessing is completely broken on PROD right now, this will fix that

### DIFF
--- a/backend/app/services/manifest/Neo4jManifest.scala
+++ b/backend/app/services/manifest/Neo4jManifest.scala
@@ -1057,7 +1057,9 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
       val relationshipsDeleted = counters.relationshipsDeleted()
       val propertiesSet = counters.propertiesSet()
 
-      if (relationshipsDeleted  < 1 || propertiesSet < 1 ) {
+      // if the blob has an EXTRACTION_FAILURE relation then that will be deleted and we expect to set at least one property
+      // (1 property if there is a TODO, more if there's a PROCESSING_EXTERNALLY relation to replace with a TODO)
+      if (relationshipsDeleted != 0 && propertiesSet < 1 ) {
         Attempt.Left(IllegalStateFailure(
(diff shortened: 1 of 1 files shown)


backend/app/services/manifest/Neo4jManifest.scala
```diff
@@ -1057,9 +1057,11 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
       val relationshipsDeleted = counters.relationshipsDeleted()
       val propertiesSet = counters.propertiesSet()
 
-      if (relationshipsDeleted  < 1 || propertiesSet < 1 ) {
+      // if the blob has an EXTRACTION_FAILURE relation then that will be deleted and we expect to set at least one property
+      // (1 property if there is a TODO, more if there's a PROCESSING_EXTERNALLY relation to replace with a TODO)
+      if (relationshipsDeleted != 0 && propertiesSet < 1 ) {
```
zekehuntergreen, line 1062: In the case where we're reprocessing a file that doesn't have an EXTRACTION_FAILURE, do we still want to check that propertiesSet > 0? If so, maybe this condition could just be `if (propertiesSet == 0)`?

zekehuntergreen (approved):

backend/app/services/manifest/Neo4jManifest.scala
philmcmahon: In that case, we don't want to check propertiesSet at all as we don't expect any changes from this function.

I think that this is confusing as one function is essentially doing two different things depending on what's happening in the database. I'll add a ticket to split this up into something easier to follow

PR #278, opened 2025-08-21
Gracefully handle multiple transcription_success messages

philmcmahon:
## What does this change?
There is a (I think rare) but possible situation where this can happen:
 - a file is uploaded to giant and transcribed externally, resulting in a relation like this:

Screenshot 2025-08-21 at 16 46 45

If the file is sent for reprocessing **twice**, then that relation will be replaced with a PROCESSING_EXTERNALLY relation and two transcription workers will be spun up for each reprocess job.

Both workers will finish, and then post messages on the SQS queue with the completed transcript. The first one will be happily received by giant, the next would (currently) result in an error like this:

`Failed to mark '<URI>' processed by ExternalTranscriptionExtractor as complete: Unexpected number of creates/deletes in markAsComplete. Created: 0. Deleted: 0`

Now, various things have gone wrong here - how was the file sent for reprocessing twice? Shouldn't the reprocess function detect that the file is already being processed and stop? I think maybe not - it's always nice to have the option of saying "I don't care what's happening right now, do it again" as giant ingest jobs should be idempotent. 

I think that giant should gracefully handle situations where it gets multiple success messages for the same file, which is what this PR achieves.

## How to test
I've tested this locally, there are a lot of examples on PROD currently we can try it out with

### DIFF
--- a/backend/app/services/manifest/Neo4jManifest.scala
+++ b/backend/app/services/manifest/Neo4jManifest.scala
@@ -570,40 +570,61 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
   override def markExternalAsComplete(uri: String, extractorName: String): Either[Failure, Unit] = transaction { tx =>
 
     logger.info(s"Marking ${uri} / ${extractorName} as complete")
-    val result = tx.run(
-      s"""
-         |MATCH (b :Blob:Resource {uri: {uri}})<-[processing:PROCESSING_EXTERNALLY]-(e: Extractor {name: {extractorName}})
-         |
-         |MERGE (b)<-[processed:PROCESSED {
-         |  ingestion: processing.ingestion,
-         |  parentBlobs: processing.parentBlobs,
-         |  languages: processing.languages
-         |  }]-(e)
-         |
-         |FOREACH (_ IN CASE WHEN exists(processing.workspaceId) THEN [1] ELSE [] END |
-         |  SET processed.workspaceId = processing.workspaceId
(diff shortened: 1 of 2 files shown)


backend/app/services/manifest/Neo4jManifest.scala
```diff
@@ -570,40 +570,61 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
...
     )
 
-    val counters = result.summary().counters()
+    val processedCount = existingProcessed.single().get("count").asInt()
 
-    if (counters.relationshipsCreated() != 1 || counters.relationshipsDeleted() != 1) {
-      Left(IllegalStateFailure(s"Unexpected number of creates/deletes in markAsComplete. Created: ${counters.relationshipsCreated()}. Deleted: ${counters.relationshipsDeleted()}"))
-    } else {
+    if (processedCount > 0) {
+      logger.info(s"There is already a processed relation between blob ${uri} and extractor ${extractorName}. Doing nothing.")
```
zekehuntergreen, line 589: If this is the second time that the file has been processed externally do we not want to perform any of the actions below like deleting the "processing" link?

zekehuntergreen (approved):

backend/app/services/manifest/Neo4jManifest.scala
philmcmahon: I really hope that I'm right! (two weeks later new pr "deal with duplicate PROCESSING_EXTERNALLY relations")

PR #279, opened 2025-08-27
Use combined transcript file from transcription service

philmcmahon:
## What does this change?
This PR gets giant to make use of the new 'combined output' file added to the transcription service output in https://github.com/guardian/transcription-service/pull/167 

As well as simplifying some of our types (no longer need for the 'bucketurls' types with srt/txt/json), this change will make it easy to start storing other output types from the transcription tool in giant. In particular we're interested in subtitle formats, which will enable us to add subtitles to giant's video player and to see the timecode of matches for giant searches. 

## How to test
I've tested this by deploying to playground, it's a bit of an 'all or nothing' change so I am fairly confident that it will work.

### DIFF
--- a/backend/app/utils/Whisper.scala
+++ b/backend/app/utils/Whisper.scala
@@ -6,7 +6,7 @@ import java.nio.file.Path
 import scala.io.Source
 import scala.sys.process._
 
-case class TranscriptionResult(text: String, language: String)
+case class WhisperResult(text: String, language: String)
 
 object Whisper extends Logging {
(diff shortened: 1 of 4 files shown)


backend/app/utils/Whisper.scala
```diff
@@ -6,7 +6,7 @@ import java.nio.file.Path
 import scala.io.Source
 import scala.sys.process._
 
-case class TranscriptionResult(text: String, language: String)
+case class WhisperResult(text: String, language: String)
```
philmcmahon, line 9: renamed as we need to use 'TranscriptionResult' for the type of the combined output

PR #280, opened 2025-09-02
Show timecodes of transcribed files in giant

philmcmahon:
## What does this change?
We'd like to make it easier for users of giant to navigate between a transcript and the associated media.  This PR:
 - gets giant to store the SRT output (which includes timecodes) from the transcription service in elasticsearch
 - exposes the SRT data in the UI

It looks like this (eep I already found a bug - the bottom right view selector highlighting is broken doh!):
Screenshot 2025-09-02 at 17 49 28

Confusingly, I've referred to the SRT data as 'VTT' everywhere. This is because eventually I'm planning to swap out the SRT data for [Web VTT](https://gist.github.com/samdutton/ca37f3adaf4e23679957b8083e061177) eventually, as this is a slightly cleaner format, and can also be used to provide subtitles to the  video player we use for media playback. That will require a transcription service update, and changes to the transcription service are currently blocked whilst we fix the whisper.cpp docker container

### Search functionality
Following @joelochlann's suggestion below, I decided to include the vtt transcript in giant's search functionality. This might have some performance impact, as there is now double the amount of text to search for each transcript, but I think elasticsearch can probably cope. The advantage is that, where a search term has a match both in the vtt file and the text, you can seamlessly switch back and forth between them:

https://github.com/user-attachments/assets/b09c5b57-6929-4221-ad79-f2920585883c

## How to test
Deploy this branch to playground and take a look at this workspace https://playground.pfi.gutools.co.uk/workspaces/773059f4-b325-4a61-b3eb-376a6cd08cb0


backend/app/commands/GetResource.scala
```diff
@@ -61,6 +61,7 @@ case class GetResource(uri: Uri, mode: ResourceFetchMode, username: String, mani
           indexed <- index.getResource(uri, query)
           comments <- annotations.getComments(uri)
         } yield {
+          println(indexed)
```
joelochlann, line 64: debug line

backend/app/AppComponents.scala
```diff
@@ -229,7 +230,7 @@ class AppComponents(context: Context, config: Config)
 
       // external extractor
       val externalWorker = new ExternalTranscriptionWorker(manifest, sqsClient, config.transcribe, transcriptStorage, esResources)
-      val externalWorkerScheduler = new ExternalWorkerScheduler(actorSystem, externalWorker, config.worker.interval)(workerExecutionContext)
+      val externalWorkerScheduler = new ExternalWorkerScheduler(actorSystem, externalWorker, FiniteDuration.apply(5, TimeUnit.SECONDS))(workerExecutionContext)
```
joelochlann, line 233: Maybe should move to config as well eventually? Unless it's just something which never varies between deployments, which it might be

PR #283, opened 2025-09-05
Add url to Giant part 1 - Remote Ingest model, worker and endpoint

philmcmahon:
## What does this change?
We want to make it easier for giant users to quickly archive web pages to giant. The plan is to add a new 'add url to workspace' button which, rather than taking a file upload, takes a url. Giant will then send off to third party services to download relevant content from the provided url, and store the results in the workspace requested by the user.

This first step towards that does this stuff:
 - New endpoint /api/workspaces/:workspaceId/remoteUrl
 - New RemoteIngest type - stored in neo4j. I decided to put the queries for this new type in a new class as Neo4jManifest is massive. I'm not sure 'manifest' is the right work here, maybe can drop that from the name for Neo4jRemoteIngestManifest...
 - New Remote Ingest worker which
     -  polls neo4j for RemoteIngest jobs and sends any new remote ingest jobs to SQS (initially the media download service task queue)
     - polls the results queue, and logs out any messages **(TODO: get this to actually ingest the files into giant)**
 

Apologies for the edits to PostgresClient - they're all whitespace! I can tidy that up. 

## How to test
I've tested this locally, it will need associated infrastructure to get running on playground - so this is a draft PR for now


twrichards (commented):
looking good.

what's the benefit of polling the table then placing on outbound queue VS just writing to the outbound queue at the point of request?? similarly why not have a thread thats constantly pulling from the inbound queue (see https://github.com/guardian/grid/blob/2cb2766393d106da4efa9b49233a0a2aecbad23a/common-lib/src/main/scala/com/gu/mediaservice/lib/aws/SimpleSqsMessageConsumer.scala#L17 and https://github.com/guardian/grid/blob/2cb2766393d106da4efa9b49233a0a2aecbad23a/image-loader/app/controllers/ImageLoaderController.scala#L63)
... both to reduce lag

PR #285, opened 2025-09-11
[Save to giant part 1.5] actually ingest/store the result of remote ingest jobs

twrichards:
Continuation of https://github.com/guardian/giant/pull/283, with lot's of refactors to get us ready for https://github.com/guardian/giant/pull/401, notable changes being...

- factor out the need to provide collection and ingestion as input to the remoteUrl endpoint, and instead fetch the default collection for the user from Neo4J and reuse the job id for the `ingestion`
- actually ingest the resulting file(s) from remote ingest job
- Correctly pass username parameter in backend/app/services/users/Neo4jUserManagement.scala
- Create ingestion before attempting to ingest file to it
- Use separate bucket for remote ingest
- Upload to new remote ingest bucket
- Save blobUri, clean up file post ingestion
- Explicitly create folder when ingested file comes back
- Implement addOrGetFolder function to ensure remote ingest folder only gets created once
- Create remote-ingest bucket in local minio
- neo4j `insertRemoteIngest` operation AFTER successfully sending to queue
- model remote ingest 'added by' with proper relationship to User

### DIFF
--- a/backend/app/services/ingestion/RemoteIngestWorker.scala
+++ b/backend/app/services/ingestion/RemoteIngestWorker.scala
@@ -52,10 +52,10 @@ class RemoteIngestWorker(
               else Left(RemoteIngestFailure(s"Remote ingest job ${parsedJob.id} did not complete successfully. \n$parsedJob"))
             )
             job <- remoteIngestStore.getRemoteIngestJob(parsedJob.id)
-            workspaceMetadata <- annotations.getWorkspaceMetadata(job.userEmail, job.workspaceId)
+            workspaceMetadata <- annotations.getWorkspaceMetadata(job.addedBy.username, job.workspaceId)
             _ <- remoteIngestStore.updateRemoteIngestJobStatus(parsedJob.id, "INGESTING")
             ingestion <- Collections.createIngestionIfNotExists(Uri(job.collection), job.ingestion, manifest, index, pages, s3Config)
-            parentFolder <- annotations.addOrGetFolder(job.userEmail, job.workspaceId, job.parentFolderId, job.title)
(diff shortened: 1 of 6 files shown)


backend/app/services/ingestion/RemoteIngestWorker.scala
```diff
@@ -2,33 +2,100 @@ package services.ingestion
...
+              // TODO: Do we care if this sending to dead letter queue fails?
+              amazonSQSClient.sendMessage(config.outputDeadLetterQueueUrl, sqsMessage.getBody)
+              maybeParsedJob.map(parsedJob =>
+                remoteIngestStore.updateRemoteIngestJobStatus(parsedJob.id, "FAILED")
+              )
+            },
+            job => logger.info(s"Successfully ingested remote file for job with id ${job.id} in workspace ${job.workspaceId}")
+        ).onComplete { _ =>
+          amazonSQSClient.deleteMessage(config.outputQueueUrl, sqsMessage.getReceiptHandle)
+          //TODO: Delete file from S3
```
twrichards, line 98: this is probably crucial to fix before merging

philmcmahon: at this point we don't have the key for the file in s3, making it harder to delete. In the failure scenario we could have failed to fetch the item from neo4j, and we don't currently provide the key of the file in S3 in the message that comes back from SQS. There is a 7 day expiry on the bucket though as a fallback.

philmcmahon: (note we do delete the file in the success scenario)

twrichards: ahh great, shall we remove the TODO

PR #289, opened 2025-09-19
Shrink any pdfs with a page size larger than a1

philmcmahon:
## What does this change?
We've been running into problems with giant when trying to ingest PDFs containing large images. This is common for PDFs containing screenshots of entire web pages. When it tries to process these files, ocrmypdf crashes.

Helpfully, before it crashes we get a warning from the python image library 'pillow':

`DecompressionBombWarning: Image size (353800000 pixels) exceeds limit of 250000000`

After some investigation, I'm fairly sure this happens as a result of pdf files having a very large 'page size'. This isn't the resolution - the resolution of the images in the files causing giant to crash was way below the limit suggested above of 250 million pixels. This is the print size of the page. For the file I was testing the page size was `2772 x 4135.5 pts` - which in metres is roughly 1x1.5m

When we run ocrmypdf we set `--redo-ocr` to true, so the PDF gets turned into an image and then re-ocrd. The image size will be determined by the pdf page size and the dpi - which defaults to 300. With a page size like the one above we end up with 350 million pixels.

This PR aims to resolve that by capping the page size at A1. To determine the page size I'm using pdfinfo (https://www.xpdfreader.com/pdfinfo-man.html) - this has output that is relatively easy to parse. An alternative would have been to use `qpdf --json file.pdf` but that outputs the entire pdf as JSON, providing page size values for each page in the PDF. pdfinfo only checks the first page, but I'm hoping that will be enough for our purposes.

Why A1? Well it's quite big, hopefully big enough to still get decent quality OCR, I guess we'll find out...

## How to test
I've tested this on playground and it appears to solve the problem, there are some PDFS you can use for testing linked to from the trello ticket https://trello.com/c/54A8v8Ji/3334-fix-giant-ocrmypdf-extractor-crashing-for-certain-pdfs - will try and get some beter examples that I can attach here.

shrink it!


hoyla:
This should close #390

Brewfile
```diff
@@ -5,3 +5,4 @@ brew "ocrmypdf"
 brew "imagemagick"
 brew "qpdf"
 brew "ffmpeg"
+brew "xpdf"
```
zekehuntergreen, line 8: Seems like a shame to add yet another pdf-related dependency :/ but sounds like its worth avoiding the painful parsing with qpdf

zekehuntergreen (approved):
Looks good!

PR #405, opened 2025-10-15
Update to Scala 2.13.11

mbarton:
These warnings were introduced in Scala 2.13.11 for forward compatibility with Scala 3.

eg

```
home/michael/giant/backend/app/services/Config.scala:127:16: Implicit definition should have explicit type (inferred play.api.libs.json.OFormat[services.PostgresConfig])
 implicit val format = Json.format[PostgresConfig]
```

Also fix an `okhttp` deprecation warning in the CLI: `method create in class RequestBody is deprecated`

Running locally I just used Java 21 as that was what was on my machine until I noticed Giant is on Java 11. With 21 I had to upgrade to the newer Scala version to fix an odd bug building the compiler bridge on Java 21 (https://github.com/scala/bug/issues/12783).

No problem but I figured these changes were self contained enough that they were worth contributing back. Obv the actual Java upgrade for deployment is one for yourselves :)

### DIFF
--- a/backend/test/controllers/genesis/GenesisTest.scala
+++ b/backend/test/controllers/genesis/GenesisTest.scala
@@ -19,7 +19,7 @@ class GenesisTest extends AnyFreeSpec with Matchers with Results with AttemptVal
 
   private val controllerComponents = Helpers.stubControllerComponents()
 
-  implicit override val patienceConfig = PatienceConfig(scaled(Span(2, Seconds)), scaled(Span(15, Millis)))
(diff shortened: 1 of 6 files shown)


philmcmahon:
> Oh whoops sorry, looks like there's some test compile failures too. I can blast through them tonight

I think this will do the trick https://github.com/guardian/giant/commit/07dd9930899569172b3ff2643a7dd86b281eeb0b let's see if the build passes

philmcmahon:
That worked on playground @mbarton - just need that extra commit and we could merge this

PR #406, opened 2025-10-16
Fix genesis user creation in dev

### LINKED ISSUE #391: Giant genesis user flow is broken
Currently, with a fresh setup of giant the only way to create a genesis user is to go to localhost:3000/register - and that throws an error as well. Joe and I could only get it working by manually creating a user in neo4j

Philip McMahon [Aug 28, 2025, 1:37 PM](https://trello.com/c/Ig2Mrhzs/3154-giant-genesis-user-flow-is-broken#comment-68b04d7a14d90a887b9c5529): 
> Update today - actually creating users at all is broken, even if you’re already logged in as a root user

Luke Hoyland [Sep 25, 2025, 5:02 PM](https://trello.com/c/Ig2Mrhzs/3154-giant-genesis-user-flow-is-broken#comment-68d56788adbcc946d4bc33ae)

> @philipmcmahon1 do you mean creating users using the CLI? I create regular users quite often (though maybe not for a few weeks)

mbarton:
Fixes #391 

The create react app was not correctly proxying request to the `/setup` endpoint which the frontend uses to decide to show the "genesis user" flow on first startup.

It added a trailing slash to the proxied request and Play was fussy and complained that `/setup/` didn't exist (because it's only `/setup` in the routes).

I think this might have broken via a transitive upgrade to http-proxy, used by http-proxy-middleware. At least https://github.com/chimurai/http-proxy-middleware/issues/1016 suggests this is this case and http-proxy is not maintained.

Genesis user creation should still have been working for non dev builds as they don't use the create react app proxy.

WRT #391 I tested creating the genesis user then creating another user and was able to log in as both. Only tested in dev though.

### DIFF
--- a/frontend/src/setupProxy.js
+++ b/frontend/src/setupProxy.js
@@ -4,6 +4,16 @@ const { createProxyMiddleware }  = require('http-proxy-middleware');
 // (proxy everything without `Accept: text-html`) breaks PDF preview
 module.exports = function(app) {
   app.use('/api', createProxyMiddleware({ target: 'http://localhost:9001/api', changeOrigin: true }));
-  app.use('/setup', createProxyMiddleware({ target: 'http://localhost:9001/setup', changeOrigin: true }));
+  // Path rewrite required because http-proxy (unmaintained, upstream of http-proxy-middleware) adds a 
+  // trailing slash and then Play is fussy and claims the route doesn't exist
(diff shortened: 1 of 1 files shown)


hoyla:
> Genesis user creation should still have been working for non dev builds as they don't use the create react app proxy.

Aha! Thanks!

PR #460, opened 2025-12-15
workspace file moving revamp

twrichards:
_Follows https://github.com/guardian/giant/pull/462_

Move operations in big workspaces have occasionally brought down Neo4J (spiked CPU then unresponsive) and this is likely because moving many files results in separate requests, each of which trigger a `getWorkspace` call upon completion, which can be an expensive operation (recently even more complicated since additions to support 'Capture from URL')

## What does this change?
Dispatch the `PUT`s to move each item via 5 parallel batches, each of which moves items *sequentially* and also adds a throttle to those `getWorkspace` calls to be no more than once per second, which should still feel pretty real-time during a big file move operation (which takes time anyway).

To make this whole process clearer to the user, I've also added to the footer of workspace view:

- a display of the selection count (including descendants count) , when more than one item is selected
- a progress bar in the bottom right when a move is in progress

moving

Also if they try to close during a move operation, they'll be prompted if they're sure (unfortunately the message [isn't customisable in modern browsers](https://stackoverflow.com/a/1119324)).
image

Also cleans up an unused param on a vaguely related function.

## How to test
Move loads of files in a big workspace and see that it performs better 🤞

No crashes 🤞 and operations in big workspaces are quicker and more reliable.

### DIFF
--- a/frontend/src/js/components/workspace/Workspaces.tsx
+++ b/frontend/src/js/components/workspace/Workspaces.tsx
@@ -234,7 +234,7 @@ class WorkspacesUnconnected extends React.Component<Props, State> {
                       </ReactTooltip>
                     </>}
                     <ItemName canEdit={canEdit} id={entry.id} name={entry.name} onFinishRename={curryRename}/>
-                    {isWorkspaceNode(entry.data) && <FileAndFolderCounts {...entry.data} />}
+                    {isWorkspaceNode(entry.data) && <FileAndFolderCounts marginLeft="5px" {...entry.data} />}
                 </React.Fragment>;
             },
             sort: (a: TreeEntry<WorkspaceEntry>, b: TreeEntry<WorkspaceEntry>) => {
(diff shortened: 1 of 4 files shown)


philmcmahon (approved):
Assuming this works on playground, code LGTM

throttle

frontend/src/js/components/workspace/Workspaces.tsx
```diff
@@ -389,6 +393,24 @@ class WorkspacesUnconnected extends React.Component<Props, State> {
           }
           this.setState({haveAppliedWorkspaceLocationParam: true})
         }
+
+        const isMoveInProgress = this.state.itemsBeingMoved > 0;
+        const prevIsMoveInProgress = prevState.itemsBeingMoved > 0;
+        if(isMoveInProgress !== prevIsMoveInProgress) {
+          if(window.onbeforeunload) {
+            console.log("Removing beforeunload handler");
```
philmcmahon: could remove these console.logs?

philmcmahon (approved):
Nice work 👍🏻

PR #465, opened 2025-12-17
AWS SDK upgrade for autoscaling

srbd:
## What does this change?

Upgrades autoscaling to AWS SDK v2. Should be a no-op

## How to test

tested on playground - things scaled!


philmcmahon (approved):
💯 

autoscale

PR #476, opened 2025-12-29
Flip workspaces and datasets in search sidebar

hoyla:
Since people use workspaces far more than datasets, I want to make workspaces more visible in the search UI so people can more intuitively filter searches directly from that page. 

Changes like this should become redundant with the long-proposed UI overhaul but we might as well make this tiny change in the meantime.


philmcmahon (approved):
💯
flip pancake

PR #479, opened 2025-12-29
Add "Copy link to folder" to contextual menu

hoyla:
As of https://github.com/guardian/giant/issues/430 we're updating the URL to reflect a particular position in a workspace. To share this location, the user must to go to the location bar, copy that text (correctly), and then message it to a colleague. To facilitate doing so, and to surface the specific URL functionality better, this change adds an option to the contextual menu to get the URL of that location, if the contextual menu is being shown for a folder.


hoyla:
Because my incorrect Prettification has resulted in hundreds of lines of changes, we're parking this pending @philmcmahon and @twrichards proposals about implementing Prettier and eslint for Giant. There's nothing remotely urgent about this PR so it can wait until the linting is in place and then we can update these PRs to comply.

PR #502, opened 2026-01-21
Use brotli compression

philmcmahon:
## What does this change?
This PR introduces brotli compression to giant using https://github.com/guardian/play-brotli-filter

The main benefit will be that, for users on slow connections, we're sending a lot less data. For example there is a 75% reduction in the size of the workspaces JSON blob when opening this (large) workspace on playground: https://playground.pfi.gutools.co.uk/workspaces/794bff73-681c-4ab4-b821-d3bb47191a34

Screenshot 2026-01-21 at 13 16 09
Screenshot 2026-01-21 at 13 01 26

The main performance problem for workspaces isn't data transfer time so this won't result in drastic improvements to our larger workspaces unless a user is using a bad internet connection, but seems a nice change nevertheless

Depends on https://github.com/guardian/giant/pull/501

## How to test
Tested on playground - doesn't seem to cause any problems with viewing media/pdfs.

### DIFF
--- a/backend/app/AppComponents.scala
+++ b/backend/app/AppComponents.scala
@@ -2,7 +2,6 @@ import org.apache.pekko.actor.{ActorSystem, CoordinatedShutdown}
 import org.apache.pekko.actor.CoordinatedShutdown.Reason
 import cats.syntax.either._
 import com.amazonaws.client.builder.AwsClientBuilder
-
 import java.net.URI
 import software.amazon.awssdk.services.sns.SnsClient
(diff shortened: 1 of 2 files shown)


twrichards (approved):
that's a lovely reduction in data transfer and definitely useful for slow connections (e.g. my commute into London 😜 )

I think it would be worth measuring how much time that adds to the server (and client) but I think we could do this check by releasing this PR, I've recorded three loads of 17MB workspace... 

Image

PR #505, opened 2026-01-21
Workspace contextual menu: get link to here

### LINKED ISSUE #514: Workspaces: copy url within workspace via contextual menu
Now that we have URLs for every node in a workspace it would be good to have a menu option to copy that url to the clipboard, rather than rely on people to accurately select the complete URL from the location bar

tested locally and on playground

hoyla:
Closes #514 (I hope)

Reattempt of #479 without the linting issues.

Adds a "Copy link to file/folder" option to the workspace contextual menu, allowing users to copy a full URL of any workspace entry to share with colleagues. 

Since we needed to know we're working on files or folders, I updated a couple of other menu items to say "folder" instead of "file", where appropriate. 

And since we were doing something similar when users click on an item to update the location bar and when they click on the workspace title itself, I thought it'd better to consolidate those three sets of logic into one. 

### Changes

1. New menu item to copy file/folder link to clipboard via `navigator.clipboard.writeText`.
2. Detect whether the right-clicked entry is a file or folder and adjusts labels accordingly (e.g. "Copy folder name" / "Delete folder" instead of always saying "file").
3. Type guard refactor: `isWorkspaceLeaf` / `isWorkspaceNode` are called once at the top of `renderContextMenu` and stored as typed variables (`workspaceLeaf`, `workspaceNode`). This avoids redundant type guard calls and works around TypeScript not narrowing types through boolean aliases. I'm told.
4. The `/workspaces/:id/:entryId?` route was previously constructed in three different places: `setSelectedEntry` (click navigation), `getEntryLink` (copy link), and `handleTitleClick` in `WorkspaceSummary` (workspace title click). These are now consolidated into a single `workspaceEntryPath(workspaceId, entryId?)` helper in `workspaceUtils.ts`, so the route pattern lives in one place.

### Testing

Tested locally & on playground

### DIFF
--- a/frontend/src/js/components/workspace/Workspaces.tsx
+++ b/frontend/src/js/components/workspace/Workspaces.tsx
@@ -53,6 +53,7 @@ import {
 } from "../../util/treeUtils";
 import { setFocusedEntry } from "../../actions/workspaces/setFocusedEntry";
 import {
+  getEntryLink,
   processingStageToString,
   workspaceEntryPath,
   workspaceHasProcessingFiles,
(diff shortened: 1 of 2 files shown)


philmcmahon (approved):
@hoyla I made a few changes to this - rebased, cut down a bit on duplication and pulled getEntryLink out of the main Workspaces class because it doesn't rely on anything else in there so can live by itself - doesn't really change anything except that in a universe where we wrote some tests it would make it easier to test it without having to create a whole Workspaces class

I think the tweaks I made are significant enough to need a quick test on playground once the build finishes, otherwise this is good to go

PR #535, opened 2026-02-09
Ensure we always close PDDocument

philmcmahon:
## What does this change?
I noticed in the logs thousands of instances of `Warning: You did not close the PDF Document` (50k in last 24 hours). This is documented here https://pdfbox.apache.org/2.0/faq.html#why-do-i-get-a-%22warning%3A-you-did-not-close-the-pdf-document%22%3F - we're supposed to wrap the `PDDocument.load` in a try/finally.

I don't expect this to have a huge impact on giant performance but it will clean up the logs a bit. 

It might be nice to switch all of our try/finally uses to [Using](https://www.scala-lang.org/api/2.13.6/scala/util/Using$.html)  at some point but here I've used using where it's a lot less verbose and try/finally in GetPages where there was already a try/finally.

I've tested this on playground by deploying it and uploading a single file.


philmcmahon:
Screenshot 2026-02-10 at 10 52 57

PR #548, opened 2026-02-15
Add drag-and-drop file upload to workspace tree

### LINKED ISSUE #314: Drag and drop to trigger to upload to specific workspace or folder within a workspace ❤️❤️❤️❤️
This is one of those features that I would just love.

At present to upload to a workspace from your computer you have to select a target location in the workspace (which could be a subfolder, but this easily goes wrong), then hit the Upload To Workspace button, then choose to upload either files or folders, then navigate to the source through a find file dialog, select the files, then check that the candidates for upload are correct in the upload to giant dialog.

In many browser-based cloud file management systems (Google Drive, Mega, Dropbox WeTransfer), it’s possible to directly drag a selection (file/files/files and folder(s) with subfolder hieararchy preserved, etc) from the local file system into the target location in the browser window, and thereby trigger an upload to that location.

In the following video I drag a file into a google drive location, and hover the loaded cursor over subfolders therein. The highlighted area is selected as the target for the upload:

https://github.com/user-attachments/assets/3a44afb3-ab5f-4759-b72a-668772f96222

Decades of HCI research has shown that that object>action (i.e. picking a thing and then do something to it) is more natural and usually quicker than the inverse (decide to do something, then decide what to do it to). 
 
People have not stopped asking for "drag to upload" in all the years of Giant’s existence.

So let's make it possible to drag a file/files/folders from the Finder (or Windows Explorer or wherever) into a location displayed in Giant, to trigger an upload process. On releasing the mouse over the target location the files should not be immediately uploaded, but instead Giant’s existing confirmation dialog should be shown followed by the editable list of candidate files and the target location. This step has proved useful as a way of checking that the files I've asked to upload are indeed the files I meant to upload. Plus it gives a confirmation of the target directory. 

Image

then

Image

See what I mean? It makes you happy just imagining it doesn't it? Dear gosh life is good

hoyla:
Addresses #314.

Users can now drag files and folders from their file system directly onto any folder in the workspace tree (or the empty area below it) to trigger an upload to that location. The upload modal opens pre-populated with the dropped files and targeted at the correct workspace folder.

### What this does

- New shared utility `dropZoneUtils.ts`:
- - `readFilesFromDragEvent` reads files and recursively reads directories from the `DataTransfer`
- - `dragEventContainsFiles` distinguishes file-system drops from internal tree item drags.
- Detects file-system drops via a new `onDropFiles` prop, falling through to the existing internal-move logic otherwise.
- Resolves the target folder from the drop location and sets `droppedFiles` state, which flows through WorkspaceSummary to UploadFiles.
- *Consumes `droppedFiles` via a `useEffect` to prepopulate the existing upload dialog.

### Key implementation details

- `FileSystemEntry` references are captured synchronously before any async work. Browsers invalidate the `DataTransfer` after the synchronous event handler returns, so awaiting inside the `dataTransfer.items` loop would result in us silently losing all items beyond the first. 
- Chrome and Safari cap directory reads at ~100 entries per call. `readDirectoryEntry` now loops until an empty batch is returned to ensure that large directories are fully read.
- Browsers that don't support `webkitGetAsEntry()` fall back to `getAsFile()`.
- `findNodeByID` has been moved to a util since both this work and #658 use the same workspace position logic.

### Tested locally and on playground with

- Single folder (containing subfolders and files)
- Multiple files
- Multiple folders
- Mixed selection of files and folders

### DIFF
--- a/frontend/src/js/components/Uploads/dropZoneUtils.ts
+++ b/frontend/src/js/components/Uploads/dropZoneUtils.ts
@@ -1,0 +1,124 @@
+import React from "react";
+import {
+  readFileEntry,
+  readDirectoryEntry,
+  FileSystemEntry,
+  FileSystemFileEntry,
+  FileSystemDirectoryEntry,
+} from "./FileApiHelpers";
+
+/**
+ * Reads files from a drag event - handles both direct file drops and directories.
+ * Works across all major browsers on all platforms using the File System Access API.
+ *
+ * Only two kinds of selection are accepted:
+ *  1. A single directory (its full hierarchy is preserved); or
+ *  2. One or more files with no directories.
+ *
+ * Any other combination (e.g. mixed files and folders, or multiple folders) is
+ * rejected to avoid the ambiguous duplicate-entry problem that arises when the
+ * OS includes both a folder and its visible children as separate DataTransfer
+ * items.
+ *
+ * @param e - The React drag event
+ * @returns A Map of file paths to File objects
+ * @throws {Error} if the selection doesn't match one of the two accepted shapes
+ */
+export async function readFilesFromDragEvent(
+  e: React.DragEvent,
+): Promise<Map<string, File>> {
+  // Collect all entries and fallback files synchronously before any async work.
+  // The browser clears the DataTransfer after the event handler returns,
+  // so awaiting inside the loop would lose items beyond the first.
+  const entries: FileSystemEntry[] = [];
+  const fallbackFiles: File[] = [];
+
+  for (const item of e.dataTransfer.items) {
+    const entry = item.webkitGetAsEntry();
+
+    if (entry) {
+      entries.push(entry);
+    } else {
+      // Fallback for browsers that don't support webkitGetAsEntry
+      const file = item.getAsFile();
+
+      if (file) {
+        fallbackFiles.push(file);
+      }
+    }
+  }
+
+  const directoryEntries = entries.filter((e) => e.isDirectory);
+  const fileEntries = entries.filter((e) => e.isFile);
+
+  // Validate the selection shape
+  if (directoryEntries.length > 1) {
+    throw new Error(
+      "Please drag a single folder, or one or more individual files. " +
+        "Dragging multiple folders at once is not supported.",
+    );
+  }
+
+  if (
+    directoryEntries.length === 1 &&
+    (fileEntries.length > 0 || fallbackFiles.length > 0)
+  ) {
+    throw new Error(
+      "Please drag either a single folder or individual files, but not both at the same time.",
+    );
+  }
+
+  // Now process collected entries asynchronously
+  const files = new Map<string, File>();
+
+  if (directoryEntries.length === 1) {
+    // Single directory: read its full hierarchy
+    const directoryFiles = await readDirectoryEntry(
+      directoryEntries[0] as FileSystemDirectoryEntry,
+    );
+
+    for (const [path, file] of directoryFiles) {
+      files.set(path, file as File);
+    }
+  } else {
+    // Files only (no directories)
+    for (const entry of fileEntries) {
+      const file = await readFileEntry(entry as FileSystemFileEntry);
+      files.set(file.name, file as File);
+    }
+
+    for (const file of fallbackFiles) {
+      files.set(file.name, file);
+    }
+  }
+
+  return files;
+}
+
+/**
+ * Checks if a drag event contains files from the file system (as opposed to
+ * internal application data like dragging items within the workspace).
+ *
+ * @param e - The React drag event
+ * @returns true if the drag event contains files from the file system and not internal app data
+ */
+export function dragEventContainsFiles(e: React.DragEvent): boolean {
+  // Internal tree drags set application/json — exclude those even if Files is also present
+  if (e.dataTransfer.types.includes("application/json")) {
+    return false;
+  }
+
+  // Check if there are any files in the dataTransfer
+  if (e.dataTransfer.types.includes("Files")) {
+    return true;
+  }
+
+  // Also check items for file entries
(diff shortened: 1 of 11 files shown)


frontend/src/js/components/Uploads/dropZoneUtils.ts
```diff
@@ -0,0 +1,152 @@
...
+  FileSystemFileEntry,
+  FileSystemDirectoryEntry,
+} from "./FileApiHelpers";
+
+/**
+ * Custom MIME type used by internal tree drags to distinguish them from
+ * file-system drops. Using a custom type avoids false positives when an
+ * external application happens to set application/json on its drag data.
+ */
+export const INTERNAL_DRAG_MIME = "application/x-giant-tree-entry";
```
philmcmahon: can this be INTERNAL_DRAG_MIME_TYPE

frontend/src/js/components/Uploads/dropZoneUtils.ts
```diff
@@ -0,0 +1,152 @@
...
+
+/** If `key` already exists in `map`, append a numeric suffix to make it unique. */
+function deduplicateKey(map: Map<string, unknown>, key: string): string {
```
philmcmahon: getDeduplicatedKey

frontend/src/js/components/Uploads/dropZoneUtils.ts
```diff
@@ -0,0 +1,152 @@
...
+  const dotIndex = key.lastIndexOf(".");
+  const base = dotIndex > 0 ? key.slice(0, dotIndex) : key;
+  const ext = dotIndex > 0 ? key.slice(dotIndex) : "";
+
+  let counter = 1;
+  while (map.has(`${base} (${counter})${ext}`)) {
+    counter++;
+  }
+  return `${base} (${counter})${ext}`;
+}
```
philmcmahon: can we simplify this?

frontend/src/js/components/Uploads/dropZoneUtils.ts
```diff
@@ -0,0 +1,152 @@
...
+    );
+
+    for (const [path, file] of directoryFiles) {
+      files.set(path, file as File);
+    }
+  } else {
+    // Files only (no directories)
+    for (const entry of fileEntries) {
+      const file = await readFileEntry(entry as FileSystemFileEntry);
+      const name = deduplicateKey(files, file.name);
```
philmcmahon: not consistent with the existing behaviour where we appear to randomly drop duplicate files

frontend/src/js/components/Uploads/dropZoneUtils.ts
philmcmahon: maybe add path-browserify

frontend/src/js/components/Uploads/dropZoneUtils.ts
```diff
@@ -0,0 +1,152 @@
...
+/** If `key` already exists in `map`, append a numeric suffix to make it unique. */
+function deduplicateKey(map: Map<string, unknown>, key: string): string {
+  if (!map.has(key)) return key;
+
+  const dotIndex = key.lastIndexOf(".");
+  const base = dotIndex > 0 ? key.slice(0, dotIndex) : key;
+  const ext = dotIndex > 0 ? key.slice(dotIndex) : "";
+
+  let counter = 1;
+  while (map.has(`${base} (${counter})${ext}`)) {
```
philmcmahon: pull out `${base} (${counter})${ext}` into an inline funciton

frontend/src/js/components/Uploads/dropZoneUtils.ts
```diff
@@ -0,0 +1,152 @@
...
+    counter++;
+  }
+  return `${base} (${counter})${ext}`;
+}
+
+/**
+ * Reads files from a drag event - handles both direct file drops and directories.
+ * Works across all major browsers on all platforms using the File System Access API.
```
philmcmahon: just last bit relevant

frontend/src/js/components/Uploads/dropZoneUtils.ts
```diff
@@ -0,0 +1,152 @@
...
+  if (e.dataTransfer.types.includes("Files")) {
+    return true;
+  }
+
+  // Also check items for file entries
+  for (const item of e.dataTransfer.items) {
+    if (item.kind === "file") {
+      return true;
+    }
+  }
```
philmcmahon: this second check could be removed

frontend/src/js/components/Uploads/dropZoneUtils.ts
```diff
@@ -0,0 +1,152 @@
...
+}
+
+/**
+ * Checks if a drag event contains files from the file system (as opposed to
+ * internal application data like dragging items within the workspace).
+ *
+ * @param e - The React drag event
+ * @returns true if the drag event contains files from the file system and not internal app data
+ */
+export function dragEventContainsFiles(e: React.DragEvent): boolean {
```
philmcmahon: isFilesystemDragEvent

frontend/src/js/components/Uploads/dropZoneUtils.ts
hoyla: removing dedupe entirely

frontend/src/js/components/Uploads/dropZoneUtils.ts
hoyla: removing dedupe entirely

frontend/src/js/components/Uploads/dropZoneUtils.ts
hoyla: we now randomly drop duplicate files consistently 🤪

philmcmahon (approved):
drag

drop

⭐  such a nice feature

PR #551, opened 2026-02-19
add a new 'fast lane' option when uploading

twrichards:
## What does this change?
Adds a new 'fast lane' option when uploading (via new header `X-PFI-Fast-Lane`) which x10 the priority of the `TODO` relations in Neo4J which determine the order the workers pickup work - see existing priority logic https://github.com/guardian/giant/blob/2061212cdcfabec9e4e42a2f7a8184ee85e3502f/backend/app/services/manifest/Neo4jManifest.scala#L457

Here is the UI change (new checkbox near the Upload button, visible only to admins)...
image

Tested on playground, by uploading a couple of PDFs, one with fast lane ticked one without and we can see the priority on the respective TODO relations (by connecting to Neo4J and running a simple query in bolt on the respective blob URIs)...
| Normal | Fast Lane |
|--------|--------|
| PRIORITY: 200 image  | PRIORITY: 2000 image | 

We now have a way to skip the queue for very urgent uploads.

### DIFF
--- a/backend/app/commands/IngestFile.scala
+++ b/backend/app/commands/IngestFile.scala
@@ -24,7 +24,9 @@ object IngestFileResult {
 
 class IngestFile(collectionUri: Uri, ingestionUri: Uri, uploadId: String, workspace: Option[WorkspaceItemUploadContext],
                  username: String, temporaryFilePath: Path, originalPath: Path, lastModifiedTime: Option[String],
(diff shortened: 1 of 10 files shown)


srbd (approved):
👍 nice - let's revisit priority logic at a later date too as discussed

Issue #557, opened 2026-02-23
Clean up old workers


philmcmahon:
Giant currently doesn't clean up old workers in neo4j. As a result on PROD we have 46k workers 🤦🏻 . This might be having some performance impact, but it seems sensible to clean them up at some point

Issue #536, opened 2026-02-09
[Sub epic] Speeding up giant ingestion


philmcmahon:
Recently we have been adding a _lot_ of files to giant. There are some ideas below on how to speed up the ingestion process.

### ✅ (DONE) 1. Add a metric to show ingestion progress

This doesn't really exist at the moment and it's not that obvious even from the logs. I think it would mean shipping the result of this query `MATCH (b:Blob)<-[t:TODO]-(:Extractor) WHERE t.attempts = 0 return count(t)`to cloudwatch. This would help assess the impact of any other steps we take.

### 2. (DONE) ✅ Resolve concurrency problems with giant 

There is some prior art on this here https://github.com/guardian/giant/issues/359

Looking at [errors in kibana kibana for the last 12 hours](http://localhost:5601/goto/c0b82ab0-059f-11f1-bfa6-73ff2398523f) there are thousands of instances of `Failed to mark 'uri' processed by 'OcrMyPdfExtractor' as complete: Unexpected number of creates/deletes in markAsComplete. Created: 0. Deleted: 0

This is by far the most common error, and quite concerning as in the same period where we had 10k of this error, we had 40k instances of "working on" - suggesting that 25% of the work we do is going to waste.

### 3. ✅ Ingest batch execution failures

_I have a vague idea these only happen when there are lots of workers - anecdotally this appears to be true last night where dropping to 5 workers drastically reduce the number of these errors_ - see http://localhost:5601/goto/a79ac6f0-05a4-11f1-bfa6-73ff2398523f. -but more investigation needed

This error sometimes appears alongside some other interesting errors - see here http://localhost:5601/goto/929e92f0-05a3-11f1-bfa6-73ff2398523f

```
[...]

Issue #359, opened 2025-09-26
Concurrency issue with work allocation


hoyla:
It seems that when there could be occurrences when several workers which are triggered at the same time or only some millisecond apart, they might end up pick up the same task as each other.

this kibana query shows all three workers locking the same blob task on blob on4SvXDmv0xdd5C3CBsvBqgOdy2Dv35JppouasncS0BzV7LNH-NUUC24tSIqnCOmUbOFvrX_SZxH4bArIx2vsA

http://localhost:5601/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15d,to:now))&_a=(columns:!(message),filters:!(('$state':(store:appState),meta:(alias:!n,disabled:!f,index:d2720940-0e30-11eb-a809-5706b391daea,key:app,negate:!f,params:(query:pfi-worker),type:phrase),query:(match_phrase:(app:pfi-worker)))),index:d2720940-0e30-11eb-a809-5706b391daea,interval:auto,query:(language:kuery,query:%22on4SvXDmv0xdd5C3CBsvBqgOdy2Dv35JppouasncS0BzV7LNH-NUUC24tSIqnCOmUbOFvrX_SZxH4bArIx2vsA%22),sort:!(!('@timestamp',desc)))
running

match (b: Blob {uri: "on4SvXDmv0xdd5C3CBsvBqgOdy2Dv35JppouasncS0BzV7LNH-NUUC24tSIqnCOmUbOFvrX_SZxH4bArIx2vsA"}) return b
shows the blob is locked by all three instances. the workers are'’t fetching new tasks to complete

 

This doesn’t seem to be causing any issue, but it’s affecting the Giant performance and it’s worth thinking of how to fix it. After some discussion, we think possible ways to imrove this could be:

Using a messaging system such as kinesis stream

Moving the work allocation away from neo4j and into postgres

mbarton:
There’s https://github.com/softwaremill/elasticmq which could let you unify everything around SQS, running that for local dev/offline

hoyla:
Moving into In Progress because the work Phil's doing in #536 connects to this

PR #558, opened 2026-02-24
Refactor fetchWork to prevent multiple workers picking up the same task

philmcmahon:
## What does this change?
This PR aims to address https://github.com/guardian/giant/issues/359

There has been a long standing bug in giant where two workers end up picking up the same task. This PR aims to resolve that by:

 - Using a property rather than a relation to lock pieces of work
 - Adding an extra check after work has been checked to drop any work that has since been locked by another worker

## Using a property rather than a relation to lock pieces of work
The main benefit of this is that we move from a state where multiple workers can lock the same piece of work to a state where 'last write wins'

Arguably, this is actually worse, as with the current behaviour at least we have visibility when two workers have locked the same work:

`MATCH (n:Worker)-[:LOCKED_BY]-(b:Blob:Resource) return n`

Screenshot 2026-02-24 at 11 50 25

_However_, combined with the 'extra check' change, I think using a property makes sense, as it means we don't have to clean up any extra LOCKED_BY relations in the check query

## The extra check
With this PR, the behaviour is:
 1. Find the top priority work that the current worker can afford, returning a list of blob uris (first transaction)
 2. Take the list of blobUris and query neo4j for all of the blobs with those URIs that _also_ have the lockedBy property set to the current worker. Increment the number of attempts on the TODO for these blobs, fetch extra metadata and return  the final list of work (second transaction)

The intention is that, even if two workers attempt to lock the same work during transaction 1, one of them will drop any work that it doesn't have a lock on in transaction 2. The worst impact of this would be that a worker ends up with a much shorter or empty work list until it calls fetchWork again.

I've added an integration test which was failing on `main` on my local machine (the failure state is nondeterministic as it relies on timing, but I was seeing fairly consistent failures). It should always pass with this change

Changing the way locks work and work is being fetched is quite a big deal. I've tested on playground by verifying that files are still picked up by workers and ingested. I think on PROD we should see fairly quickly whether this resolves the concurrency issues - see https://github.com/guardian/giant/issues/359 and https://github.com/guardian/giant/issues/536 for possible logs to check

## Deployment concerns
Upon deployment, there may be existing work locked by workers. When they are replaced in the deployment the updated `releaseLocksForTerminatedWorkers` _won't_ remove the legacy LOCKED_BY relations, we'll need to do that in a manual job. If any of the work is already at the max number of attempts then it will never get completed. With that in mind, post deployment we should reset the `attempt` of any work with a LOCKED_BY relation

After this has been rolled out for a while, we can delete all Worker items from neo4j, thus resolving https://github.com/guardian/giant/issues/557

### DIFF
--- a/backend/app/services/manifest/Neo4jManifest.scala
+++ b/backend/app/services/manifest/Neo4jManifest.scala
@@ -569,7 +569,7 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
     } yield {
       if (claimedUris.length > work.length) {
         val diff = claimedUris.length - work.length
-        logger.info(s"Worker $workerName attempted to claim ${claimedUris.length} items but $diff were (probably) locked by another worker. Attempted URIs: [${claimedUris.mkString(", ")}] Actual work items fetched: [${work.map(_.blob.uri.value).mkString(", ")}]")
+        logger.info(s"Worker $workerName attempted to claim ${claimedUris.length} items but $diff were (probably) locked by another worker so only claimed ${work.length} items. Attempted URIs: [${claimedUris.mkString(", ")}] Actual work items fetched: [${work.map(_.blob.uri.value).mkString(", ")}]")
       }
       work
     }


backend/app/services/manifest/Neo4jManifest.scala
```diff
@@ -451,51 +451,37 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
...
+         |     end
+         |  )) as tasks
+         |
+         |UNWIND tasks[0] as task
+         |  WITH task.blob as blob, task.extractor as e, task.todo as todo
+         |
+         |  SET blob.lockedBy = {workerName}
+         |
+         |RETURN blob.uri as uri
+        """.stripMargin,
```
philmcmahon: This logic is unchanged from the previous behaviour, except for the `.lockedBy` bits

backend/app/services/manifest/Neo4jManifest.scala
```diff
@@ -42,6 +42,7 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
       tx.run("CREATE CONSTRAINT ON (resource: Resource)   ASSERT resource.uri   IS UNIQUE")
       tx.run("CREATE CONSTRAINT ON (extractor: Extractor) ASSERT extractor.name IS UNIQUE")
       tx.run("CREATE CONSTRAINT ON (tpe: MimeType)        ASSERT tpe.mimeType   IS UNIQUE")
+      tx.run("CREATE INDEX ON :Resource(lockedBy)")
```
twrichards: does `Blob` extend `Resource`?

philmcmahon:
This doesn't appear to have worked - workers are still picking up the same task, my assumption is because this is happening:

Worker 1: claimWork for blob1, no commit yet
Worker 2: claimWork for blob1, no commit yet
Worker 1: commit lockedBy for blob1
Worker 1: fetchClaimedWork for blob1
Worker 2: commit claimWork for blob1
<both workers now start working on blob1>

Essentially, exactly the same problem as we had before as splitting into two transactions doesn't stop two workers getting a lock on the same blob.

There might be further mitigations that occur to me that we could do here - e.g. doing a write to lockedBy in fetchClaimedWork so that it claims a write lock which might result in better serialisation, or something like this https://github.com/guardian/giant/tree/add-extra-check to do the 'check' as close as possible to the write. It would be quick to check if these changes have any impact right now as there are concurrency issues happening every few seconds, but they all seem kind of crap because the underlying concurrency problem remains. With that in mind, let's try this https://github.com/guardian/giant/compare/barton-hashing

PR #561, opened 2026-02-25
Implement 'barton hashing' to prevent multiple workers picking up the same task

philmcmahon:
## What does this change?
Yesterday we deployed https://github.com/guardian/giant/pull/558 in an attempt to stop multiple workers picking up the same task. It didn't solve the problem - my guess is for this reason https://github.com/guardian/giant/pull/558#issuecomment-3957965737

So, what's the next plan?
 - Implement a more convoluted locking approach that works?
 - Work out why, when we have thousands of TODOs to choose from, the same workers are repeatedly picking up the same tasks. 
 - Switch to using a queue for tasks (this will take too long, we need to get our backlog of tasks dealt with)

For addressing the second option, the obvious thing would be to impose a random sort order on TODOs before we start working on them. This might be possible to do on all TODOs without doing a full table scan - see https://stackoverflow.com/questions/12510696/neo4j-is-there-a-way-how-to-select-random-nodes (thanks @mbarton ) - or we could even do something like MATCH (tasks) LIMIT (num workers * max tasks per worker * N) and then randomise the result - where N is a number suitably large so that there's a low chance that the first tasks in the list picked by a worker are unlikely to be the same as another worker's selection.

I would be happy to try the above, but I prefer the approach of this PR, which avoids dealing with randomness.

This new approach is heavily inspired by a whatsapp chat with senor @mbarton, who writes:

```
or maybe try some gonzo consistent hashing.

- Get a sorted list of all the worker names on each worker.
- ⁠Find the index of the name of the current worker
- ⁠Filter out at the very start of fetching work todos where mod(todo.id) == worker_ix

I think everything in neo4j gets a sequential integer ID but if not you can add one (would have to be a manual query for existing todos).

The naive mod approach isn’t usually used because it’s rubbish if you are increasing/decreasing the set of workers constantly but you’re probably not for this and even then at worst workers pick up work already in progress which is already happening.

So it could be a decent get of jail hack and unlike the random ordering shouldn’t require a full scan because once you reach LIMIT it can stop and return
```

With all the above in mind, this PR:
 - Reverts https://github.com/guardian/giant/pull/558 - since it didn't really achieve anything
 - Implements the barton hashing approach

I've tested this locally and on playground. Fairly limited testing but it should be immediately obvious whether this is working on PROD. 

Once we've got through the backlog, the grafana graph showing outstanding work will be useful to verify that no work has somehow ended up abandoned. 

## Known issue with this approached
As Michael points out above, if we change the number of workers then you can end up with the same workers working on the same tasks. Just like the current situation! Since we don't change the number of workers very frequently that's fine - this is only a stop gap solution anyway till we switch to using a queue

### DIFF
--- a/backend/app/services/manifest/Neo4jManifest.scala
+++ b/backend/app/services/manifest/Neo4jManifest.scala
@@ -483,7 +483,7 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
         |  MATCH (worker :Worker { name: task.worker.name })
         |
         |  SET task.todo.attempts = task.todo.attempts + 1
-        |  MERGE (blob)-[:LOCKED_BY]->(worker)
+        |  MERGE (blob)-[:LOCKED_BY {lockedAt: {lockedAt}}]->(worker)
         |
         |RETURN
         |    blob,
@@ -502,7 +502,8 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
         "maxBatchSize", Int.box(maxBatchSize),
         "maxCost", Int.box(maxCost),
         "workerCount", Int.box(workerCount),
-        "workerIndex", Int.box(workerIndex)
+        "workerIndex", Int.box(workerIndex),
+        "lockedAt", Instant.now().toEpochMilli.asInstanceOf[java.lang.Long]
       )
     )
 


twrichards (approved):
very nice - I like it

philmcmahon:
> Something to bear in mind with this one is that it removes “work stealing”. So if there’s a worker that is still in the list but for whatever reason isn’t doing any tasks those tasks will remain unallocated until that worker is removed from the list.
> 
> As @philmcmahon said though you’ll see that in your Grafana graph of outstanding work

Here's a nice (though unfortunate) example of this in practice- search results for "no work found" log line

Screenshot 2026-02-26 at 14 35 53

Jobs remaining/in progress:
Screenshot 2026-02-26 at 14 34 16

Number of workers:
Screenshot 2026-02-26 at 14 34 31

So as we run out of work some workers presumably stop finding anything to do and are just idle whilst others are still chugging away

Issue #431, opened 2025-11-14
Allow users to leave a workspace that isn't public and isn't owned by that user


hoyla:
Applicable to any workspace in the 'Shared with me' list. 

Allows users to leave workspaces that, for example,
- are no longer of interest to them;
- contain content that keeps coming up, unwanted, in their general searches; 
- contain potential upsetting material;
- etc

This would be added to the semantic UI menu for an open workspace once implemented ( #585 )

PR #585, opened 2026-03-04
Workspace buttons discloser switched to a Semantic UI menu 

hoyla:
I hope to add some more options for workspaces and I think having a stack of buttons displayed vertically will start to look strange if we keep adding more, so I've changed the discloser that shows more buttons ...

Picture 243
<br/><br/>

... to a regular dropdown menu, which will more suitably handle additions:

Picture 245

Permissions/context-based UX remains the same: don't show an option if you can never do something (like Take Ownership for non-admins); display but disable if you could do something if you owned the workspace (like Share/Rename).

- Replace custom button/div dropdown with Semantic UI Dropdown component
- Switch disclosure icon from expand-more chevron to hamburger (bars) icon
- Add controlled open mode (isOpen/onClose) to ModalAction, ShareWorkspaceModal, and TakeOwnershipOfWorkspaceModal so modals can be triggered from menu items
- Menu items use Semantic UI Dropdown.Item with icons matching the tree context menu
- Take Ownership is hidden for non-admins; Share/Rename/Delete shown disabled for non-owners
- The controlled/uncontrolled open state pattern was duplicated across ShareWorkspaceModal, TakeOwnershipOfWorkspaceModal, and ModalAction so, in anticipation of further menu options that will introduce additional modals (I have some in mind - see e.g. #431 and #588 - we extracted it into a reusable useControlledOpen hook.

Tested locally and on playground with no noticeable ill effects

### DIFF
--- a/frontend/src/js/components/workspace/WorkspaceSummary.tsx
+++ b/frontend/src/js/components/workspace/WorkspaceSummary.tsx
@@ -210,7 +210,6 @@ export default function WorkspaceSummary({
         title={`Rename workspace '${workspace.name}'`}
         value={workspace.name}
         onConfirm={(newName) => renameWorkspace(workspace.id, newName)}
-        disabled={!isOwner}
         isOpen={renameModalOpen}
         onClose={() => setRenameModalOpen(false)}
       />
@@ -219,7 +218,6 @@ export default function WorkspaceSummary({
         actionDescription="Delete"
         title={`Delete workspace '${workspace.name}'?`}
         onConfirm={() => deleteWorkspace(workspace.id)}
-        disabled={!canDelete}
         isOpen={deleteModalOpen}
         onClose={() => setDeleteModalOpen(false)}
       />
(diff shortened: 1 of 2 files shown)


frontend/src/js/components/workspace/WorkspaceSummary.tsx
```diff
@@ -65,7 +62,13 @@ export default function WorkspaceSummary({
...
   clearFocus,
 }: Props) {
-  const [isShowingMoreOptions, setIsShowingMoreOptions] = useState(false);
+  const [shareModalOpen, setShareModalOpen] = useState(false);
+  const [takeOwnershipModalOpen, setTakeOwnershipModalOpen] = useState(false);
+  const [renameModalOpen, setRenameModalOpen] = useState(false);
+  const [deleteModalOpen, setDeleteModalOpen] = useState(false);
+
+  const isOwner = currentUser.username === workspace.owner.username;
+  const canDelete = isOwner || (isAdmin && workspace.isPublic);
```
zekehuntergreen: this is much more readable 👍

PR #610, opened 2026-03-13
[hold] Document view #5: Improve text overlay alignment for PageViewer selection accuracy

hoyla:
This is an attempt to address the way text selection in the PageViewer is often misaligned vertically. A little robot tells me that the browser was adding default padding vertically, affecting selection. This removes that. 

This doesn't fix the much more serious problem of text selection and highlighting being misaligned in terms of character counts down the page. But it does at least get the user onto the right line. 

Add `line-height: 1` and `white-space: pre` to **.pfi-page__pdf-text** to prevent browser defaults from distorting the invisible text overlay geometry. 
- The default `line-height` (~1.2) added extra vertical space above/below each text div, causing the selectable area to be taller than the actual glyph; and on pages with lots of lines eventually shifts the selection and highlight to below the text.
- `white-space: pre` prevents the browser from collapsing whitespace or wrapping text, keeping positions aligned with where PDF.js placed them. Allegedly.

### Before:
Picture 65

To select "workspaces section" the user actually has to drag the cursor across what appears to be the line below to get the selection right. The highlight itself is also too tall, although this actually helps since it overlaps the text, unlike the required cursor action. 

### After:
Picture 66

Cursor drag (and the highlight itself) aligns more accurately to the displayed text.

### Testing:      
- [x] Looks good on local dev but..
- [ ] Probably needs some quite thorough testing with different documents on playground

### DIFF
--- a/frontend/src/js/components/PageViewer/PageHighlight.tsx
+++ b/frontend/src/js/components/PageViewer/PageHighlight.tsx
@@ -18,7 +18,7 @@ export const PageHighlight: FC<PageHighlightProps> = ({
   const isFind = type === "FindHighlight";
   return (
     <>
-      {highlight.data.map((span) => {
+      {highlight.data.map((span, i) => {
         const style: CSSProperties = {
           position: "absolute",
           left: span.x * scale,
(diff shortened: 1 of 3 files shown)


hoyla:
I'm going to close this. It only addresses a tiny aspect of the wider problem and isn't worth it

PR #611, opened 2026-03-13
Add agents.md

philmcmahon:
## What does this change?
A starting point for improving the quality of AI generated code on this repository. Note that claude.md gets added to the context of every claude session so 'less is more' - maybe this is already too verbose. 

Not tested, but I guess we'll find out soon enough if it is helpful.

This is not an endorsement of [illegal wars](https://www.theguardian.com/technology/2026/mar/13/anthropic-pentagon-artificial-intelligence)


hoyla:
😳

PR #631, opened 2026-03-20
Document view #4: Use real page dimensions for PageViewer layout and width-fitted rendering

### LINKED ISSUE #396: When opening a 'Unified Page Viewer' doc, zoom to fit active page to width
We have some funny shaped documents. But most are portrait pages of text. For readability, having the text large enough to read is more important than showing the full depth. So the default view should be to zoom to fit the width of the active page to the available space onscreen.

c.f. other zoom related features:
#344 
#325 
#345 
#349

hoyla:
Closes #396 
Requires #607 - don't submit before rebasing after that

Improve PageViewer rendering by using actual page dimensions from Elasticsearch for layout calculations, and always fitting pages to container width.

## Why

Previously, the virtual scroll assumed a square (1:1) aspect ratio for page slot heights, and `renderPagePreview` fitted pages to whichever dimension was longest. This caused layout jank — visible pages would jump as real content loaded with a different aspect ratio, and landscape documents were rendered smaller than necessary.

### Backend
- **PagesController**: Extend the `getPageCount` endpoint to also return the dimensions of page 1 alongside the count. The dimensions query recovers gracefully if the data is unavailable (e.g. older documents).
- **Pages2**: Add `getFirstPageDimensions` — a lightweight source-filtered ES query with a dedicated `PageDimensionsHitReader`.

### Frontend
- **VirtualScroll**: Use the real page aspect ratio (from the backend) to calculate slot heights, falling back to A4 portrait (297/210). Correctly invert the aspect ratio when pages are rotated 90°/270°.
- **PdfHelpers**: Simplify `renderPagePreview` to always scale to container width, giving a more natural reading experience for portrait documents. 
- **PageViewerOrFallback / PageViewer**: Thread `firstPageDimensions` from the API response down to `VirtualScroll`.
- **layout.ts**: Extract `pageSlotHeight` into a standalone utility for testability.

## Integration /unittests

- Backend integration tests for `Pages2.getFirstPageDimensions` (happy path + empty case).
- Frontend unit tests for `pageSlotHeight` covering A4 fallback, custom dimensions, rotation (90°/180°/270°), and landscape pages.

## Manual tests

- [x] Manual testing across portrait/landscape documents and rotated views.
- [ ] Ditto, in playground

### DIFF
--- a/frontend/src/js/components/PageViewer/VirtualScroll.spec.ts
+++ b/frontend/src/js/components/PageViewer/VirtualScroll.spec.ts
@@ -1,4 +1,4 @@
-import { pageSlotHeight } from "./layout";
+import { pageSlotHeight, pageTransform } from "./layout";
 
 describe("pageSlotHeight", () => {
   const MARGIN = 10;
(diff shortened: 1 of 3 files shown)


philmcmahon (approved):
Nice, happy for this to go out once you've had a chance to test on playground

hoyla:
Phil spotted an ussue with page container slot size when pages are rotated; hopefully fixed in https://github.com/guardian/giant/pull/631/changes/158a3e8f261a9b1974fde699471e9a7ac2b10a20 

Will test in playground with some interestly shaped docs when playground is available

PR #654, opened 2026-04-01
Add support for second spot asg in worker control

philmcmahon:
## What does this change?

We'd like to introduce a spot instance pool to Giant when running in AWS. This will reduce costs when performing large ingestions. This PR updates the worker control in giant to support having a second worker autoscaling group.

The general idea is:
 - If we need more workers and the spot ASG is not full, we scale up the spot ASG
 - If we are trying to scale up the spot ASG but we notice that actual instances < desiredInstances then we scale up the on demand ASG instead, assuming a problem with spot capacity
 -  Giant will scale DOWN both asgs if it runs out of work, starting with the on demand ASG

The associated infrastructure changes are here: https://github.com/guardian/investigations-platform/pull/620

 - [ ] Tested locally
 - [x] Tested on playground

### DIFF
--- a/backend/app/services/Config.scala
+++ b/backend/app/services/Config.scala
@@ -186,12 +186,13 @@ case class BucketConfig(
 }
 
 case class AWSDiscoveryConfig(
-  region: String,
-  stack: String,
-  app: String,
-  stage: String,
-  runningLocally: Option[Boolean],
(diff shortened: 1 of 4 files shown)


twrichards (approved):
looks awesome 👍  - noticed there's a merge conflict though

Issue #620, opened 2026-03-16
Doc views: previous and next doc buttons are almost invisible


hoyla:
- They're almost impossible to spot
- Previous and next should not be far apart; people move back and forth
- They don't exist on PageViewer views

PR #658, opened 2026-04-03
Document viewer nav #1: Karpow that workspace folder

### LINKED ISSUE #470: Document viewer: revisit Previous and Next buttons
Weirdly lots of people have suddenly been asking about this: they'd like a way to cycle through documents within a folder of a workspace without going back to the workspace, i.e. "previous and next file in the same location".

Relatedly, when we perform a search in Giant, if a user double clicks on a search result AND that search result is for a file type (e.g. jpeg, wordfile, video, audio, email) that isn't rendered in our combined document viewer, then the view has a bar at the bottom showing very little arrows to go to the previous or next search result from the list in the earlier search. 

In the combined document viewer no such feature exists; however there are up and down arrows to cycle through search hits within the current document (the top layer being for hits of the original search time and the bottom layer being for an additional "find x in document" unrelated to the original search). 

Ideally we'd amalgamate all these features into something a bit more coherent - 

PREVIOUS/NEXT DOC
Take account of the context from which a document is opened, and once opened, in every single way we render documents,
- If a user opens the document from a search, previous and next buttons should cycle through the documents that matches that search. 
- If a user opens the document from a dataset or workspace, previous and next buttons should cycle through the documents at the same file hierarchy level. 
- If a user opens a document directly from a URL, disable the previous and next buttons. (but c.f. https://github.com/guardian/giant/issues/393)
 
Plus more granular search-related navigation:

SIMPLE FIND WITHIN DOC
In the combined document viewer, and in the text, timecodes, OCR, translated views of other document viewers:
- An initially empty text entry field with previous and next buttons to "find in file". (This feature is already present in the [combined document viewer ](https://github.com/user-attachments/assets/51722cb7-2cd2-4366-8b39-4adaf6d9bc10)and is also provided by our older PDF viewer see [comment below](https://private-user-images.githubusercontent.com/773940/528596348-e47b44f8-47e9-488a-80cf-ae19a059405d.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjYxNDYzMzYsIm5iZiI6MTc2NjE0NjAzNiwicGF0aCI6Ii83NzM5NDAvNTI4NTk2MzQ4LWU0N2I0NGY4LTQ3ZTktNDg4YS04MGNmLWFlMTlhMDU5NDA1ZC5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjUxMjE5JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI1MTIxOVQxMjA3MTZaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT1hZGI2M2M3NWZhYjkzOGY2NTg0Y2FmYTFlMTY4ZDVlYjBmNWUwOWU0NzM4ODYyNDY1MzlmZDdlYjUxMjUyZmRmJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.g7Ux7gBEh20jUbLEcrfq3YaGbtZ7L65Mpt4LJmZVzVI))

WHEN A DOCUMENT IS OPENED FROM SEARCH INTO THE COMBINED DOCUMENT VIEWER
Show an additional find/

### LINKED ISSUE #670: Check after 'karpow that workspace folder': Add hasPrevious/hasNext to WorkspaceNavigation type
\`DocumentFooter\` references \`workspaceNav.hasPrevious\` and \`workspaceNav.hasNext\`, and \`Viewer.tsx\` constructs a fallback object with those fields, but the \`WorkspaceNavigation\` type only defines \`goToPrevious\` and \`goToNext\`. This causes a TypeScript compilation error.

**Recommendation:** Either:
1. Add \`hasPrevious: boolean\` and \`hasNext: boolean\` to the \`WorkspaceNavigation\` type and compute them in \`computeWorkspaceNavigation\`, or
2. Derive them in consumers as \`goToPrevious !== undefined\` and remove the explicit fields from \`DocumentFooter\` and \`Viewer\`.

This fix belongs on \`ljh-workspace-nav\` since that's where the type is defined.

**Files involved:**
- \`frontend/src/js/util/workspaceNavigation.ts\`
- \`frontend/src/js/components/viewer/DocumentFooter.tsx\`
- \`frontend/src/js/components/viewer/Viewer.tsx\`

hoyla:
> [!NOTE]
> This PR is ~700 lines but a big chunk of that is a massive test file. Will cut the tests back if you say it's overkill.

## Previous/next document navigation from workspaces

Partially addresses #470 and #620 , both of which will disappear entirely in the subsequent stages.

Previously, the prev/next navigation only worked when a document was opened from a search result. Users working through documents in a workspace had to go back to the workspace tab, click the next file, and wait for a new viewer tab to open. This is particularly tedious when reviewing many files in sequence.

With this change, when a user opens a document from a workspace, they can use previous/next buttons (and `shift+left`/`shift+right` keyboard shortcuts) to cycle through sibling documents in the same folder — matching the existing behaviour for cycling through search results.

### How it works

1. **On double-click in workspace:** Before opening the new tab, the workspace page writes the sibling leaf URIs (files in the same folder, in the user's current sort order) to `sessionStorage`, keyed by a unique nonce.
2. **In the viewer tab:** A `useWorkspaceNavigation` hook reads the sibling URIs from `sessionStorage` on mount and provides `goToPrevious`/`goToNext` functions. The nonce is passed via `?navId=` query param.
3. **Fallback logic:** When a document is opened from search, search result navigation takes priority. When opened from a workspace, workspace navigation is used instead. The existing StatusBar buttons and keyboard shortcuts handle both — no separate UI is added.
4. The keyboard shortcuts associated with the prev/next icons when used from search (shift left, shift right) also work in this mode. This is the real win. To be able to navigate through a folder using the keyboard while checking each doc is a significant usability gain. 

This approach avoids any additional API calls — the workspace page already has the tree in memory, and `sessionStorage` is scoped to the originating tab so there's no leakage to shared links. Old entries are pruned by count (max 20) rather than consumed on read, so refreshing the page doesn't break navigation.

### Design changes

New previous and next buttons have been added to the "combined" (pageviewer) mode:

- Both buttons are now adjacent on the right side of the toolbar (instead of at opposite ends)
- Icons changed from subtle chevrons to filled play-arrow triangles for better visibility
- Hover state and clearer active/inactive styling

### What this doesn't do

- Doesn't touch `StatusBar.js` or add a _new_ footer component — that's covered in the next stage (branch `ljh-document-footer`). As a consequence the nav in non-pageviewer view modes is still the old style.
- Doesn't mess with the document prev and next function for search results, so that remains unavailable in pageview documents. (It's addressed in `ljh-document-footer`.)
- No change to the Pageviewer's control widget that provides find-in-document, zoom, and rotate features. Those will be consolidated into a the page footer in the third branch (`ljh-combined-view`).
- Navigation is scoped to siblings in the same folder, not the entire workspace tree. I didn't want to take on trees with a million leaves.

### Dependency graph
    main
     └── ljh-workspace-nav        ← this PR
          └── ljh-document-footer  (DocumentFooter refactor + search stepper)
               └── ljh-combined-view (find-in-doc, zoom, rotate for paged view)

## Integration tests
200-odd lines of tests covering `leafUrisOfChildren`, `findNodeById`, `computeWorkspaceNavigation`, and `storeWorkspaceSiblingUris`. All 133 existing tests continue to pass.

## Manual tests
 - [x] Tested locally
 - [x] Tested on playground (build 6433 [ljh-workspace-nav] )

### DIFF
--- a/backend/app/utils/WorkerControl.scala
+++ b/backend/app/utils/WorkerControl.scala
@@ -107,7 +107,7 @@ class AWSWorkerControl(config: WorkerConfig, discoveryConfig: AWSDiscoveryConfig
 
   private def scaleUpOrDownIfNeeded(state: AWSWorkerControl.State, workerAutoScalingGroupName: String)(implicit ec: ExecutionContext): Unit = {
       val operation = AWSWorkerControl.decideOperation(state, Instant.now().toEpochMilli, config.controlCooldown.toMillis)
-      
+
       logger.info(s"AWSWorkerControl desiredNumberOfWorkers: ${state.desiredNumberOfWorkers}, inProgress: ${state.inProgress}, outstandingFromIngestStore: ${state.outstandingFromIngestStore}, outstandingFromTodos: ${state.outstandingFromTodos} lastEventTime: ${new Date(state.lastEventTime)}, minimumNumberOfWorkers: ${state.minimumNumberOfWorkers}, maximumNumberOfWorkers: ${state.maximumNumberOfWorkers}, operation: $operation")
 
       val scaleResult = operation match {
--- a/frontend/src/js/components/PageViewerOrFallback.tsx
+++ b/frontend/src/js/components/PageViewerOrFallback.tsx
@@ -105,12 +105,14 @@ export const PageViewerOrFallback: FC<{}> = () => {
   const navId = searchParams.get("navId");
   const navIndexParam = searchParams.get("navIndex");
   const navIndex = navIndexParam !== null ? parseInt(navIndexParam, 10) : null;
-  const workspaceNav = useWorkspaceNavigation(
-    uri,
-    navId,
-    Number.isFinite(navIndex) ? navIndex : null,
-    history.push,
-  );
+  const workspaceNav = navId
+    ? useWorkspaceNavigation(
+        uri,
+        navId,
+        Number.isFinite(navIndex) ? navIndex : null,
+        history.push,
+      )
+    : undefined;
 
   const [response, setResponse] = useState<PageCountResponse | null>(null);
   const view = useSelector<GiantState, string | undefined>(
@@ -174,18 +176,6 @@ export const PageViewerOrFallback: FC<{}> = () => {
         </div>
         {resource && (
           <div className="document__status">
-            {workspaceNav.goToNext && (
-              <KeyboardShortcut
-                shortcut={keyboardShortcuts.nextResult}
-                func={workspaceNav.goToNext}
-              />
-            )}
-            {workspaceNav.goToPrevious && (
-              <KeyboardShortcut
-                shortcut={keyboardShortcuts.previousResult}
-                func={workspaceNav.goToPrevious}
-              />
-            )}
             {/* Left spacer: document__status uses space-between to match the legacy StatusBar two-span layout */}
             <span />
             <span className="doc-nav-buttons">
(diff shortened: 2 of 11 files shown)


philmcmahon (approved):
I have reviewed this by checking it out on my machine and changing things I thought could be done better. Fingers crossed I've not broken everything with those changes, but I would be happy for this to go out now.

We discussed elsewhere the idea that using local storage is not ideal for this feature and it would be nicer to maybe store the workspace context in the url and then fetch the workspace on page load to work out next/previous. However, this would not be totally simple as there is no guarantee that the user a file is shared with will have access to the same workspace it was shared from. With that in mind we agreed to ship this as is

hoyla:
> I have reviewed this by checking it out on my machine and changing things I thought could be done better. Fingers crossed I've not broken everything with those changes, but I would be happy for this to go out now.
> 
> We discussed elsewhere the idea that using local storage is not ideal for this feature and it would be nicer to maybe store the workspace context in the url and then fetch the workspace on page load to work out next/previous. However, this would not be totally simple as there is no guarantee that the user a file is shared with will have access to the same workspace it was shared from. With that in mind we agreed to ship this as is

Created issue #671

PR #677, opened 2026-04-10
Fix double-click required to expand folders in Datasets view

hoyla:
The LazyTreeBrowser adds leaves to expandedEntries in anticipation of them becoming nodes once the API response returns. However, TreeBrowser was filtering expandedEntries through isTreeNode before passing them down as expandedNodes, which discarded the pre-emptively expanded leaves. This meant the node appeared collapsed after loading, requiring a second click to expand.

Fix by passing expandedEntries directly to Node without filtering, and widening the expandedNodes prop type from TreeNode<T>[] to TreeEntry<T>[]. Node.isExpanded() already matches by ID so the element type is irrelevant.

 - [x] Tested locally
 - [ ] Tested on playground


philmcmahon (approved):
I don't really understand the comment that was here before, and the PR description doesn't help that much, but I can see that this fixes the bug! Nice one :shipit:

hoyla:
Me neither!

PR #685, opened 2026-04-14
fix the cleanup of scratch/temp files (during email parsing)

twrichards:
We've recently observed disk space errors when ingesting thousands of emails over the course of a few days, looking at disk usage pattern for an example worker...
image
... it was clear that something wasn't being cleaned-up.

This PR fixes that.


philmcmahon (approved):
Nice!
stewie emails

PR #686, opened 2026-04-14
Add extra logging to WorkerControl

philmcmahon:
## What does this change?
In WorkerControl I am wondering if we are being _too_ blasé about the call to describe the ASG and are ending up swallowing any errors that come back. This PR attempts to get a bit more visibility if that call throws an error. Tested on PROD whilst debugging an issue.

 - [ ] Tested locally
 - [ ] Tested on playground

## Old PR context
WorkerControl is not, err, working, on PROD following the merge of https://github.com/guardian/giant/pull/654. 

I am fairly confident this is because giant is failing to fetch the state of the worker/spot worker asgs here https://github.com/guardian/giant/blob/f058c8cade982aff37c851b3a204ac52b085e3bd/backend/app/utils/WorkerControl.scala#L75 - we know this because neither publishStateMetrics nor scaleUpOrDownIfNeeded are happening

### DIFF
--- a/backend/app/utils/WorkerControl.scala
+++ b/backend/app/utils/WorkerControl.scala
@@ -72,7 +72,11 @@ class AWSWorkerControl(config: WorkerConfig, discoveryConfig: AWSDiscoveryConfig
         scheduler.scheduleWithFixedDelay(config.controlInterval, config.controlInterval)(() => {
           // Only run the check on the oldest instance to get as close as we running the checks as a "singleton"
           if(runningOnOldestInstance()) {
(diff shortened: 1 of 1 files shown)


twrichards (approved):
IIRC you deployed this manually and have since fixed the underlying problem but approving since the extra logging is no bad thing to keep permanently

philmcmahon:
> IIRC you deployed this manually and have since fixed the underlying problem but approving since the extra logging is no bad thing to keep permanently

ah yes the description is a bit out of date, I have updated it to make it clearer what the intention is

PR #672, opened 2026-04-08
Document viewer nav #2 PR: shared footer component for all document types

hoyla:
> [!NOTE]
> Based on `ljh-workspace-nav` (#658). 
> 
> Second of three big PRs to consolidate the document view navigation UI. 
> 
> The final biggie is for branch [ljh-combined-nav](https://github.com/guardian/giant/tree/ljh-combined-view), which builds on this PR.  
> 
> There are some further issues after that (see https://github.com/guardian/giant/issues/668 , https://github.com/guardian/giant/issues/669 , https://github.com/guardian/giant/issues/671 . These are refactor issues rather than features. 

## Shared DocumentFooter and search highlight stepper

Replaces the legacy `StatusBar` in `Viewer` and the inline nav buttons in `PageViewerOrFallback` with a single shared `DocumentFooter` component. Also adds a `SearchStepper` for stepping through search highlights in text views of paged documents, which was previously only available in the non-paged viewer.

## Why

The previous/next navigation for search results lived in the class-based `Viewer` component, with ~60 lines of imperative state management (`setupSearchContext`, `previousResult`, `nextResult`, `hasPreviousResult`, `hasNextResult`). Meanwhile, `PageViewerOrFallback` had its own inline nav buttons for workspace navigation but no search result navigation at all. The two code paths were diverging and becoming harder to maintain.

## Changes

- **`useSearchHighlightStepper`** — hook that extracts the highlight stepping logic (current index, total count, next/previous with wrapping) from the class component into a reusable form, reading from the existing Redux highlight state.
- **`SearchStepper`** — small presentational component showing the current query, match count, and prev/next buttons. Uses CSS modules. Parses the chip-based query format for display.
- **`DocumentFooter`** — shared footer used by both `Viewer` and `PageViewerOrFallback`. Handles the preview switcher, search result document navigation, and workspace sibling navigation. Search results take priority over workspace nav when both are available.
- **`Viewer`** — loses ~140 lines of search result state and navigation methods. Now delegates entirely to `DocumentFooter`.
- **`PageViewerOrFallback`** — gains `SearchStepperOverlay` (extracted to avoid duplication) for keyboard-shortcut-driven highlight stepping in text views, plus scroll-to-highlight behaviour. Uses `DocumentFooter` instead of inline nav buttons.
- **SCSS** — inline styles moved to proper classes (`document__viewer-fallback`, `document--text-content`, additions to `document__page-viewer-wrapper` and `document__status`).

## What this doesn't do

[...]

Issue #668, opened 2026-04-08
Document viewer nav #4: Unify Controls and usePageFind to eliminate duplicated find/step logic


hoyla:
Something to revisit in the document viewer work

\`Controls.tsx\` and \`usePageFind.ts\` both independently manage highlight state (focused index, highlights array, pending state) and implement the same fetch-then-step-through pattern. \`Controls\` always hits \`/search\`, \`usePageFind\` always hits \`/find\`.

**Recommendation:** Consolidate into a single shared hook parameterised by endpoint, removing the duplication. This is closely related to #669  (search vs find endpoint clarification).

**Files involved:**
- \`frontend/src/js/components/PageViewer/Controls.tsx\`
- \`frontend/src/js/components/PageViewer/usePageFind.ts\`

PR #689, opened 2026-04-19
Document viewer nav #3: one footer to rule them all 

### LINKED ISSUE #470: Document viewer: revisit Previous and Next buttons

### LINKED ISSUE #661: Document view nav #3: get rid of the pageviewer control widget
Once we have a shared document footer (see #660 ), we can move the pageviewer UI's zoom, rotate, find-in-file etc features from the overlapping (and comments destroying) control widget into that document footer. 

For simplicity we will keep the step-through-results feature from the control widget out of the footer. But make it much smaller and stop it looking like an editable search field, because it isn't.

hoyla:
> **Note**
> Based on `ljh-document-footer` ([#672](https://github.com/guardian/giant/pull/672)) and can't be merged without it.
>
> Third of three big PRs to consolidate the document view navigation UI.

## Move find/zoom/rotate from control widget to footer

**Remove find-in-document, zoom, and rotate controls from the floating `Controls` widget that overlaid the PageViewer document view, and into the `DocumentFooter` that (as of #658 and #672) shows at the bottom of all document views.**

This PR completes the work on consolidating different view modes' navigation features. There's more to do to in this area, but those changes have a different (usually much smaller) scope to these three. 

## Motivation for this change

1. To reorganise the logic around document navigation into more purpose-built functions.
2. It makes the navigation controls for different view modes the same.
3. In the "Combined" document viewer the `Controls` component was a floating overlay on the view that managed its own find-in-document state, zoom level, rotation, and search highlight stepping — all inline. It overlapped document content and comments, and its search input looked editable when it was actually just displaying the URL query. Moving these controls into the footer gives them a permanent, predictable home alongside the preview switcher and document-to-document navigation.

## Changes

- **`usePageFind`** — new hook that encapsulates find-in-document logic: debounced API calls to `/api/pages2/{uri}/find`, highlight tracking, and next/previous stepping. Distinct from the URL query search highlights. Eventually we should stop search from specifying a particular view; the rationale there is incomprehensible to users.
- **`model.ts`** — `HighlightsState` type and `getPreloadPages` function relocated to `model.ts` from `PageViewer.tsx` to avoid circular dependencies.
- **`PageViewerOrFallback`** — lifts rotation, scale, and find state up from `PageViewer` so `DocumentFooter` can render the controls. `PageViewerContent` element extracted into a variable to deduplicate the two rendering branches.
- **`DocumentFooter`** — gains `FindInput`, zoom (in/out), and rotate (clockwise/anti-clockwise) icon buttons when showing the combined view. These are only rendered for paged documents in combined mode since rotate is unnnecessary in text views; and these features are an integral part of the PDF Viewer component.
- **`PageViewer`** — simplified to receive `findQuery`, `findHighlightsState`, `focusedFindHighlight`, `rotation`, and `scale` as props instead of managing them internally. Retains only URL-query search highlight state.
- **`Controls`** — reduced to a thin wrapper around `SearchStepper` for the fixed-query search highlight case. Zoom, rotate, and Cmd+F concerns removed. The search stepper may in time be moved to the footer, subject to user feedback.
- **`FindInput`** — converted to `forwardRef`; styled with consistent chevron nav icons, hover states, and fixed-width layout.
- **SCSS** — new `document__footer-find` and `document__footer-view-controls` classes; inline styles moved to proper classes.

## What this doesn't do

- Doesn't unify the duplicated find/step logic between the `Controls` search stepper and `usePageFind` ([#668](https://github.com/guardian/giant/issues/668)).
- Doesn't address the view-mode selection bug where stepping through search results lands on Text view instead of Combined view for paged documents. However we have enough logic now to decide an appropriate view here, rather than in search results. 

### Dependency graph

```
main
 └── ljh-workspace-nav        (#658)
      └── ljh-document-footer  (#672)
           └── ljh-combined-view  ← this PR
```

## Manual tests

- [x] Tested locally
- [ ] Tested on playground


philmcmahon:
What's this precious? A lord of the rings reference?! 😍  maybe we should review it precious

hoyla:
> What's this precious? A lord of the rings reference?! 😍 maybe we should review it precious

My ruse appears to be working

philmcmahon (approved):
This seems to be largely moving stuff around and the nwe logic is encasulated in one or two files so I am happy for this to go out. One extra bit of testing that could be worth doing is trying it out on playground with the network tab open just to check that there's nothing bad happening e.g. hundreds of requests firing when you jump through matches - think we'd expect a few requsts to download each page around each match but worth checking there are no infinite loops or anything 

gandlaf

hoyla:
Looks OK. There are some repeated requests (e.g. if I go forward and then back to the same document it makes the same requests rather than using a cache) but low numbers and very low impact.

PR #691, opened 2026-04-20
add new (admin only) option to download all the text of a workspace

### LINKED ISSUE #693: Export workspace text as NotebookLM-friendly files
NB this directly pertains to an active journo task

twrichards:
Sometimes it's useful to be able to extract all the text for a workspace to use in other tools (notable example being NotebookLM), fortunately we have all the text (original text plus OCRed text) available in ElasticSearch (originally/primarily for the purposes of search/highlighting).

## What does this change?
Adds a new POST endpoint `/api/workspaces/:workspaceId/text` which if passed an array of blobUris will returns the text (for each language, typically `english`) preferring the OCR text if available. If passed an empty list then it returns a word count for the entire workspace.

Added a new menu option for the workspace, which launches a new modal, fetching/displaying the total word count and then presents a number field (min 1, max number of unique blobUris in the workspace) of approximately how many files to distribute the text across (approximate number of files, since the different files will have different word counts and so won't fit perfectly evenly between the files, so might end up with a few more or a few less than the target).

The text content of each original file will be wrapped as follows...
```----- START OF ${file} -----\n\n${text}\n\n----- END OF ${file} -----\n\n\n```
... where `file` is `${blobUriToWorkspacePath[blobUri]} (${lang})`.

Once the user clicks Start Download they're presented with a progress bar, and then a series of files will start downloading named after the workspace (and workspace ID) followed by the file number with the extension `.txt`.

We distribute the text across a configurable number of files, so they're easy to upload to tools like NotebookLM (which has limits on the number of files [based on licence tier] as well as hard limit of 500k words per file - and anecdotally we've found it performs better with smaller files, i.e. dividing the total words into as many files as possible under the file limits).

NOTE: this feature is implemented such that most of the hard work is done in the browser, to keep load on the server to a minimum.

 - [x] Tested locally
 - [x] Tested on playground

## Things to think about

## Images
image
image
image
image

### DIFF
--- a/backend/app/controllers/api/Workspaces.scala
+++ b/backend/app/controllers/api/Workspaces.scala
@@ -160,14 +160,19 @@ class Workspaces(
       .map(workspace => Ok(Json.toJson(workspace)))
   }
 
+  def getTotalWordCount(workspaceId: String) = ApiAction.attempt { req =>
+    for {
+      _ <- annotation.getWorkspaceMetadata(req.user.username, workspaceId) // check workspace exists and user has access
+      totalWordCount <- index.getTotalWordCountForWorkspace(workspaceId)
+    } yield Ok(Json.toJson(totalWordCount))
+  }
+
   def getText(workspaceId: String) = ApiAction.attempt(parse.json) { req =>
     for {
       _ <- annotation.getWorkspaceMetadata(req.user.username, workspaceId) // check workspace exists and user has access
       blobUris <- req.body.validate[List[String]].toAttempt
-      result <-
-        if (blobUris.isEmpty) index.getTotalWordCountForWorkspace(workspaceId).map(Json.toJson(_))
-        else index.getTextForBlobs(blobUris).map(Json.toJson(_))
-    } yield Ok(result)
+      result <- index.getTextForBlobs(workspaceId, blobUris)
+    } yield Ok(Json.toJson(result))
   }
 
   def updateWorkspaceFollowers(workspaceId: String) = ApiAction.attempt(parse.json) { req =>
--- a/backend/app/controllers/api/Workspaces.scala
+++ b/backend/app/controllers/api/Workspaces.scala
@@ -160,6 +160,16 @@ class Workspaces(
       .map(workspace => Ok(Json.toJson(workspace)))
   }
 
+  def getText(workspaceId: String) = ApiAction.attempt(parse.json) { req =>
+    for {
+      _ <- annotation.getWorkspaceMetadata(req.user.username, workspaceId) // check workspace exists and user has access
+      blobUris <- req.body.validate[List[String]].toAttempt
+      result <-
+        if (blobUris.isEmpty) index.getTotalWordCountForWorkspace(workspaceId).map(Json.toJson(_))
+        else index.getTextForBlobs(blobUris).map(Json.toJson(_))
+    } yield Ok(result)
+  }
+
   def updateWorkspaceFollowers(workspaceId: String) = ApiAction.attempt(parse.json) { req =>
     for {
       data <- req.body.validate[UpdateWorkspaceFollowers].toAttempt
--- a/backend/app/services/index/ElasticsearchResources.scala
+++ b/backend/app/services/index/ElasticsearchResources.scala
@@ -513,21 +513,28 @@ class ElasticsearchResources(override val client: ElasticClient, indexName: Stri
     }
   }
 
+  private def buildWorkspaceIdFilter(workspaceId: String) = {
+    nestedQuery(IndexFields.workspacesField,
+      termQuery(s"${IndexFields.workspacesField}.${IndexFields.workspaces.workspaceId}", workspaceId)
+    )
+  }
+
   override def getTotalWordCountForWorkspace(workspaceId: String): Attempt[Long] = {
     val query = search(indexName)
       .query(
-        nestedQuery(IndexFields.workspacesField,
-          termQuery(s"${IndexFields.workspacesField}.${IndexFields.workspaces.workspaceId}", workspaceId)
-        )
+        buildWorkspaceIdFilter(workspaceId)
       )
       .aggregations(
         scriptedMetricAggregation("workspaceWordCount")
           .initScript(Script("state.total = 0L").lang("painless"))
           .mapScript(Script("""
-                              |if (params._source != null && params._source.text != null) {
-                              |  for (entry in params._source.text.entrySet()) {
-                              |    if (entry.getValue() != null && entry.getValue().length() > 0) {
-                              |      state.total += /\s+/.split(entry.getValue()).length;
+                              |if (params._source != null) {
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: frontend/src/js/components/workspace/WorkspaceSummary.tsx [hunk 1]
```diff
@@ -166,6 +167,11 @@ export default function WorkspaceSummary({
               onClick={() => setTakeOwnershipModalOpen(true)}
             />
           )}
+          <Dropdown.Item
+            icon="file alternate outline"
+            text="Donwload all text"
```
hoyla: "Donwload">"Download".

Also can we put this at the bottom of the menu since

- It's rarely to be used and potentially high impact, so should probably not be one of the top options in the menu
- It's different in kind to all the other options (it's about doing something with the workspace rather than doing something to the workspace). Does this UI scheme have a separator line? Would that look good or crap? (If it does I think we should also consider organising the workspace contextual menu a bit also)

frontend/src/js/components/workspace/DownloadTextModal.tsx
```diff
@@ -0,0 +1,134 @@
...
+  return (
+    <Modal
+      isOpen={isOpen}
+      dismiss={dismiss}
+      >
+      <div style={{padding: "10px"}}>
+        <h2>Download Workspace as Text</h2>
+        <p>
+          This will download the text for all the <strong>{allBlobUris.length.toLocaleString()}</strong> files in this workspace,
+          with each file's text separated by a header indicating the file path and language.
```
hoyla: Should we put a cautionary note in this dialog? Something along the lines of: 

> Caution: please exercise appropriate security measures with any material you export from Giant.

frontend/src/js/components/workspace/WorkspaceSummary.tsx
```diff
@@ -166,6 +167,11 @@ export default function WorkspaceSummary({
               onClick={() => setTakeOwnershipModalOpen(true)}
             />
           )}
+          <Dropdown.Item
+            icon="file alternate outline"
```
hoyla: Maybe the icon should emphasise that this action will cause content to be exported? e.g. "cloud download"? I'm not sure. I like the "convert to text" notion as well, but I suspect people just see text document icons as document icons generally

frontend/src/js/components/workspace/WorkspaceSummary.tsx
twrichards: it is an admin only option (for now at least) but I can move it to the bottom and add a separator line (an `<hr />` if nothing else)

frontend/src/js/components/workspace/DownloadTextModal.tsx
twrichards: we can, but let's add that if/when we open it up beyond admins?

frontend/src/js/components/workspace/WorkspaceSummary.tsx
twrichards: again, not too worried about this until we open it up beyond admins, however I'd like to think once I've had a chance to explore the open source alternatives to NotebookLM we might aim for a more end-to-end integration, so non-admin users needn't deal with these intermediate text files

backend/app/services/index/ElasticsearchResources.scala
```diff
@@ -513,6 +513,68 @@ class ElasticsearchResources(override val client: ElasticClient, indexName: Stri
...
+          """.stripMargin.trim).lang("painless"))
+          .combineScript(Script("return state.total").lang("painless"))
+          .reduceScript(Script("long total = 0; for (s in states) { total += s } return total").lang("painless"))
+      )
+      .size(0)
+
+    execute(query).map { response =>
+      response.aggregations.data("workspaceWordCount").asInstanceOf[Map[String, Any]]("value") match {
+        case n: Number => n.longValue()
+        case other => other.toString.toLong
```
philmcmahon: in what instances would it not be a Number?

backend/app/controllers/api/Workspaces.scala
```diff
@@ -160,6 +160,16 @@ class Workspaces(
       .map(workspace => Ok(Json.toJson(workspace)))
   }
 
+  def getText(workspaceId: String) = ApiAction.attempt(parse.json) { req =>
+    for {
+      _ <- annotation.getWorkspaceMetadata(req.user.username, workspaceId) // check workspace exists and user has access
+      blobUris <- req.body.validate[List[String]].toAttempt
+      result <-
+        if (blobUris.isEmpty) index.getTotalWordCountForWorkspace(workspaceId).map(Json.toJson(_))
```
philmcmahon: this is slightly surprising behaviour for a getText endpoint - would it make sense to have a separate getWordCount endpoint?

backend/app/services/index/Index.scala
```diff
@@ -44,4 +44,7 @@ trait Index {
   def deleteWorkspace(workspaceId: String): Attempt[Unit]
 
   def updateIngestionPath(oldIngestionPath: String, newIngestionPath: String): Attempt[Unit]
+
+  def getTotalWordCountForWorkspace(workspaceId: String): Attempt[Long]
+  def getTextForBlobs(workspaceId: String, blobUris: List[String]): Attempt[Map[String, Map[String, String]]] = ???
```
philmcmahon: think you can drop the ??? here

backend/app/controllers/api/Workspaces.scala
```diff
@@ -160,6 +160,16 @@ class Workspaces(
       .map(workspace => Ok(Json.toJson(workspace)))
   }
 
+  def getText(workspaceId: String) = ApiAction.attempt(parse.json) { req =>
```
philmcmahon, line 170: should we check that the user is an admin on the server side as well? e.g. https://github.com/guardian/giant/blob/3ea5eabf5a1b76f6fcb478771a8a164e0bc63eab/backend/app/controllers/api/Blobs.scala#L105

twrichards: I did think about this, but permissions isn't currently integrated into the Workspaces controller, and we have a check that they can't fetch text from blobs not in the requested workspace (and we obviously check they have access to the workspace), the 'admin only' in the UI is just to hide it from general users rather than a strict prevention. All that said, I can add in the permission check server-side if you feel strongly @philmcmahon

philmcmahon (approved):
Great work, some small comments above but happy for this to go out

downloading

backend/app/services/index/ElasticsearchResources.scala
twrichards: good point, i've simplified now - thanks

backend/app/controllers/api/Workspaces.scala
twrichards: fair enough, I'd been trying to avoid repetition/boiler-plate but agree its surprising, I've split it out now

philmcmahon: looks great

PR #707, opened 2026-04-28
fix for `ColorConversionNeededError`

### LINKED ISSUE #679: Address problems with colour conversion in ocrmypdf for certain PDFs
We've seen 13 instances of PDFs with a colour space that causes the below issue. I think if we wanted to deal with this we would have to run ocrmypdf, detect this type of failure and then if it happens try again with `--color-conversion-strategy` set to RGB or something

However, 13 is not that many so maybe this isn't a priority

```
Error in 'OcrMyPdfExtractor processing ArbVREi_dU0nQSH-GoRwgq1dRqSEXvtVh_QiycqaCKmADPSJnbgjf3bCWI6l3KZdx1Mas2U_jVveS5j1nofZTw': java.lang.IllegalStateException: OcrMyPdfExtractor error openjpeg warning: unspec CS. 1 component so assuming gray.
openjpeg warning: unspec CS. 1 component so assuming gray.
openjpeg warning: unspec CS. 1 component so assuming gray.
Start processing 2 pages concurrently
    1 redoing OCR
    2 redoing OCR
    3 redoing OCR
    4 redoing OCR
Postprocessing...
ColorConversionNeededError: The input PDF has an unusual color space. Use
--color-conversion-strategy to convert to a common color space
such as RGB, or use --output-type pdf to skip PDF/A conversion
and retain the original color space.

	at extraction.ocr.BaseOcrExtractor.extract(BaseOcrExtractor.scala:36)
	at extraction.FileExtractor.extract(FileExtractor.scala:22)
	at extraction.Worker.safeInvokeExtractor(Worker.scala:150)
	at extraction.Worker.$anonfun$executeBatch$3(Worker.scala:94)
```

twrichards:
Paired with @philmcmahon 

Typically large/complex PDFs often have a mix of colour spaces e.g. CMYK and RGB in the same PDF and then after the worker has done all the hard/long work of OCRing it fails at the final step of converting to PDF-A since it requires a consistent colour space. 

Fixes #679 

## What does this change?
Run a lightweight pre-run of `ocrmypdf` to detect `ColorConversionNeededError` and if detected then pass `--color-conversion-strategy=RGB` to the actual main run of `ocrmypdf` to avoid it failing at the very end of OCRing

 - [ ] Tested locally
 - [x] Tested on playground

## Things to think about
This will add a few seconds to every OCR, but still more efficient than a % of all PDFs failing after a lot of wasted work.

### DIFF
--- a/backend/app/utils/Ocr.scala
+++ b/backend/app/utils/Ocr.scala
@@ -181,9 +181,30 @@ object Ocr extends Logging {
       }
     }
 
+    def calculateColorConversionStrategyFlag(path: Path): String = {
+      val startTime = System.currentTimeMillis()
+      var colorConversionStrategyFlag = ""
+      Process(
+        s"ocrmypdf --skip-text --output-type pdfa --pages 1 --tesseract-timeout 0 ${path.toAbsolutePath} /dev/null"
+      ) ! ProcessLogger ( log =>
+        if (log.contains("ColorConversionNeededError:")) {
+          colorConversionStrategyFlag = "--color-conversion-strategy=RGB"
+        }
+      )
+      val duration = System.currentTimeMillis() - startTime
+      logger.info(s"Checked if color conversion strategy flag is needed. Took ${duration}ms. Result: $colorConversionStrategyFlag")
+      colorConversionStrategyFlag
+    }
+
     def ocrWithOcrMyPdf(flag: OcrMyPdfFlag, overrideFile: Option[Path] = None): Int = {
-      val sourceFilePath = overrideFile.getOrElse(inputFilePath)
-      val cmd = s"ocrmypdf ${flag.flag} -l $lang ${dpi.map(dpi => s"--image-dpi $dpi").getOrElse("")} ${sourceFilePath.toAbsolutePath} ${tempFile.toAbsolutePath}"
+      val sourceFilePath = overrideFile.getOrElse(inputFilePath).toAbsolutePath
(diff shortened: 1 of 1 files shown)


backend/app/utils/Ocr.scala
```diff
@@ -181,9 +181,30 @@ object Ocr extends Logging {
       }
     }
 
+    def calculateColorConversionStrategyFlag(path: Path): String = {
+      val startTime = System.currentTimeMillis()
+      var colorConversionStrategyFlag = ""
+      Process(
+        s"ocrmypdf --skip-text --output-type pdfa --pages 1 --tesseract-timeout 0 ${path.toAbsolutePath} /dev/null"
```
philmcmahon, line 188: when we were testing we did `${path.toAbsolutePath} - > /dev/null` though this looks like it would work as well

twrichards: yeah, the `>` was what was causing it not to work btw, `Process` doesn't use a shell and so this redirect was having undesired effects, whilst explaining other clearly documented things incorrectly...
image
... it was actually CoPilot that suggested removing the `>` and actually using `/dev/null` directly as the output (rather than `-`) ... and it works 🎉

philmcmahon: nice!

philmcmahon (approved):

philmcmahon:
colour

Issue #704, opened 2026-04-27
Support commenting in combined (pageviewer) view mode


hoyla:
- Show existence of comments across different view modes (text, combined, ocr, translation?, preview?, side-by-side transcript mode?). It may not be possible to position them exactly, but it'd still be good to know that they exist (and if all else fails click through to them in text view to see what they highlight)
- Allow highlight and comment (and other supported highlight options) from combined mode

Issue #702, opened 2026-04-27
Link to comment


hoyla:
Allow users to direct colleagues to a passage within a document by highlight, comment, link to comment

(c.f. snippets, link to highlight, etc - but I think the above is probably much easier)

Issue #701, opened 2026-04-27
Show/hide/minimise/maximise comments


hoyla:
If we want to show comments in the combined viewer, we can't just reflow the text to make space.

<img width="1149" height="553" alt="Image" src="https://github.com/user-attachments/assets/ae95d93c-a7d9-4c10-8841-fc630e987218" />

PR #711, opened 2026-04-29
Giant document view: show list of comments in sidebar

### LINKED ISSUE #700: Rollover for text selection options
Image

What do these do? Do we really want to search the whole of Giant for this term?

### LINKED ISSUE #705: Document-level comments
Not for a a highlight. For a document.

### LINKED ISSUE #708: Comments visible in one text view are not visible in another
If I make a comment in the Text view mode of a document, I can't see it in the OCR view, and vice versa. I think the same goes for transcriptions (and translations thereof). Plus as we know I can't see comments at all in the Combined, Preview, Video and Audio views.

### LINKED ISSUE #709: Someone viewing a document doesn't know it contains comments unless they happen to be onscreen
If a user gets to a document from search, they are taken to the first instance of the search result in a document. If another user has commented elsewhere in the document, there's no indication that such a comment exists.

Similarly, if a user gets to a document from a document link or by double clicking on it in a file tree (dataset or workspace), they're shown the start of the document. If comments exist lower down, there's no indication of that to the user. 

Basically you have to see a comment to know a comment exists, which reduces their utility.

hoyla:
An attempt to fix or improve upon the following:

- Our preferred document view, "Combined", doesn't support the comments feature in any way. You wouldn't even know comments existed. https://github.com/guardian/giant/issues/704
- Users can't make comments about documents; only about text highlights in documents. https://github.com/guardian/giant/issues/705
- If a user makes a comment in "Text" view mode, it's not visible in OCR (or any other text based) view mode; and vice versa. https://github.com/guardian/giant/issues/708
- Even in view modes where comments are supported, there is no indication that comments exist elsewhere in a document. #709 
- When users select some text in a text based view mode, they get two icons but not much signal for what the icons do. Add some rollover text for these tools. https://github.com/guardian/giant/issues/700
- There's no facility to "see all the comments for this document".

**Before:**

https://github.com/user-attachments/assets/0b35428c-ff7c-4c2f-b875-0fa0da50c5c0

**After:**

I haven't attempted inline comments in the PageViewer, because exchanging the positions across different view modes would be very clunky. But you can at least see all comments in all views, and navigate to them in situ in the text view modes.

Plus you can make unanchored comments - i.e. about the whole document - in any view mode.

## Changes

- **Unified comment sidebar** — shows all comments across all views with view tags (Text/OCR/Document), text snippets, collapse/expand, and document-level comment creation (#701, #704)
- **Text selection tooltips** — added title attributes to search and comment popover buttons (#700)
- **Delete confirmation dialog** — comments now require confirmation before deletion, showing author and text (#705)
- **Scroll-to-comment** — clicking an anchored comment in the sidebar switches to the correct view and scrolls to the highlight (#702)
- **Dev Tools moved to bottom** of sidebar so comments sit in a more prominent position
- **Bug fix** — document-level comments (no anchor) no longer crash the text view highlight mapper

## Testing

- [x] Integration tests: 171 lines of new tests for `commentUtils` utility functions
- [x] Played with it on local for various document types
- [x] Playground

### DIFF
--- a/frontend/src/js/components/viewer/CommentHighlighter.tsx
+++ b/frontend/src/js/components/viewer/CommentHighlighter.tsx
@@ -104,6 +104,8 @@ function HighlightWrapper({
     {
       class: focused ? `${elementType}--focused` : "",
       "data-highlight-offset": highlight.range.startCharacter,
+      "data-comment-id":
+        highlight.type === "comment" ? highlight.id : undefined,
       ref: onMountOrUnmount,
       onClick: (e: React.MouseEvent) => {
         if (highlight.type === "comment") {
--- a/frontend/src/js/components/viewer/CommentPanel/CommentPanel.tsx
+++ b/frontend/src/js/components/viewer/CommentPanel/CommentPanel.tsx
@@ -12,6 +12,7 @@ import {
 import { AddComment } from "./AddComment";
 import { filterCommentsInView } from "../../../util/commentUtils";
 import { HighlightRenderedPositions } from "../TextPreview";
+import ModalAction from "../../UtilComponents/ModalAction";
 
 import { sortBy, sumBy, takeWhile } from "lodash";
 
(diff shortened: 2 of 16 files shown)


philmcmahon (approved):
Fantastic, code looks ok to me and I see you have tested in playground

no comment

PR #712, opened 2026-04-29
Add separate manifest for fetchWork test to stop it randomly failing

philmcmahon:
## What does this change?
The fetchWork test had a bug where it would randomly fail because the two workers would pick up less than 5 bits of work (possibly also more than 5 bits of work). 

I think this was caused by there being lots of random stuff in the neo4j db from the other tests and the neo4j sorting being nondeterministic if you don't tell it what to sort by.

I am hoping this change will fix the problem, by having a separate neo4j manifest for this fetchWork test

 - [x] Tested locally
 - [ ] Tested on playground

### DIFF
--- a/backend/test/services/manifest/Neo4JManifestITest.scala
+++ b/backend/test/services/manifest/Neo4JManifestITest.scala
@@ -44,6 +44,7 @@ class Neo4JManifestITest extends AnyFreeSpec
   val ingestionServices: IngestionServices = stub[IngestionServices]
 
   var manifest: Manifest = _
+  var fetchWorkTestManifest: Manifest = _
 
   override type Containers = Neo4jContainer
 
@@ -53,6 +54,12 @@ class Neo4JManifestITest extends AnyFreeSpec
     val neo4jDriver = new Neo4jTestService(neo4jContainer.container.getBoltUrl).neo4jDriver
 
     manifest = {
+      Neo4jManifest.setupManifest(neo4jDriver, global, new Neo4jQueryLoggingConfig(1.second, logAllQueries = false)).toOption.get
+    }
+
+    //fetchWork test needs its own manifest as we need to be sure of the number of work items available - otherwise
+    // the concurrency check may pass just because two workers have randomly picked different blobs
+    fetchWorkTestManifest = {
       Neo4jManifest.setupManifest(neo4jDriver, global, new Neo4jQueryLoggingConfig(1.second, logAllQueries = false)).toOption.get
     }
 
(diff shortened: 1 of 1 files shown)


backend/test/services/manifest/Neo4JManifestITest.scala
```diff
@@ -416,17 +423,17 @@ class Neo4JManifestITest extends AnyFreeSpec
...
-          blob(ingestion.chain("every.zip"), List(extractors("ArchiveExtractor")), ingestion.value),
-          blob(ingestion.chain("second.zip"), List(extractors("ArchiveExtractor")), ingestion.value),
-          blob(ingestion.chain("counts.zip"), List(extractors("ArchiveExtractor")), ingestion.value),
-          blob(ingestion.chain("fridge.zip"), List(extractors("ArchiveExtractor")), ingestion.value),
-          blob(ingestion.chain("time.zip"), List(extractors("ArchiveExtractor")), ingestion.value)
+          blob(ingestion.chain("every.zip"), List(extractors("ArchiveExtractor")), ingestion.value, size = 100L),
+          blob(ingestion.chain("second.zip"), List(extractors("ArchiveExtractor")), ingestion.value, size = 100L),
+          blob(ingestion.chain("counts.zip"), List(extractors("ArchiveExtractor")), ingestion.value, size = 100L),
+          blob(ingestion.chain("fridge.zip"), List(extractors("ArchiveExtractor")), ingestion.value, size = 100L),
+          blob(ingestion.chain("time.zip"), List(extractors("ArchiveExtractor")), ingestion.value, size = 100L)
```
philmcmahon, line 434: I made the size of these blobs really small to be extremely sure that they fit inside a single fetchWork call ( the default size is 1024)

zekehuntergreen (approved):
> neo4j sorting being nondeterministic if you don't tell it what to sort by 

Is it worth considering changing the query so that is has deterministic ordering? Seems like if there's a way to avoid having two manifests in the test that might be preferable.

philmcmahon:
> > neo4j sorting being nondeterministic if you don't tell it what to sort by
> 
> Is it worth considering changing the query so that is has deterministic ordering? Seems like if there's a way to avoid having two manifests in the test that might be preferable.

Yes it would be better to enforce a sort order - we were thinking of adding a createdTime or something to our TODOs so that we can order them by recency. I think that would be a separate bit of work to fixing this test though.

Really the problem here is that our tests don't clean up after themselves so stuff in the db from previous tests interferes with future tests. There would be a cost to cleaning up, but it would be nice to have a 'clean up' query that just drops everything in the manifest that you can optionally run at the start of a test

PR #716, opened 2026-05-01
Fix/add document viewer keyboard shortcuts

hoyla:
We use mousetrap to intercept keyboard events when the user is not in an input field. Our convention is shfift + character, e.g. shift right does "next document in search/worksace". We had shift X and shift C when there were only two document views, but shift C didn't work. This adds shorcuts for all four main view types.

- Shift+T: Switch to Text view (was Shift+X)
- Shift+C: Switch to Combined view (new)
- Shift+P: Switch to Preview/Audio/Video view (was Shift+C, broken)
- Shift+O: Switch to first OCR/Transcript view (new)

If there are multiple OCR or transcript views, the first one is activated. 

Fix: the `showPreview` handler called a non-existent `this.supported()` method, meaning the preview shortcut has never worked. Replaced it with a correct `this.canPreview()` guard.

Add new handlers for Combined and OCR/Transcript views with appropriate availability guards.

### DIFF
--- a/frontend/src/js/components/viewer/PreviewSwitcher.js
+++ b/frontend/src/js/components/viewer/PreviewSwitcher.js
@@ -105,8 +105,31 @@ class PreviewSwitcher extends React.Component {
     this.props.setResourceView("text");
   };
 
+  showCombined = () => {
+    if (this.props.totalPages > 0) {
+      this.props.setResourceView("combined");
+    }
+  };
+
   showPreview = () => {
-    if (this.supported()) this.props.setResourceView("preview");
+    if (this.canPreview(this.props.resource?.previewStatus)) {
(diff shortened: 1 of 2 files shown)


frontend/src/js/components/viewer/PreviewSwitcher.js
```diff
@@ -105,8 +105,31 @@ class PreviewSwitcher extends React.Component {
     this.props.setResourceView("text");
   };
 
+  showCombined = () => {
+    if (this.props.totalPages > 0) {
+      this.props.setResourceView("combined");
```
philmcmahon, line 111: maybe a little comment here to explain why having totalPages > 0 means that combined view will be available

philmcmahon (approved):
shortcut

Issue #719, opened 2026-05-01
Is there still a problem with overlapping pages when combined viewer displays landscape documents?


hoyla:
I thought we had fixed this but it has been reported again

hoyla:
This was due to a rotation property within document PDFs, meaning their orientation is misreported when calculating depth. See #720 (which is hopefully fixed by #721 )

PR #721, opened 2026-05-01
Giant doc viewer: fix page overlap for PDFs with /Rotate metadata

### LINKED ISSUE #720: Giant doc view: problem with PDFs that have a hard-coded rotation value
The mechanism for determining the dimensions of a PDF appear to get into trouble when the PDF has an inbuilt rotation value. So for example, this document:

Height: 595
Width: 842
Rotation: 90°

Will get rendered as 

Height: 595
Width: 842
No rotation

Meaning that once "scaled to fit to width, minimum 1000px:, the image of each page is cropped vertically to the canvas area and each page overlaps. 

See e.g. 

https://giant.pfi.gutools.co.uk/viewer/N4H22roWYmb8bgSYan2SpDNK9YP0VC2uhz_7-pg8uk9dqHRHafmYQmMv6IslBR9K1ifIa29oA-N4fmxHt46Srg?navId=98fbb5d8-48cd-47ad-a343-d6daca004633&navIndex=4&view=combined

These pages are actually portrait. But this:

Image

<br/>
shows in Giant like this:

Image

hoyla:
Fixes #720 and, hopefully, #719 (if those are indeed the same problem)

The backend OCR extractors used `PDFBox`'s `getMediaBox()` which returns raw dimensions ignoring the PDF's `/Rotate` property. This caused a mismatch with `pdfjs` (which applies rotation), leading to incorrect slot heights in the virtual scroll layout and overlapping pages.

Backend:
- OcrMyPdfExtractor and TesseractPdfOcrExtractor now account for page rotation when storing dimensions, swapping width/height for 90/270 degree rotations.

Frontend:
- VirtualScroll resolves the true page dimensions from the first `pdfjs` preview (which always respects `/Rotate`) and uses those for layout, fixing overlap for existing indexed documents.
- Removed hardcoded `min-height: 1000px` from page container CSS which was a leftover from before real dimensions were available.
- Set explicit height on page containers based on calculated slot height.

Before:

Picture 360

After:

Picture 359

### DIFF
--- a/backend/app/extraction/ocr/OcrMyPdfExtractor.scala
+++ b/backend/app/extraction/ocr/OcrMyPdfExtractor.scala
@@ -71,12 +71,15 @@ class OcrMyPdfExtractor(scratch: ScratchSpace, index: Index, pageService: Pages,
       val (pages, _) = (1 to numberOfPages).foldLeft(base) { case ((pages, offsetHeight), pageNumber) =>
         val page = firstDoc.getPage(pageNumber - 1)
         val pageBoundingBox = page.getMediaBox
+        val isRotatedSideways = page.getRotation % 180 != 0
+        val effectiveWidth = if (isRotatedSideways) pageBoundingBox.getHeight else pageBoundingBox.getWidth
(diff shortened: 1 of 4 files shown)


philmcmahon (approved):
rotate the board

PR #722, opened 2026-05-04
Add admin-only option to reprocess folder contents in workspace

hoyla:
Add a new context menu item 'Reprocess folder contents…' for workspace folders, visible only to users with CanPerformAdminOperations permission. 

Picture 364
 ____

Shows a confirmation dialog indicating the total file count; the number with errors; and providing three options: Reprocess Errored (only shown when there are errors), Reprocess All, and Cancel.

Picture 365

### Backend:
- Add `workspaceTreeToFailedBlobIds` to `TreeEntry` to collect only failed blob URIs from a workspace tree
- Add `reprocessFolder` endpoint on Workspaces controller, gated behind `CanPerformAdminOperations`, supporting both all and errored-only modes via rerunSuccessful/rerunFailed query params
- Add route: POST /api/workspaces/:id/folders/:folderId/reprocess
- Add tests for `workspaceTreeToFailedBlobIds`

### Frontend:
- Add `reprocessFolder` API function in WorkspaceApi with mode parameter
- Add `reprocessFolder` Redux action supporting 'all' and 'errored' modes
- Add `ReprocessFolderModal` confirmation dialog component
- Add 'Reprocess folder contents…' context menu item for folders, visible only to admins.

 - [x] Tested locally
 - [x] Tested on playground

## Things to think about

### DIFF
--- a/backend/app/model/frontend/TreeEntry.scala
+++ b/backend/app/model/frontend/TreeEntry.scala
@@ -1,7 +1,7 @@
 package model.frontend
 
 import model.Uri
-import model.annotations.{WorkspaceEntry, WorkspaceLeaf}
+import model.annotations.{ProcessingStage, WorkspaceEntry, WorkspaceLeaf}
 
 sealed trait TreeEntry[+T] {
   def id: String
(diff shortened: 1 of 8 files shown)


backend/app/model/frontend/TreeEntry.scala
```diff
@@ -64,6 +64,15 @@ object TreeEntry {
     }
     case treeNode: TreeNode[WorkspaceEntry] => treeNode.children.flatMap(workspaceTreeToBlobIds)
   }
+
+  def workspaceTreeToFailedBlobIds(tree: TreeEntry[WorkspaceEntry]): List[Uri] = tree match {
```
philmcmahon: getFailedBlobUris

philmcmahon (approved):
Such a nice feature! I am a bit concerned that reprocess is so broken that this will deliver mixed results, but for workspaces with just a few failures it seems worth it

PR #724, opened 2026-05-05
Fix ocr when running locally on macos

philmcmahon:
## What does this change?
There's a long standing issue around running tesseract/leptonica on macos where it can't deal with files in the `/tmp` directory - see https://github.com/tesseract-ocr/tesseract/issues/4333

This PR works round that by overriding the ingestion.scratch location locally to point at a directory other than /tmp

 - [x] Tested locally
 - [ ] Tested on playground

#


hoyla:
Having done this manually about four times now, THANK YOU

zekehuntergreen (approved):

PR #725, opened 2026-05-05
Add language detection functionality to giant

philmcmahon:
## What does this change?

In preparation for integrating giant with the new translation service https://github.com/guardian/transcription-service/pull/284 we need to know what language documents are in so we can decide whether or not they need translation.

Here I'm making use of [tikka's language detection functionality](https://tika.apache.org/2.3.0/detection.html#Language_Detection) to detect the language of a piece of text. I don't think this will be very reliable compared to other language detection methods, but it is quick and minimises additional dependencies. Giant really just needs to know whether or not a file is in english or not - any non-english files will be sent for translation.

Note that this PR doesn't have any user facing changes - it just adds some new fields to elasticsearch with the detected language information

## Language support in giant
Giant has long had awareness of languages. This is because tesseract needs to know the language of a PDF file in order to more accurately OCR it. Sam added language support to giant back here https://github.com/guardian/pfi/pull/327 with this aim in mind. 

The existing language support is at the `Ingestion` level - so you set one or more languages for a batch of files. This is quite wasteful - if you have an ingestion with 1 french file and 500 english files and you set the ingestion languages to 'french,english' then giant will run OCR for french and english on every file in the ingestion, essentially doubling the cost of OCR for the ingestion. There's no real way round this though unless you know the language of a document up front, as no metadata will tell you the language of the file, and we know that tesseract (via ocrmypdf) will perform poorly on cyrillic documents if you don't warn it what language you want to OCR in.

This PR introduces a document level language field to elasticsearch at `metadata.enrichedMetadata.detectedLanguageCode`, which is added by the DocumentBodyExtractor. This field will get set wherever tikka manages to detect the language of a document with a high degree of confidence. 

If `metadata.enrichedMetadata.detectedLanguageCode` is defined and set to a [supported language in giant](https://github.com/guardian/giant/blob/b007841285a525c5a13e61219c9caa5b0d0e66ed/common/src/main/scala/model/Language.scala#L42) at the point we come to OCR a document, then we pass that value to ocrmypdf/tesseract and ignore any top level languages set on the ingestion.

Note: there is no enforced ordering the DocumentBodyExtractor/OCRExtractor extractors. The DocumentBodyExtractor has a higher priority so where we only have 1 worker it will always happen first, but if we have e.g 2 workers and 1 document uploaded then OCR and DocumentBody extraction will happen at the same time so OCR will fall back to the Ingestion languages.

## OCR language detection
Given the above, we can assume that in some situations we will have multiple versions of the OCR text for a given document - one for each language specified on the ingestion. So when we do language detection we need to run it for the output of the ocr for each language, resulting in multiple different 'OCR detectedLanguageCode' values - one for each language.

To store this data in elasticsearch I've added a new ocrMetadata field within the metdata object with a `languages` property which is a map of language->OcrLanguageMetadata where OcrLanguageMetadata is an object with (currently) a single field - detectedLanguageCode. In future we could expect to add 'translationExtracted' or similar to this metadata field to track translations.

## Updating elasticsearch mapping
This PR requires changes to the elasticsearch mapping. This will be applied automatically to a new elasticsearch cluster but existing clusters will need to update the mapping like this:

```bash
curl -X PUT "localhost:9200/pfi/_mapping?pretty" -H 'Content-Type: application/json' -d'
{
  "properties": {
    "languageData": {
      "properties": {
        "text": {
          "properties": {
            "detectedLanguageCode": { "type": "text" },
            "translation": { "type": "text" }
          }
        },
        "emailBody": {
          "properties": {
            "detectedLanguageCode": { "type": "text" },
            "translation": { "type": "text" }
          }
        },
        "emailSubject": {
          "properties": {
            "detectedLanguageCode": { "type": "text" },
            "translation": { "type": "text" }
          }
        }
      }
    }
  }
}
```

## Language detection notes
In testing, I noticed that the DocumentBodyExtractor failed to detect the language for [this  PDF](https://github.com/user-attachments/files/27401231/Toast.sandwich.Wikipedia.pdf) (it guessed english but at not enough certainty) where it had succeded when that same document had been pasted in to a .txt file. The language detection of the OCR did work - possibly because it had all the urls dropped. With that in mind, I added a few AI generated regexs to drop various strings that are likely to be unhelpful when running language detection on a document

 - [x] Tested locally
 - [x] Tested on playground

## Things to think about

### DIFF
--- a/backend/app/extraction/DocumentBodyExtractor.scala
+++ b/backend/app/extraction/DocumentBodyExtractor.scala
@@ -1,17 +1,18 @@
 package extraction
 
 import java.io.InputStream
-
 import model.manifest.Blob
+import org.apache.tika.language.detect.LanguageDetector
 import services.Tika
 import services.index.Index
+import services.ingestion.IngestionServices
 import utils.Logging
 import utils.attempt.AttemptAwait._
 import utils.attempt.{Failure, UnsupportedOperationFailure}
 
 import scala.concurrent.ExecutionContext
 
-class DocumentBodyExtractor(tika: Tika, index: Index)(implicit ec: ExecutionContext) extends Extractor with Logging {
+class DocumentBodyExtractor(tika: Tika, index: Index, ingestionServices: IngestionServices)(implicit ec: ExecutionContext) extends Extractor with Logging {
   val mimeTypes: Set[String] = Set(
     "application/html",
     "application/json",
@@ -50,7 +51,8 @@ class DocumentBodyExtractor(tika: Tika, index: Index)(implicit ec: ExecutionCont
     if(passesSafetyCheck(mimeType, blob.size)) {
       tika.parse(stream, mimeType).flatMap { case (metadata, body) =>
         val rawMetadata = metadata.names().map(name => name -> metadata.getValues(name).toSeq).toMap
-        val enrichedMetadata = MetadataEnrichment.enrich(rawMetadata)
+        val detectedLanguage = ingestionServices.detectLanguage(blob.uri.value, body.trim())
+        val enrichedMetadata = MetadataEnrichment.enrich(rawMetadata, detectedLanguage)
         // Optionally having a body will allow documents without text to default to preview, useful for un-OCR'd documents
         val optionalBody = if (body.trim().isEmpty) None else Some(body)
 
--- a/backend/app/extraction/DocumentBodyExtractor.scala
+++ b/backend/app/extraction/DocumentBodyExtractor.scala
@@ -52,7 +52,6 @@ class DocumentBodyExtractor(tika: Tika, index: Index, ingestionServices: Ingesti
       tika.parse(stream, mimeType).flatMap { case (metadata, body) =>
         val rawMetadata = metadata.names().map(name => name -> metadata.getValues(name).toSeq).toMap
         val documentBodyDetectedLanguage = ingestionServices.detectLanguage(blob.uri.value, body.trim())
-        println("DETECTED LANGUAGE: " + documentBodyDetectedLanguage)
         val enrichedMetadata = MetadataEnrichment.enrich(rawMetadata)
         // Optionally having a body will allow documents without text to default to preview, useful for un-OCR'd documents
         val optionalBody = if (body.trim().isEmpty) None else Some(body)
--- a/backend/app/extraction/DocumentBodyExtractor.scala
+++ b/backend/app/extraction/DocumentBodyExtractor.scala
@@ -51,12 +51,13 @@ class DocumentBodyExtractor(tika: Tika, index: Index, ingestionServices: Ingesti
     if(passesSafetyCheck(mimeType, blob.size)) {
       tika.parse(stream, mimeType).flatMap { case (metadata, body) =>
         val rawMetadata = metadata.names().map(name => name -> metadata.getValues(name).toSeq).toMap
-        val detectedLanguage = ingestionServices.detectLanguage(blob.uri.value, body.trim())
-        val enrichedMetadata = MetadataEnrichment.enrich(rawMetadata, detectedLanguage)
+        val documentBodyDetectedLanguage = ingestionServices.detectLanguage(blob.uri.value, body.trim())
+        println("DETECTED LANGUAGE: " + documentBodyDetectedLanguage)
+        val enrichedMetadata = MetadataEnrichment.enrich(rawMetadata)
         // Optionally having a body will allow documents without text to default to preview, useful for un-OCR'd documents
         val optionalBody = if (body.trim().isEmpty) None else Some(body)
 
-        index.addDocumentDetails(blob.uri, optionalBody, rawMetadata, enrichedMetadata, params.languages).awaitEither()
+        index.addDocumentDetails(blob.uri, optionalBody, rawMetadata, enrichedMetadata, params.languages, documentBodyDetectedLanguage).awaitEither()
       }
     } else {
       Left(UnsupportedOperationFailure("Failed safety check"))
--- a/backend/app/extraction/ocr/BaseOcrExtractor.scala
+++ b/backend/app/extraction/ocr/BaseOcrExtractor.scala
@@ -1,28 +1,41 @@
 package extraction.ocr
 
 import extraction.{ExtractionParams, FileExtractor}
+import model.Languages
 import model.manifest.Blob
+import org.apache.tika.language.detect.LanguageDetector
 import services.ScratchSpace
+import services.index.Index
 import utils.Ocr.{OcrMyPdfTimeout, OcrSubprocessInterruptedException}
 import utils.OcrStderrLogger
 import utils.attempt.{Failure, OcrTimeout, SubprocessInterruptedFailure}
 
 import java.io.File
+import scala.concurrent.duration.DurationInt
+import scala.concurrent.{Await, ExecutionContext}
 import scala.util.control.NonFatal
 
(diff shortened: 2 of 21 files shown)


philmcmahon:
**UPDATE**

Following an IRL discussion about the best way to implement this, I realised that there is actually a reasonably large number of fields in the giant elasticsearch model that we might consider translating. With that in mind, rather than putting detected language code into enrichedMetdata/ocrMetadata I've pulled it out into a separate top level `languageData` field which has the below structure:

```
{
   ...otherFields
   languageData: {
      field: {
         detectedLanguageCode: "code",
         translation: "currently_always_undefined",
      }
   }
}
```

Here it is for an EML file:

Screenshot 2026-06-11 at 22 43 24

Here it is for a french PDF document:

Screenshot 2026-06-11 at 22 43 39

I'm hoping this sufficiently flexible to allow for more fields to easily be added. 

The new elasticsearch mapping can be applied using the script below - note that the ocr field properties are handled by the `addLanguage` function in ElasticsearchResources.scala


```
curl -X PUT "localhost:9200/pfi/_mapping?pretty" -H 'Content-Type: application/json' -d'
{
  "properties": {
    "languageData": {
      "properties": {
        "text": {
          "properties": {
            "detectedLanguageCode": { "type": "text" },
            "translation": { "type": "text" }
          }
        },
        "emailBody": {
          "properties": {
            "detectedLanguageCode": { "type": "text" },
            "translation": { "type": "text" }
          }
        },
        "emailSubject": {
          "properties": {
            "detectedLanguageCode": { "type": "text" },
            "translation": { "type": "text" }
          }
        }
      }
    }
  }
}

```

twrichards:
> Here it is for a french PDF document:
> 
> Screenshot 2026-06-11 at 22 43 39

How come the ocr one has a further level of nesting with key as the full language name?

philmcmahon:
> > Here it is for a french PDF document:
> > Screenshot 2026-06-11 at 22 43 39
> 
> How come the ocr one has a further level of nesting with key as the full language name?

OCR was the complicated one here. The languages for an ingestion are set at the ingestion level, so if uploading lots of french and spanish files we would set the ingestion languages to french,spanish. When we come to OCR a document, assuming it is an image with no machine readable text then we'll OCR it with tesseract in french, and then again in spanish. We then need to run language detection on the output OCR from both french and spanish, which leads to this object:
```
ocr: {
   french: "bonjour",
   spanish: "bonjour",
   }
}
languageData: {
...,
ocr: {
   detectedLanguageCode: {
     "french" : fr,
     "spanish": fr
  },
  translation: {
      french: "good day", 
     //spanish translation probably never happens because both OCR output were detected to be french
}
```

backend/app/extraction/ocr/BaseOcrExtractor.scala
```diff
@@ -1,28 +1,41 @@
...
 
-abstract class BaseOcrExtractor(scratchSpace: ScratchSpace) extends FileExtractor(scratchSpace) {
+abstract class BaseOcrExtractor(scratchSpace: ScratchSpace, index:Index)  (implicit ec: ExecutionContext)  extends FileExtractor(scratchSpace) {
   def extractOcr(blob: Blob, file: File, params: ExtractionParams, stdErrLogger: OcrStderrLogger): Unit
   def buildStdErrLogger(blob: Blob): OcrStderrLogger
 
   final override def extract(blob: Blob, file: File, params: ExtractionParams): Either[Failure, Unit] = {
-    if (params.languages.isEmpty) {
+    // extractors are synchronous so we have to await here
+    val detectedLanguageCode = Await.result(index.getTextDetectedLanguage(blob.uri).asFuture, 3.seconds).toOption
```
twrichards, line 24: is 3 seconds a long enough timeout?

backend/app/extraction/ocr/BaseOcrExtractor.scala
```diff
@@ -1,28 +1,41 @@
...
+
+    if (params.languages.isEmpty && detectedLanguageCode.isDefined) {
       throw new IllegalStateException(s"${this.name} requires a language")
     }
 
+    // if we have detected a supported language code, use that, otherwise OCR in every language set for the ingestion
```
twrichards, line 30: 👍

backend/app/extraction/DocumentBodyExtractor.scala
```diff
@@ -50,11 +51,13 @@ class DocumentBodyExtractor(tika: Tika, index: Index)(implicit ec: ExecutionCont
     if(passesSafetyCheck(mimeType, blob.size)) {
       tika.parse(stream, mimeType).flatMap { case (metadata, body) =>
         val rawMetadata = metadata.names().map(name => name -> metadata.getValues(name).toSeq).toMap
+        val documentBodyDetectedLanguage = ingestionServices.detectLanguage(blob.uri.value, body.trim())
+        println("DETECTED LANGUAGE: " + documentBodyDetectedLanguage)
```
twrichards: does `println` need to be a logger line or can/should it be removed

backend/app/services/index/ElasticsearchResources.scala
twrichards, line 1: does the shape of the ES mapping changes need to be updated in the PR description ( to reflect the final shape, in light of the comments further down )

backend/app/services/ingestion/IngestionServices.scala
```diff
@@ -45,12 +46,50 @@ trait IngestionServices {
...
+      .replaceAll("""[\w.+-]+@[\w.-]+\.\w+""", "")        // email addresses
+      .replaceAll("""[A-Za-z]:\\[\\\w.\-]+""", "")        // Windows paths
+      .replaceAll("""/[\w./\-]+""", "")                   // Unix paths (conservative)
+      .replaceAll("""<[^>]+>""", "")                      // HTML/XML tags
+      .replaceAll("""[0-9a-fA-F]{16,}""", "")             // long hex strings
+      .replaceAll("""[A-Za-z0-9+/=]{40,}""", "")          // base64-like strings
+      .replaceAll("""\b\d+\b""", "")                      // standalone numbers
+      .replaceAll("""\s{2,}""", " ")                      // collapse whitespace
+      .trim
+  }
```
twrichards, line 67: I would struggle to think of all the permutations of all these regex rules (thought the comments are of course helpful) - were these AI generated and/or have they been checked with AI and/or a set of sample files?

twrichards (approved):
LGTM (few comments) - great work 🙌 

My initial hesitation around the limits/efficacy of tika (and why I spent a bit of time exploring alternatives approaches) is largely assuaged by this pertinent line from the description...
> Giant really just needs to know whether or not a file is in english or not - any non-english files will be sent for translation.

backend/app/services/index/ElasticsearchResources.scala
philmcmahon: yes good shout will do

backend/app/services/ingestion/IngestionServices.scala
philmcmahon: They were AI generated, tested on 2 documents. I've eyeballed each one to check it looked sensible, and just pushed some better documentation explaining what they actually do. 

I'm fairly relaxed about this though as it's just used to drop stuff from the text before it goes into language detection. The worst case scenario is we drop too much text and can't detect the language, that will show up down the line as documents not in english missing a translation, which isn't a disaster

backend/app/extraction/ocr/BaseOcrExtractor.scala
philmcmahon: I think so for a single elasticsearch query

PR #727, opened 2026-05-06
Scale giant workers according to available work

philmcmahon:
## What does this change?
This PR makes an adjustment to how giant scales. I keep seeing situations where we have more workers than work remaining. This changes the scaling logic so that we allocate a minimum of 2 jobs to each worker, and if the remaining work is less than 2*numWorkers then we scale down, otherwise we scale up

 - [ ] Tested locally
 - [ ] Tested on playground

### DIFF
--- a/backend/app/utils/WorkerControl.scala
+++ b/backend/app/utils/WorkerControl.scala
@@ -236,8 +236,6 @@ class AWSWorkerControl(config: WorkerConfig, discoveryConfig: AWSDiscoveryConfig
 
 object AWSWorkerControl {
 
-  val MINIMUM_TASKS_PER_WORKER = 2
-
   case class AsgState(
     desiredNumberOfWorkers: Int,
     lastEventTime: Long,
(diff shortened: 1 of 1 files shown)


twrichards (approved):
this is worth a try, but I notice we haven't changed the `fetchWork` function where the minimum pieces of work picked up by the worker is still 1 (which happens when the pieces of work are massive), your changes here could/should still work to solve the scaling problem though

backend/app/utils/WorkerControl.scala
```diff
@@ -267,15 +269,17 @@ object AWSWorkerControl {
     val inCooldown = state.workerAsg.lastEventTime > (now - cooldown) || state.spotWorkerAsg.lastEventTime > (now - cooldown)
     val manuallyScaledDown = state.workerAsg.desiredNumberOfWorkers == 0
 
-    val outstandingInTotal = state.outstandingFromIngestStore + state.outstandingFromTodos
+    val outstandingWork = state.outstandingFromIngestStore + state.outstandingFromTodos
+    val currentWorkers = state.workerAsg.desiredNumberOfWorkers + state.spotWorkerAsg.desiredNumberOfWorkers
```
twrichards:
```suggestion
    val currentWorkerCount = state.workerAsg.desiredNumberOfWorkers + state.spotWorkerAsg.desiredNumberOfWorkers
```

twrichards (changes requested):
After a further discussion we think it best to remove the `MINIMUM_TASKS_PER_WORKER` multiplier to keep it simple (since that's arbitrary and divergent from the `fetchWork` logic)

twrichards (approved):
much cleaner, thanks for making changes

PR #729, opened 2026-05-07
Don't treat silent files as transcription errors

philmcmahon:
## What does this change?
This PR stops giant from getting annoyed whenever it tries to get a transcription for a file with no audio. It builds on the functionality added here https://github.com/guardian/transcription-service/pull/251 where the transcription service provides a `noAudioDetected` value back in the error message if ffmpeg doesn't find an audio track.

 - [ ] Tested locally
 - [x] Tested on playground

### DIFF
--- a/backend/app/extraction/ExternalTranscriptionExtractor.scala
+++ b/backend/app/extraction/ExternalTranscriptionExtractor.scala
@@ -66,7 +66,8 @@ case class TranscriptionOutputFailure(
                                       id: String,
                                       originalFilename: String,
                                       userEmail: String,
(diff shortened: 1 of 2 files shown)


twrichards (approved):
@philmcmahon , @zekehuntergreen and I chatted about this in person a little and we think it does make sense that transcription service still considers this a failure (albeit with a better message since https://github.com/guardian/transcription-service/pull/251) and it comes to giant via the failure queue, and that giant handles this scenario as success (given an upload to giant is rarely for the sole purpose of transcription). Code LGTM.

PR #735, opened 2026-05-12
Filter out useless system files before GUI upload

### LINKED ISSUE #380: Handle files with split resource forks better
Files saved as apple doubles upload as two files, data fork and resource fork. If users attempt to click on the "._" version of the file they get told that it is too large or still processing, instead of that it doesn't contain any content.

Image

We should probably just get rid of resource forks and other apple specific dot files such as .ds_store

hoyla:
Should address #380 

Problem: when users select directories from macOS or Windows machines, these system artifacts can number in the thousands and have no value as uploaded workspace content. Filtering them pre-upload avoids wasted bandwidth and keeps the upload list clean.

## What this PR does

During upload to workspace, we tell the client to filter out OS junk files (e.g. `.DS_Store`, `Thumbs.db`, AppleDouble `._` resource forks, `__MACOSX/` directories, `.Spotlight-V100`, `.Trashes`, etc.) via the `FilePicker` component. This happens before the the upload inventory reaches the upload dialog.

## How

A new `filterSystemFiles` utility applies to the file map in both the file-input `onChange` and drag-and-drop `onDrop` handlers. The logic checks:

1. Known junk basenames (`.ds_store`, `thumbs.db`, `desktop.ini`, `.apdisk`)
2. AppleDouble resource fork prefix (`._`)
3. Junk path segments (`__macosx`, `.spotlight-v100`, `.trashes`, `.fseventsd`, `.temporaryitems`)

All matching is case-insensitive.

## Why not the CLI also?

This is deliberately **not** applied in the CLI. The CLI can be used for forensic ingestion of evidence drives, disk images, and encrypted volumes where completeness matters. Files like `.DS_Store` contain directory listing metadata (file sizes, icon positions, timestamps) and `Thumbs.db` contains image thumbnail caches. These may be useful for forensic analysis. 

The GUI upload flow serves a different audience: end-users adding files from their laptops to a workspace. Nobody wants their `.DS_Store` files there.

## Tests

- [x] Local
- [x] Playground

### DIFF
--- a/frontend/src/js/components/Uploads/filterSystemFiles.ts
+++ b/frontend/src/js/components/Uploads/filterSystemFiles.ts
@@ -1,0 +1,62 @@
+// Filenames and path segments commonly produced by macOS and Windows
+// that have no value as uploaded content.
+const JUNK_BASENAMES = new Set([
+  ".ds_store",
+  "thumbs.db",
+  "desktop.ini",
+  ".apdisk",
+]);
+
+const JUNK_PATH_SEGMENTS = ["__macosx"];
+
+// Directory names that are macOS system artifacts — if any segment
+// of the path matches one of these, the file is junk.
+const JUNK_DIRECTORY_NAMES = new Set([
+  ".spotlight-v100",
+  ".trashes",
+  ".fseventsd",
+  ".temporaryitems",
+]);
+
+function basename(path: string): string {
+  const lastSlash = path.lastIndexOf("/");
+  return lastSlash === -1 ? path : path.substring(lastSlash + 1);
+}
+
+function isSystemFile(path: string): boolean {
+  const name = basename(path).toLowerCase();
+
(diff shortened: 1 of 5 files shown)


frontend/src/js/components/Uploads/filterSystemFiles.ts
```diff
@@ -0,0 +1,43 @@
...
+  ".spotlight-v100",
+  ".trashes",
+  ".fseventsd",
+  ".temporaryitems",
+]);
+
+function basename(path: string): string {
+  const lastSlash = path.lastIndexOf("/");
+  return lastSlash === -1 ? path : path.substring(lastSlash + 1);
+}
```
philmcmahon, line 22: here's a nicer one from stackoverflow:

```
function basename(path) {
   return path.split('/').reverse()[0];
}
```

philmcmahon (approved):
Just one improvement, otherwise looks great

Issue #369, opened 2025-09-26
WORKSPACE UX IMPROVEMENTS


hoyla:

hoyla:
I'm in no way suggesting that any of this is right, but it could provide some pointers. This is Claude Opus's analysis of the problem:

# Large Workspace Performance Analysis

> **Date**: March 2026
>
> Very large workspaces — sometimes containing hundreds of thousands of items — take a very long time to serve to users and frequently fail. Most of the delay is server-side. This document analyses the root causes and proposes fixes.

## Root Cause

The entire workspace tree (every node and leaf) is fetched from Neo4j, assembled into a deeply nested JSON structure in Scala, serialised, and sent to the frontend in a **single unbounded HTTP response** with **no pagination, caching, or streaming**.

For a workspace with 100K items this means:

- ~25–35 MB JSON payload
- A Neo4j query touching 100K+ nodes with expensive per-node property lookups
- An O(N²) tree-building algorithm in Scala
- The frontend parses and stores the entire tree in Redux, then re-fetches it every 60 seconds

---

## Identified Bottlenecks

### 1. O(N²) tree assembly in `buildNode`

**File**: [`backend/app/services/annotations/Neo4jAnnotations.scala` (line 152)](backend/app/services/annotations/Neo4jAnnotations.scala)

```scala
val children = entries.filter(n => n.data.maybeParentId.contains(currentEntry.id)).map(buildNode)
```

For **every** tree node, this scans the **entire** flat entries list to find children. With N items this is O(N) per node × N nodes = **O(N²)**. For 100K items that is on the order of 10 billion comparisons.

#### Fix

Pre-build a `Map[Option[String], List[TreeEntry]]` grouped by `maybeParentId` before recursing. This makes each lookup O(1) and the overall assembly O(N):

```scala
val childrenByParent: Map[Option[String], List[TreeEntry[WorkspaceEntry]]] =
  entries.groupBy(_.data.maybeParentId)

def buildNode(currentEntry: TreeEntry[WorkspaceEntry]): TreeEntry[WorkspaceEntry] = {
  currentEntry match {
    case leaf: TreeLeaf[WorkspaceEntry] => leaf
    case node: TreeNode[WorkspaceEntry] =>
      val children = childrenByParent.getOrElse(Some(node.id), List.empty).map(buildNode)
      // ... rest of the copy logic unchanged
  }
}
```

---

### 2. Expensive Cypher query with per-node Resource lookups

[...]

PR #740, opened 2026-05-19
Giant search in folder: use targeted Cypher for blob rather than fetching the entire workspace

hoyla:
## Summary
- Replaces the full workspace-tree fetch + tree-walk used by search-within-folder with a targeted Cypher query that returns only the relevant blob URIs.
- For large workspaces (100k+ items) this avoids dragging the dominant Neo4j drain cost (c.f. [issue #369](https://github.com/guardian/giant/issues/369) — drain is ~35% of server-side time on a 30k-item workspace and ~95% of `getWorkspaceContents`) into every search-within-folder request.
- This change eliminates one of the scenarios in which we demand a full workspace tree, which the proposed work on lazy loading seeks to eliminate for large workspaces. 

## Changes
- New trait method `Annotations.getBlobUrisInWorkspaceFolder(currentUser, workspaceId, folderId): Attempt[List[String]]`.
- Neo4j implementation uses a targeted Cypher query with `[:PARENT*0..]` to walk only the relevant subtree, returning blob URIs of file descendants.
- `SearchContext.buildBlobFiltersForWorkspaceFolder` becomes a thin delegate to the new method.
- Existing error semantics preserved: workspace-not-found / folder-not-found / folder-is-a-leaf each produce the matching failure.
- Removes the now-orphaned `findBlobsInWorkspaceFolder` tree-walking helper and `SearchContextTest`.

## Notes
- Variable-length path `[:PARENT*0..]` is bounded by workspace tree depth (typically <10 levels) so should perform well; if a future profile shows it's slow we can fall back to a deeper-bounded variant.

## Testing
- [x] Local
- [x] Playground


philmcmahon (approved):
✅  nice improvement!

twrichards (approved):

PR #741, opened 2026-05-19
DO NOT MERGE - for discussion. Experiment with separating workspace structure fetch from status

hoyla:
> [!NOTE]
> The new endpoints in this PR are not actually used so should not be harmful.
> This PR allows us to test the endpoints and if we think it's a good idea then 
> we would have to actually engage them via (not insignificant) frontend changes.

## Background context

Workspace fetch was a single Neo4j query producing one JSON payload that mixed two things with very different change profiles:

- **Structure** — tree shape, names, parents, ownership. Changes only when a user adds/moves/deletes items.
- **Per-file processing status** — `Processing`/`Processed`/`Failed`. Changes constantly while ingestion runs, irrelevant for folders.

The frontend re-fetches the whole tree every 60s while *anything* is processing, meaning we pay the cost of the stable thing on every poll. Splitting them apart lets polling be cheap (status only) and lets structure become independently
cacheable.

Background and analysis is in #369; but the many comments in that thread make it atough read. The TL;DR:

1. First-pass analysis identified per-row `Resource` lookups in the contents query as the main opportunity.
2. Local benchmarks misattributed the cost to Scala-side row parsing — turned out to be misleading because local Docker has zero network between app and Neo4j.
3. Playground instrumentation against a 100k item workspace showed the real cost was **Neo4j → app drain**   (the Cypher does ~200k node-lookups against  `:Resource` and ships their  aggregated results over the network).
4. A variant-comparison spike confirmed that stripping the per-file `Resource` lookups reduces drain by **57%** at 101k items.

The instrumentation that produced (2)–(4) lives on the throwaway branch `ljh-workspace-perf-instrumentation` — it's not the basis for this PR, just the evidence behind the plan.

## What this PR does

Two new endpoints alongside the existing `/api/workspaces/:id`:

- `GET /api/workspaces/:id/structure` — tree without per-file processing state.  Same Cypher as `getWorkspaceContents` minus the per-row `Resource` lookups  (`OPTIONAL MATCH (:Resource {uri: node.uri})…`) and the `EXISTS { … EXTRACTION_FAILURE … }` subquery.
- `GET /api/workspaces/:id/status` — flat list of `{uri, processingStage, numberOfTodos, note, hasFailures}` for files only, deduped by URI.

Existing `/api/workspaces/:id` and `/nodes` paths are unchanged; nothing currently consumes the new endpoints. The frontend rewrite (parallel fetch on initial load, status-only polling, client-side roll-up recompute) is a separate session.

A new `ProcessingStage.Unknown` placeholder is set on file leaves in `/structure` — the frontend will overwrite it once `/status` lands. All folder processing roll-up counts (`descendantsProcessingTaskCount`, `descendantsFailedCount`) are pinned to 0 in `/structure` for the same reason; the frontend will recompute them client-side after the merge.

## Empirical validation (deployed to playground)

Same 101,021-item real workspace as the variant-comparison in #369. Six runs per endpoint, warm-up discarded, median of runs 2–6:

| Metric | `/get` (control) | `/structure` | `/status` |
|---|---|---|---|
| TTFB | 5.96s | **3.63s** | 3.07s |
| Total | 12.11s | 11.08s | 5.49s |
| Body bytes (uncompressed) | 46.53 MB | 46.16 MB | 17.49 MB |
| Body bytes (gzipped, `/status`) | — | — | 6.76 MB |

**`/structure` TTFB is −39% vs `/get`**, comfortably beating the −31% TTFB delta the variant-comparison predicted. Body bytes within 0.8% as predicted (structural fields — UUIDs, names, parent pointers — dominate the payload; the win is entirely on the Neo4j drain side, not on serialised size).

User-perceived "tree available" time after the frontend lands will be `max(TTFB(structure), TTFB(status)) = 3.63s` — the two endpoints fire in parallel and the tree renders as soon as structure arrives.

TTFB is the metric that matters because the ALB enforces a 60s `idle_timeout` on TTFB specifically (see `router.yaml` in `investigations-platform`). At 1M items, linear extrapolation from these numbers puts the structural endpoint comfortably under the cap; the current `/get` does not.

## Test suggestions:

- [x] Playground curl on a 101k-item real workspace: TTFB, body bytes, HTTP 200, deterministic body across runs.
- [ ] Reviewer eyes on `ProcessingStage.Unknown` as the structure placeholder — shape, JSON encoding, and the small handful of new pattern-match cases (all existing `match` sites on `ProcessingStage` use `case _ =>` fallbacks, so adding the variant is safe).

### DIFF
--- a/backend/app/controllers/api/Workspaces.scala
+++ b/backend/app/controllers/api/Workspaces.scala
@@ -3,7 +3,7 @@ package controllers.api
 import software.amazon.awssdk.services.sns.SnsClient
 import commands.DeleteResource
 import model.Uri
-import model.annotations.{ProcessingStage, Workspace, WorkspaceEntry, WorkspaceLeaf}
+import model.annotations.{ProcessingStage, Workspace, WorkspaceEntry, WorkspaceFileStatus, WorkspaceLeaf}
 import model.frontend.{TreeEntry, TreeLeaf, TreeNode}
 import model.ingestion.{ RemoteIngest, RemoteIngestStatus}
 import model.user.UserPermission.CanPerformAdminOperations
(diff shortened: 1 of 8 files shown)


philmcmahon:
I had a look through this, sounds sensible and the code is either mostly copy of existing code (the structure endpoint, which I didn't review very much) or short enough to review (status endpoint). I like the test that makes sure the output matches the /get endpoint.

Would be happy for this to go out, but I see that it probably doesn't make sense without the follow up PR otherwise we'll be maintaining code for endpoints that aren't used.

philmcmahon:
Hang on, I was thinking this was the first part of https://github.com/guardian/giant/pull/745 but it's actually a separate thing, will look at the lazy loading stuff now

Issue #744, opened 2026-05-23
[PLAN FOR ASSESSMENT] Workspace lazy loading


hoyla:
# Workspace lazy loading — staged implementation plan

For discussion.

**What this needs from reviewers:** (1) a code review of open **PR #745** (Stage 2 — ~600 additive lines including tests, with access-control and eager/lazy-parity coverage; reviewing it commits you to nothing beyond Stage 2); (2) a view on **Key decision 4** (incremental mutations land before the lazy flip); (3) a view on **Key decisions 1–2** (the size gate, and accepting the interim counts regression above it).

> **Note:** of the 11 stages below, **stages 1–6 (plus the "cheap counts" subset of Stage 8) have been validated by a throwaway POC** (branch `ljh-lazy-loading-poc`; see `POC_KNOWN_LIMITATIONS.md` on that branch). On playground it opened a **101k-item workspace's top level in ~2s (vs ~20s eager)**, drilling down stays snappy, and deep-links resolve. Building it corrected a few assumptions below and made the headline stage **lower-risk and cheaper** than first estimated. POC-derived changes are folded into the relevant sections and tagged **[POC]**.

## Context

The workspace tree model already anticipates lazy loading. [`TreeLeaf.isExpandable`](backend/app/model/frontend/TreeEntry.scala#L85-L93) already means "has children on the server not yet fetched," - but it's hard-coded `false` so workspaces eager-load.  [LazyTreeBrowser.tsx](frontend/src/js/components/viewer/LazyTreeBrowser.tsx) is a complete, working fetch-on-expand implementation for the resource viewer, using the same generic `TreeBrowser` + `isExpandable` + `onExpandLeaf` + flat-merge pattern. 

[...]

PR #745, opened 2026-05-24
Workspace lazy loading (#744) — Stage 2: backend read primitives

hoyla:
> [!NOTE]
> **This PR is part of the #744 workspace-lazy-loading series.** Large workspaces are slow to open because the entire node tree is fetched and walked on every load — for 100k+ item workspaces this can time out (epic #369). #744 seeks to address this by fetching the tree **lazily**, one folder level at a time; the approach was validated end-to-end by a throwaway POC `ljh-lazy-loading-poc`: a 101k-item workspace was displayed in less than two seconds vs ~20s previously. 
> 
> The proposal for #744 is to make a series of manageably sized, independently reviewable PRs. **Nothing changes for users until Stage 5** flips lazy loading on — and even then only for workspaces above a size gate. Normal-sized workspaces keep today's eager path untouched.

### Context: the #744 series

| Stage | Branch / PR | Status | What it does |
|---|---|---|---|
| 1 | #740 | 🟢 merged | Targeted Cypher for search-in-folder (drops a full-tree walk) |
| 2 | `ljh-744-stage2-workspace-read-primitives` | 🔵 **this PR** | Backend read primitives: children / ancestors / aggregate (additive) |
| 3 | `ljh-744-stage3-partial-tree-plumbing` | ⚪ next | Frontend partial-tree store + merge plumbing (no behaviour change) |
| 3.5 | `ljh-744-stage3-5-root-marker` | ⏸ planned (after #742 (🟢  merged) + this PR merge) | `rootNodeId` marker on the Workspace node → O(1) root load; absent-marker fallback |
| 4 | `ljh-744-stage4-incremental-mutations` | ⚪ upcoming | Mutations refresh only the affected parent, not the whole tree |
| 5 | `ljh-744-stage5-lazy-size-gate` | ⚪ upcoming | **Flip to lazy** behind the size gate — the headline change (needs #742). Completes Phase I (the MVP) |
| 6 | `ljh-744-stage6-deep-linking` | ⚪ upcoming | Deep-link reveal via the ancestors endpoint |
| 7 | `ljh-744-stage7-polling` | ⚪ upcoming | Processing-status polling under a partial tree. Completes Phase II (near parity with eager loading) |
| 8–9 | Phase III | ⏸ later | Restore per-folder rollup counts; background backfill. Effectively full parity with eager loading  |
| 10 | Phase IV | ⏸ later | In-folder pagination for huge flat folders |

**Branch strategy:** stages are stacked (each based on the previous) and rebased onto `main` as earlier stages merge. This PR is based on `main`.

## Changes in this PR — Stage 2: backend read primitives

Adds the three additive read endpoints lazy loading will consume. **NOTHING CALLS THEM YET**, so this changes no behaviour. It's just new API endpoints, parity-testable in isolation.

| Endpoint | Route | Returns |
|---|---|---|
| Root + depth-1 children | `GET /api/workspaces/:id/rootChildren` | the root node with its direct children |
| A node's children | `GET /api/workspaces/:id/nodes/:nodeId/children` | the node with its direct children |
| Ancestor path | `GET /api/workspaces/:id/nodes/:nodeId/ancestors` | root→parent, each populated with its children (one round-trip) |
| Workspace totals | `GET /api/workspaces/:id/aggregate` | file / folder / processing-task / failed counts |

- Child folders come back as **lazy `TreeNode`s** (empty `children`) so they keep behaving structurally like folders (drop target, context menu, stable sort); child files carry processing status inline (a per-folder-scoped lookup).
- The four aggregate counts mirror the eager tree's root-node rollup, so the two paths can be checked for **parity** — there's an integration test asserting exactly that.
- All endpoints enforce the existing workspace access rule and return 404 for a missing node/workspace.

### Specific decisions that might be worth noting

- **The children query is split so a per-folder fetch is an index seek.** It was originally one `parent.id = $nodeId OR NOT (parent)-[:PARENT]->()`. `PROFILE` on `neo4j:2026.02.2` showed that `OR` stops the planner using the `WorkspaceNode.id` index, so *every* fetch expanded all `PART_OF` nodes — a single-folder fetch was ~1,600 db hits on an 800-node workspace. Split, a known node is an index seek (~9 db hits) and the root is matched by "has no `PARENT`".
- **The first-paint root load is still O(total)** — finding the parentless root needs a NOT-EXISTS scan, which can't use an index. Consciously deferred (~2s on a 101k-item workspace, vs the eager path it replaces) with a code comment. **Stage 3.5** (above) adds a `rootNodeId` marker on the Workspace node, surfaced beside `nodeCount`, to make this an index seek too.

 - [x] Integration tests: `backend/Test/compile` green; the new Cypher (children / ancestors / aggregate) validated against the pinned `neo4j:2026.02.2` engine on a fixture graph: correct shapes, count **parity** with the eager rollup (including non-zero processing/failed counts), and 404s on missing/inaccessible nodes. New integration tests in `WorkspacesITest` **passed on CI** (they can't run on a recent local Docker Desktop — Engine 29's minimum Docker API 1.40 rejects testcontainers' docker-java v1.32 calls).
 - [x] Playground: regression testing only. Additive endpoints have no consumers yet, so there's nothing to exercise via the UI. That will come at Stage 4 onward.


backend/app/model/annotations/Workspace.scala
```diff
@@ -187,6 +187,23 @@ object WorkspaceMetadata {
   }
 }
 
+// Workspace-wide totals computed in a single aggregate query, for lazy loading (issue #744).
+// Under lazy loading the client never holds the whole tree, so it can't roll these up itself —
+// this feeds the summary header (an accurate total at any size) and the "is anything still
+// processing?" polling decision. The four fields mirror the eager tree's root-node rollup
+// (descendantsLeafCount / descendantsNodeCount / descendantsProcessingTaskCount /
+// descendantsFailedCount) so the two paths can be checked for parity.
+case class WorkspaceAggregate(
```
philmcmahon, line 196: WorkspaceAggregateStats maybe?

Issue #751, opened 2026-06-02
Giant: documents with undefined document-level  resource.text aggregate crash the document renderer


hoyla:
Re Mandelson docs vol2 part I and II not rendering, but part II (and all the separated sections) render fine. 

Wot a bot sed:

### What's erroring

The blank page is a client-side React crash, not a data-loading failure (every API call returns 200):
```
at hasTextContent          → frontend/src/js/util/resourceUtils.ts:81
at PreviewSwitcher.render   → frontend/src/js/components/viewer/PreviewSwitcher.js:193

resourceUtils.ts:81:

export function hasTextContent(resource: Resource): boolean {
  return resource.text.contents.trim() !== "";   // ← resource.text is undefined → throws
}
````

`PreviewSwitcher.js:193`calls it unconditionally on every render. `PreviewSwitcher` is mounted by `DocumentFooter`, which `PageViewerOrFallback` renders whenever a resource is present — so it runs on the combined PDF view. There's no error boundary, so the throw unmounts #root entirely → blank page. (I verified #root has 0 children after the crash.)

### Why parts I and II, but not part III
I pulled the actual resource shapes from the live API. Parts I & II come back without a text field at all (text, ocr, transcript all undefined, previewStatus: "disabled"); part III has text. So:

- Part III → resource.text exists → hasTextContent returns cleanly → renders.
- Parts I & II → resource.text is undefined → .contents throws.

The root flaw is a type lie: `Resource.ts:74` declares `text: HighlightableText `(non-optional), but the API genuinely omits it. That's why `hasTextContent` was written with no null-check and still compiled. Tellingly, `getDefaultView` in the same file does guard `if (resource.text === undefined) (resourceUtils.ts:96)` — the author knew text can be missing; `hasTextContent` just never got the same treatment.

## Why?

Because we force documents more than 500 pages long to a BasicResource, to lower the burden on downstream devices. 

See #750

PR #752, opened 2026-06-02
Guard the viewer against resources with no extracted text

hoyla:
Addresses the crash that surfaced #750 and #751

Make `Resource.text` optional and guard the viewer call sites that dereference it, so the document viewer no longer crashes to a blank page when the API returns a resource without document-level extracted text.

## Why

The API can legitimately return a resource with no `text` field — e.g. when a document is returned as a `BasicResource` rather than a full document (large documents above the page-count gate, or the content-length fallback). However the frontend type declared `text: HighlightableText` as always-present, so `hasTextContent` was written as:

```ts
return resource.text.contents.trim() !== "";
```

When `resource.text` is `undefined`, this throws `TypeError: Cannot read properties of undefined (reading 'contents')` inside `PreviewSwitcher.render`. With no error boundary, the whole viewer unmounts and the page renders blank.

## Changes

- `Resource.text` is now optional (`text?: HighlightableText`), matching what the API actually returns.
- `hasTextContent` and the other call sites that dereferenced `resource.text` (`definitelyNotAUnifiedViewer`, the table view in `PageViewerOrFallback` and `Viewer`, and `Viewer`'s extracted-text branch) are guarded against the undefined case.
- Added a regression test for `hasTextContent` with undefined `text`.

## Behaviour change

The only behavioural difference is in cases that previously crashed: the legacy `Viewer` fallback now degrades gracefully (children tree / "cannot display this document") instead of throwing. The standard page-by-page combined view is unaffected — it never reads `resource.text`.

## Testing

- [x] Local dev - manually verified: a document returned without text now renders via the combined page view instead of blanking.
- [ ]  Preview

### DIFF
--- a/frontend/src/js/components/PageViewerOrFallback.tsx
+++ b/frontend/src/js/components/PageViewerOrFallback.tsx
@@ -97,6 +97,9 @@ const PageViewerContent: FC<{
   }
 
   if (view === "table") {
+    if (!resource.text) {
+      return renderNoPreview();
+    }
     return <TablePreview text={resource.text.contents} />;
   } else if (view === "preview") {
     return <Preview resource={resource} />;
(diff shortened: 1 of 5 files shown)


philmcmahon (approved):
nothing to see here

PR #757, opened 2026-06-07
Fix workspace tree keyboard navigation and scroll behaviour

hoyla:
This addresses three related bugs in the workspace tree view.

### 1. Folder re-collapses instead of opening the child (after collapse + re-expand)
`Node` accumulated child component refs by *pushing* in the ref callback but never resetting the array — unlike `TreeBrowser/index.tsx`, which resets it each render. When a child row unmounts on collapse its stale, detached ref stayed at index 0; on re-expand, CursorDown focused that detached row (a no-op), leaving DOM focus on the folder, so the next Return keystroke re-collapsed the parent folder instead of acting on the "selected" item. Whereas a direct mouse click focuses the live row properly, so Return opens it, as you'd expect. 

Fix: reset `childReactComponents` each render in `Node`.

### 2. Selected row scrolls up behind the toolbar during keyboard nav
The row key handlers moved the selection on arrow keys but never called `preventDefault`, so the browser *also* ran its native cursor scroll. That extra scroll fought the focus-driven `scrollIntoView` and pushed the selected row up out of view behind the "New Folder / last refreshed" toolbar. 

Fix: cancel the default for arrow keys in `Node` and `Leaf`, making both selection change and scroll behaviour more like that seen in the MacOS Finder.

### 3. Bottom of the scroll viewport sat off-screen
`.workspace__tree` used `height: 100%` inside a non-flex `.workspace`, so it resolved to the full workspace height and ignored the 50px toolbar above it, pushing the bottom of the scroll area off-screen. 

Fix: make `.workspace` a flex column and let the tree flex into the remaining space. (Separate, smaller issue from #2 — kept here as it's adjacent.)

Before:

before

After:

after

## Testing
- [x] Local dev.
- Monsieur Claude tells me that `keyboardNavigation.spec.tsx` has the repo's first component-mount tests (hand-rolled `ReactDOM.render` + `act` + `Simulate`, as there's no testing-library) — flagging in case this pattern is not acceptable. We could just remove the test.

### DIFF
--- a/frontend/src/js/components/UtilComponents/TreeBrowser/Leaf.tsx
+++ b/frontend/src/js/components/UtilComponents/TreeBrowser/Leaf.tsx
@@ -62,6 +62,14 @@ export default class Leaf<T> extends React.Component<Props<T>, {}> {
 
   onKeyDown = (e: React.KeyboardEvent<HTMLTableRowElement>) => {
     e.stopPropagation();
+
+    // The arrow keys drive selection and expansion. Prevent the browser's
+    // native arrow-scroll, which otherwise fights the focus-driven scroll and
+    // shoves the selected row up out of view behind the toolbar.
+    if (e.key.startsWith("Arrow")) {
+      e.preventDefault();
+    }
+
     switch (e.key) {
       case "Enter":
         this.props.onSelect(this.props.entry);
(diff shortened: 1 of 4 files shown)


philmcmahon (approved):
nice 👍🏻 

I am a bit unsure as to the benefit of the test (as we are unlikely to change this logic much) but think it's probably better having it than not

Issue #733, opened 2026-05-11
Discuss why Giant search results push a view mode


hoyla:
Search results show results in each 'view mode' of a document - so for example there could be different results in the Text and OCR views. Consequently search results point the user to a particular view (I imagine whichever comes first if there are multiple). This used to be very useful when there was no such thing as the Combined document view, because you'd be taken to where the actual search match is, rather than simply to the document that has them. 

It still makes sense in these circumstances:
- The match is not available in Combined view (i.e. it can only be found in another view)
- There is no Combined view for that document

But it doesn't make sense when the match can be found in multiple views, of which one is Combined. 

Does elasticsearch even know there's a combined view? Which view mode should be picked when there is one (and it features in the matches for that document), and should it be up to search to decide that, or for the "document" to decide it on open (as happens when you open things from a dataset or workspace).

PR #761, opened 2026-06-07
Take search results to the Combined view when the match is there

### LINKED ISSUE #760: In Giant, no search ever takes you to the Combined view
When you open a document from an ingestion or workspace, the view mode is not baked in: it chooses the preferred view. If there's a "Combined" view available (because the doc has pages), that will be chosen. 

If however you open a document from a URL that specifies the view mode, that will be honoured. 

And that's what happens with search results. When you search in Giant, elasticsearch delivers matches in each field for a given document: text, OCR, transcript. The search results pane shows the top matches for the document in each field that contains a match.  In the example below there are three for text and two for OCR:

Image

When you click on a search result, the view with the most matches is appended to the end of the URL, e.g. "?view=text" (plus the search query string). So the document opens in that view, guaranteeing that you can land on an actual match in the document.

The problem here is that the Combined view is a frontend concept - elasticsearch doesn't even know it exists. So it never finds anything in "the Combined view" (there's no such field), and therefore never proposes "?view=combined" to the document viewer. 

Consequently search results never land on the view that we actually want people to use.

hoyla:
**This change means that when you click a Giant search result for a paged document, it opens in the Combined view with the match highlighted — provided the match is actually reachable in Combined.** Previously all search results took users to the Text or OCR view even when Combined was available. 

Fixes #760. Background discussion in #733. Most of the the 350 lines are tests and comments rather than actual code changes. 

### Why

Elasticsearch has no concept of the "Combined" view — it's assembled in the frontend from the page index. Search picks the document field with the most highlights (`text`, `ocr.english`, etc) and forces that view via `?view=`, which suppressed the Combined default and dropped users into flat text/OCR even when Combined would show the same match in the view we actually want people to use.

### How

The viewer needs to know, before choosing a view, whether the search query is reachable in the page index. Elasticsearch can't say during search itself, so the answer is folded into a request the viewer already makes:

**Backend.** `GET /api/pages2/:uri/pageCount` accepts an optional `q` and returns `searchMatchesInPages` alongside the count — `true`/`false` when a query was sent, `null` otherwise. It is answered by `Pages2.hasSearchMatch`: a single Elasticsearch count query built with the same query builder as the page highlighters, so quoting and stemming semantics match exactly what the Combined view will highlight. A failed or unparseable query recovers to "no match" — opening a document can never break because of `q`. (Deliberately *not* answered via the existing `/pages2/:uri/search` endpoint, which fans out per hit page — permission check, ES multi-search, S3 PDF download, PDFBox geometry extraction — far too heavy to use as a boolean.)

**Frontend.** All decision inputs arrive in that one response, so the view is decided exactly once, inside the pageCount fetch callback, by a pure unit-tested helper (`resolveInitialPagedView`):

- paged document, no view → Combined (the existing default, e.g. opening from a workspace)
- paged document, search-forced text/OCR view, match reachable in the pages → Combined
- anything else → honour the current view

Nothing re-fires when the view changes, so a user switching to the Text/OCR tab afterwards is never overridden. The switch is dispatched before the page count is stored, so no frame ever renders in the view being left. It goes through `setResourceView`, which the middleware syncs to the URL via `replace` — no back-button trap; after a search click the URL settles on `?view=combined&q=…`.

The match-driven rule self-heals the cases where page text is *not* a superset of the document fields — lossy Tesseract-era OCR, cross-field matches, any future translation view: those answer "no match", so the search-forced flat view is kept and the user still lands on a real match.

### Behaviour

| Open via | Result |
|---|---|
| Search hit in body of a **paged** PDF (`?view=text`/`ocr`) | **Combined**, match highlighted |
| Search hit only reachable outside the pages (lossy OCR / cross-field / future translation) | forced flat view, as before |
| Search hit in a **non-paged** doc (email, plain text, image-only) | forced view, as before |
| Opened from workspace/dataset (no `?view=`) | Combined default, as before |
| User clicks Text tab after landing on Combined | stays on Text |

### Robustness

- `PageViewerOrFallback` is **not remounted** when the `/viewer/:uri` route param changes, so the cached page count is tagged with the uri it was fetched for and a mismatched count reads as "not yet known". Without this, stepping to the next/previous result imposed Combined on a non-paged document using the *previous* document's count, and 404'd.
- The pageCount fetch has a cancellation guard: a late response from a superseded request (stepping quickly between results) can neither replace the current document's data nor dispatch a view change after the document has been left.
- `useHighlightNavigation` catches fetch failures, so a pages 404 can never surface as an uncaught runtime error.
- Deploy skew: a backend without the parameter simply omits the field, which resolves to keeping the search-forced view — the pre-existing behaviour.

### A note on shared URLs

A shared `?view=text` with no query is honoured. A shared `?view=text&q=…` resolves by match reachability — it's indistinguishable from a fresh search click. This is an accepted trade-off: a `view+q` URL never preserved the sharer's intent anyway (search navigation always focuses the *first* match, so it lands on match number 1 in any view). Directing a colleague to a specific passage is better addressed by snippet/selection sharing, which encodes an anchor rather than re-running a query.

### OCR engines

Under the **OcrMyPdf** engine, page text retains the born-digital layer and adds OCR, so a `text`/`ocr` match is almost always also in the pages → answer "yes". Under the **Tesseract** engine, page text is a lossy OCR re-derivation, so a born-digital `text` match can be missing from the pages → answer "no" → the forced flat view is kept. The fix is correct under either engine; the per-document answer is what makes it robust.

### Testing

- `resolveInitialPagedView`: 11 unit tests, green.
- `ElasticsearchPagesITest`: 10/10 green, including 5 new `hasSearchMatch` cases (match, no match, quoted phrases, cross-document isolation, no pages). Local caveat: this branch predates the #755 testcontainers bump, so against Docker Engine 29+ the suite only runs with that bump applied (verified that way locally); CI runs it as-is.
- Full frontend suite 222/222; `CI=true npm run build` typechecks clean.
- Before removing WIP: a browser-level end-to-end pass of the final wiring (search click → Combined; stepping across paged/non-paged results).

### Follow-up (out of scope)

Stepping next/previous between search results navigates without a `?view=` (deliberately dropped by `buildLink`), so it always lands on the Combined default without consulting match reachability. Unifying it with the click path would mean passing the stepped-to result's `fieldWithMostHighlights` (already available on `state.search.currentResults`) through `DocumentFooter`.

### DIFF
--- a/backend/app/controllers/api/PagesController.scala
+++ b/backend/app/controllers/api/PagesController.scala
@@ -19,17 +19,32 @@ import utils.controller.{AuthApiController, AuthControllerComponents}
 class PagesController(val controllerComponents: AuthControllerComponents, manifest: Manifest,
     index: Index, pagesService: Pages2, annotations: Annotations, previewStorage: ObjectStorage) extends AuthApiController {
 
-  def getPageCount(uri: Uri) = ApiAction.attempt { req =>
+  def getPageCount(uri: Uri, q: Option[String]) = ApiAction.attempt { req =>
     val countAttempt = pagesService.getPageCount(uri)
     val dimensionsAttempt = pagesService.getFirstPageDimensions(uri)
       .recoverWith { case _ => Attempt.Right(None) }
+    // Answers "would the Combined view show a match for this search?" alongside
+    // the count, so the frontend can choose the initial view for a document
+    // opened from search results in a single round trip (issues #733 / #760).
+    // A failure here (e.g. a query elasticsearch can't parse) must not stop the
+    // document opening: it reports as "no match", which keeps the search-forced
+    // flat view.
+    val searchMatchAttempt = Attempt.sequenceOption(
+      q.filter(_.trim.nonEmpty).map { query =>
+        Attempt.catchNonFatalBlasé(Chips.parseQueryString(query).query)
+          .flatMap(parsed => pagesService.hasSearchMatch(uri, parsed))
+          .recoverWith { case _ => Attempt.Right(false) }
+      }
+    )
     for {
       count <- countAttempt
       dimensions <- dimensionsAttempt
+      searchMatchesInPages <- searchMatchAttempt
     } yield {
       Ok(Json.obj(
         "pageCount" -> count,
-        "dimensions" -> dimensions
+        "dimensions" -> dimensions,
+        "searchMatchesInPages" -> searchMatchesInPages
       ))
     }
   }
--- a/backend/app/services/index/Pages2.scala
+++ b/backend/app/services/index/Pages2.scala
@@ -151,6 +151,23 @@ class Pages2(val client: ElasticClient, indexNamePrefix: String)(implicit val ex
     }
   }
 
+  // Cheap existence check: does the query match anywhere in this document's
+  // pages? A single count query — unlike findInPages + the geometry pipeline
+  // behind findInDocument/searchInDocument, this never touches S3 or computes
+  // highlight geometry. Used to decide the initial view for a document opened
+  // from search results (issues #733 / #760).
+  def hasSearchMatch(uri: Uri, query: String): Attempt[Boolean] = {
+    execute {
+      count(textIndexName).query(
+        must(buildQuery(query)).filter(
+          termQuery(PagesFields.resourceId, uri.value)
+        )
+      )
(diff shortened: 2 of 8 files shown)


philmcmahon (approved):
combine

PR #764, opened 2026-06-07
Add 'Download folder as text' to the workspace tree context menu

hoyla:
## What & why

Builds on the new feature https://github.com/guardian/giant/pull/691 that allows us to take a LLM-friendly compiled text export of the content of a workspace - and, potentially, integrate that with our own LLM services. 

Adds an admin-only **"Download folder as text"** action to the workspace-tree context menu, scoped to a selected folder and its tree descendants. It's the first stage of generalising the existing workspace-wide "Download all text" to folders, reusing that feature's machinery plus the word-count aggregation already used for workspaces.

This is **Stage 1 (tree nodes only)** — self-contained and shippable, with **no behaviour change to the existing workspace download**.

## How it works

- **Frontend** — `DownloadTextModal` now takes an optional `rootEntry`/`label`, so it can be scoped to a folder subtree. It builds full workspace-root-relative paths (for provenance in the `----- START OF … -----` headers) and restricts them to the folder's blobs. A new context-menu item (gated on `isAdmin && isWorkspaceFolder && !isRemoteIngest`, alongside "Reprocess folder contents…") opens the modal scoped to the folder.
- **Backend** — folder word counts go via a new `POST /api/workspaces/:id/wordCount` (`getWordCountForBlobs`), which reuses the **same** Elasticsearch word-count aggregation as the whole-workspace figure, swapping the workspace-tag filter for an ids filter (plus the workspace filter, mirroring `getTextForBlobs`'s security property). The aggregation is extracted into a shared helper so the two counts are computed identically and therefore reconcile.

## Consistency note

The only divergence risk was the displayed word count. Because folder and workspace counts now use the identical aggregation, the sum of top-level folder counts ≈ the workspace total (modulo any ES-tag↔tree drift). The thing that *would* break reconciliation — attachment/archive sub-documents (`parentBlobs`) — is deliberately out of scope here.

## Deferred to Stage 2 (rides the lazy-loading work)

Bundled because both need server-side URI derivation:
- Move folder **and** workspace URI derivation server-side via `getBlobUrisInWorkspaceFolder` (fixes lazy-loading fragility for both).
- Include `parentBlobs` sub-documents using folder-search's `buildWorkspaceBlobFilter` authorisation model.
- Re-point the workspace count/download to the same derivation so reconciliation holds with attachments included. **Heads-up for that PR:** the existing "Download all text" will then start including attachment/archive text it currently omits.

## Testing

- New ITests for `getWordCountForBlobs`: counts words for in-workspace blobs, returns 0 for blobs outside the workspace (no cross-workspace probing), and prefers OCR over extracted text. (ITests run in CI.)
- `react-scripts build`, backend `compile`, and `Test/compile` all green.
- **Manually verified** in the running app as admin: folder menu item appears; modal scopes to the folder with a count from the new endpoint; downloads contain only the folder's files; existing workspace-wide download unchanged.

### DIFF
--- a/backend/app/services/index/ElasticsearchResources.scala
+++ b/backend/app/services/index/ElasticsearchResources.scala
@@ -617,9 +617,10 @@ class ElasticsearchResources(override val client: ElasticClient, indexName: Stri
     )
   }
 
-  // Counts whitespace-separated words across the text (or, where present, OCR) of every language for
-  // every document matched by the query. Shared by the workspace- and folder-scoped word counts so they
-  // always use the same counting method (only the set of matched documents differs).
+  // Counts whitespace-separated words in every language of a single field per document, picked in
+  // precedence order: OCR where present, otherwise extracted text, otherwise the transcript (so
+  // audio/video documents are counted too). Shared by the workspace- and folder-scoped word counts so
+  // they always use the same counting method (only the set of matched documents differs).
   private val wordCountAggregation =
     scriptedMetricAggregation("workspaceWordCount")
       .initScript(Script("state.total = 0L").lang("painless"))
--- a/backend/app/services/index/ElasticsearchResources.scala
+++ b/backend/app/services/index/ElasticsearchResources.scala
@@ -7,6 +7,7 @@ import com.sksamuel.elastic4s.requests.common.FetchSourceContext
 import com.sksamuel.elastic4s.requests.script.Script
 import com.sksamuel.elastic4s.requests.searches.DateHistogramInterval
 import com.sksamuel.elastic4s.requests.searches.SearchResponse
+import com.sksamuel.elastic4s.requests.searches.queries.Query
 import com.sksamuel.elastic4s.requests.searches.queries.compound.BoolQuery
 import com.sksamuel.elastic4s.requests.update.UpdateByQueryRequest
 import extraction.{EnrichedMetadata}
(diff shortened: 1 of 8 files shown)


twrichards:
Please can you add a reference to https://github.com/guardian/giant/pull/691 in the PR description to improve tracing between related PRs

backend/app/services/index/ElasticsearchResources.scala
```diff
@@ -616,37 +617,52 @@ class ElasticsearchResources(override val client: ElasticClient, indexName: Stri
...
-                              |  }
-                              |}
-          """.stripMargin.trim).lang("painless"))
-          .combineScript(Script("return state.total").lang("painless"))
-          .reduceScript(Script("long total = 0; for (s in states) { total += s } return total").lang("painless"))
-      )
+  // Counts whitespace-separated words across the text (or, where present, OCR) of every language for
+  // every document matched by the query. Shared by the workspace- and folder-scoped word counts so they
+  // always use the same counting method (only the set of matched documents differs).
+  private val wordCountAggregation =
```
philmcmahon, line 624: it looks like this will count the words of ocr OR text OR transcript, which makes sense I think. Someday we may want to allow people to download translated text as well

philmcmahon: maybe mention the transcript fallback in the comment above?

philmcmahon (approved):
LGTM 

folder copy

PR #768, opened 2026-06-09
also include transcribed text in the 'download workspace as text files' feature

twrichards:
https://github.com/guardian/giant/pull/691 introduced a new (admin only) option to download all the text of a workspace, but it didn't included transcribed text... this PR fixes that.

There are quite a few other fields it should probably include... but they need more thought and I'd like get this improvement released.

It would also be nice to have links to giant in the resulting text files, for easier navigation back to the original, but again that can follow.

### DIFF
--- a/backend/app/services/index/ElasticsearchResources.scala
+++ b/backend/app/services/index/ElasticsearchResources.scala
@@ -529,7 +529,8 @@ class ElasticsearchResources(override val client: ElasticClient, indexName: Stri
           .initScript(Script("state.total = 0L").lang("painless"))
           .mapScript(Script("""
                               |if (params._source != null) {
-                              |  Map text = params._source.ocr != null ? params._source.ocr : params._source.text;
+                              |  Map text = params._source.ocr != null ? params._source.ocr :
+                              |     (params._source.text != null ? params._source.text : params._source.transcript);
                               |  if (text != null) {
                               |    for (entry in text.entrySet()) {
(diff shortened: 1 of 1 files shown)


backend/app/services/index/ElasticsearchResources.scala
```diff
@@ -561,13 +562,17 @@ class ElasticsearchResources(override val client: ElasticClient, indexName: Stri
       )
       .sourceInclude(
         "text.*",
-        "ocr.*"
+        "ocr.*",
+        "transcript.*"
```
philmcmahon, line 566: I'm not sure but it might be worth using the vttTranscript format (which has timecodes) rather than just the transcript to help people find that bit of audio https://giant.pfi.gutools.co.uk/viewer/XJ5tDUKttEeh03mC4oKVLdOsj5-mx1zGZd0x_v60xlQ3aJRckFdERA1_A2R8iBTeZ9kvPMAeMPTmfYWiEatqRw?navId=18d3ccbb-d4f1-4a00-8980-6d13247dddce&navIndex=3&view=vttTranscript.english

twrichards: I did think about that, but wasn't sure if that's gonna throw LLM off - I expect user to select some of the text that's informed an LLM answer and search for it in the original file - so I'm going to leave it as is for now.

philmcmahon (approved):

PR #770, opened 2026-06-12
workspace level progress bar show overall amount processed

twrichards:
I recently had need to wait for processing to complete on a workspace with a few thousand items (all top level) and it was a bit annoying having to scroll up/down and eyeball roughly how many were processed to see roughly how close/long it would be until I could do something (which needed everything to be processed), this would be actually be near impossible on a much bigger workspace.

Seemed straight-forward to add a small progress bar (like was added for moving files - see #460) to display this, needed an additional server-side computed field to count the actual files in processing state (as we only had the task count, which can be more than one per file), but this was simple to add.

image


philmcmahon (approved):
Nice!

PR #776, opened 2026-06-29
Fix reprocess  following Neo4j upgrade

philmcmahon:
## What does this change?
There are various problems with reprocessing but right now it appears to be completely broken. I think this was probably caused by the neo4j upgrade, but because reprocessing is so flaky anyway we might have ignored failures during testing.

This change fixes the top level error. There are other reprocessing related problems which will need a bit more work.

The error this fixes looks like this - neo4j even tells us how to fix it:

```
Caused by: org.neo4j.driver.exceptions.Neo4jException: 22N27: Invalid input 'RELATIONSHIP' for processedRelation. Expected to be MAP. Hint: use properties(...) on the right-hand side.
```

 - [x] Tested locally
 - [ ] Tested on playground


srbd (approved):
good spot

PR #748, opened 2026-06-01
Replacing minio for Garage for local development 

srbd:
## What does this change?

Minio has gone [out of support](https://github.com/minio/minio?tab=readme-ov-file). [Garage](https://garagehq.deuxfleurs.fr/) seems to be a well-funded alternative.  

It's mostly a drop-in replacement although:
- the initiatlisation is done a bit differently (we've done it in a separate script) 
- you have to setup a ui (we've chosen [garage ui](https://github.com/Noooste/garage-ui) as it seems well-maintained) separately

## How has this change been tested?
 - [x] Tested locally

PR #778, opened 2026-06-30
Fix presigned urls when using garage locally

philmcmahon:
## What does this change?
In https://github.com/guardian/giant/pull/748 we forgot to test presigned urls locally (only used for transcription/translation).

This makes sure we can generate presigned urls when using garage.

 - [x] Tested locally
 - [ ] Tested on playground


srbd (commented):
Oops thanks for the fix

srbd (approved):

PR #784, opened 2026-07-16
Local DEV setup improvements

philmcmahon:
## What does this change?
I set up giant on my laptop and discovered:
 - wkhtmltopdf isn't supported on macos and the brew thing doesn't exist. It's still there for linux. We use this to provide a preview of HTML files uploaded to giant
 - We need to run the postgres migrations after setting up giant for the first time otherwise there will be lots of errors and some things (like blob deletion) won't work at all. We obviously can't run the migrations without starting docker so I've pulled the docker stuff out into a separate script
 - We need to install tesseract-lang to support multilingual OCR with tesseract

 - [x] Tested locally
 - [ ] Tested on playground


twrichards (commented):
LGTM - thanks

twrichards (approved):

PR #785, opened 2026-07-16
Add translation search highlights

philmcmahon:
## What does this change?
This PR wires in the various bits and pieces needed to show highlights for search results from inside translated text.

As part of this work I've ended up abandoning the dropdown switcher for viewing translations introduced in https://github.com/guardian/giant/pull/780 and instead made the translation views just more tabs in the bottom right. I think the UX here could use some work (as you can get a lot of tabs) but didn't want to put time into dealing with highlights and url query syncing for the translation viewer as part of this bit of work, instead putting it off for some future UX overhaul (hopefully post full typescriptification).

Depends on ~https://github.com/guardian/giant/pull/783~ https://github.com/guardian/giant/pull/787 which should be merged and rebased into this before this is reviewed - it was too awkward working with a js PreviewSwitcher.

The main changes:
 - new HighlightableLanguageData which has HighlightableText instead of just string types (the standard LanguageData type is used for to/from elasticsearch, HighlightableLanguageData is constructed based off search results from elastic
 - New 'translated body text' chip - we currently don't have an OCR text chip so I just did body text for now. (Pending @hoyla's mega chips overhaul)
 - Various helper functions for digging out the heavily nested `languageData.text.translation` field
 - I updated the `getSupportedLanguages` endpoint so that it returns all language data so that we can match up language codes to language names in the client (to show on the tabs in the bottom right)

 - [x] Tested locally
 - [x] Tested on playground


philmcmahon:
Note- since the above gif this has been updated to show the actual language names on the tabs rather than 'Text Face'

backend/app/services/index/ElasticsearchResources.scala
```diff
@@ -486,7 +486,14 @@ class ElasticsearchResources(override val client: ElasticClient, indexName: Stri
...
           )).filter(SearchContext.buildFilters(parameters, context))
         )
-        .fetchContext(FetchSourceContext(fetchSource = true, excludes = Set(IndexFields.text, s"${IndexFields.ocr}.*")))
+        .fetchContext(FetchSourceContext(fetchSource = true, excludes = Set(
+          // exclude massive fields from source context to avoid errors when fetching large documents
+          IndexFields.text,
+          s"${IndexFields.ocr}.*",
+          TranslationIndexFields.text,
+          TranslationIndexFields.ocr,
+          TranslationIndexFields.emailBody
```
twrichards, line 495: I'm curious, does the non-translated email body go into IndexFields.text currently?? if so for emails, what ends up in TranslationIndexFields.text (in that TranslationIndexFields.emailBody is separate)?

twrichards, line 495: also I notice there's mention of translating emailSubject in the TranslationIndexFields class, but that we're not doing search highlights, is that correct??

twrichards (approved):
nice work - couple of comments and I've not tried it out in playground, but trust you have

backend/app/services/index/ElasticsearchResources.scala
philmcmahon: yes that's correct, we currently do language detection for the email subject and body but don't actually translate or do any highlighting on them

philmcmahon: Good question. I just checked this, and, with the current behaviour, both text and emailBody will have exactly the same data in.

I've added an issue documenting this here https://github.com/guardian/giant/issues/825

PR #793, opened 2026-07-28
Fix: delete using path/segment boundaries rather than stems

hoyla:
This is actually a two-line change but I also added a test

There's a bug that we should definitely fix before letting anyone else use CLI. Incidentally, the bug is present in the backend generally - it's not in the CLI per se - but mercifully we currently have no UI to delete ingestions.

`deleteResourceAndDescendants` matched resources with `uri STARTS WITH $uri` — a raw string prefix. This means that if I attempted to delete ingestion `collection/batch1`, I would also inadvertently  delete every resource of any sibling whose name shares the spelling - e.g. `collection/batch10`, `collection/batch11` etc - and all the files therein. Likewise, deleting collection `luke` would destroy `luke_2`. 

The match now requires either the exact URI or a `/` immediately after the prefix, so a deletion can never leak into a sibling.

`deleteBlob`'s descendant clause used the same raw prefix. It was safe in practice only because blob URIs are fixed-length hashes (one can never be a strict prefix of another); it now uses the same explicit boundary rather than relying on that property.

Two new integration tests in `Neo4JManifestITest` (ingestion-sibling and collection-sibling cases). Both **fail against the previous query** — the sibling really is destroyed — and pass with the fix. Backend unit suite unaffected.

There are no index implications: the query still anchors on `uri` prefix matching, which uses the existing Resource uri index the same way.

### DIFF
--- a/backend/app/services/manifest/Neo4jManifest.scala
+++ b/backend/app/services/manifest/Neo4jManifest.scala
@@ -1292,7 +1292,7 @@ class Neo4jManifest(driver: Driver, executionContext: ExecutionContext, queryLog
       |OPTIONAL MATCH (w: WorkspaceNode { uri: $uri})
       |OPTIONAL MATCH (b: Blob:Resource { uri: $uri})-[:PARENT]->(f: File)
       |OPTIONAL MATCH (descendant :Resource)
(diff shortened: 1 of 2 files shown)


twrichards (approved):
excellent catch!

PR #796, opened 2026-07-28
[infra] use SSM param for the postgres VPC

twrichards:
corrects a long standing mistake/oversight in the default value for the `VpcId` CFN parameter on the stack. This is a fairly academic change since it would only apply to any 'new' postgres stacks created, given the CODE & PROD ones appear to have had their SSM path updated (presumably for a long time) 
image

... but its better to get this right!

NOTE how this is abstracted into `GuStackWithGiantVPC`, which may seem premature but is in anticipation of migrating further infra definitions into CDK here


philmcmahon (approved):
✅

PR #800, opened 2026-07-31
FIXes viewer sidebar getting stale when stepping through search results

### LINKED ISSUE #799: Giant doc viewer: Sidebar references can persist from previously viewed documents after revisiting search results
If you perform a search that returns, for example, multiple versions of a document and then
1.  click on one search result to view the document;
2.  **hit back to return to the search results page**;
3.  click on a different result in the results list to the one you selected at stage 1

.. when you open that document, it will display in the main pane but **the sidebar will still show the information pertinent to the document you first opened from the search** 

Image

- The metadata will be wrong
- The location will be wrong and so if you click on part of the path, you'll get to something that doesn't correspond to what you're actually looking at
- Worst of all if you opt to download the file you will in fact download a different file. If, as in the example, those files are similar you may be entirely unaware of this.

twrichards:
When trying to download an original file, having stepped through a number of search results, it offered a download of the first search result I'd clicked on, I then noticed all the other sidebar content was stale too. 

The fix is simple, the guard (if check) around the fetch for the right information checking the OLD state of `props.isLoadingResource` rather than the new, but regardless this check is a bit pointless since the URI check is sufficient, so I've simplified.

Fixes https://github.com/guardian/giant/issues/799

### DIFF
--- a/frontend/src/js/components/viewer/ViewerSidebar.js
+++ b/frontend/src/js/components/viewer/ViewerSidebar.js
@@ -35,18 +35,14 @@ class ViewerSidebar extends React.Component {
   };
 
   UNSAFE_componentWillReceiveProps(props) {
-    if (
-      !this.props.isLoadingResource &&
-      props.match.params.uri !== this.props.match.params.uri
-    ) {
+    if (props.match.params.uri !== this.props.match.params.uri) {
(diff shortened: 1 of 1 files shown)


twrichards:
none of the things I tried on this PR actually fix the problem, so closing this PR for now

PR #801, opened 2026-08-02
Viewer: fetch and render the resource only when it matches the route uri

hoyla:
Re: #799 

`state.resource` in redux can outlive the document it belongs to, and nothing on the paged-document path clears or revalidates it:

1. Opening a **paged** document renders `PageViewer`, which fetches pages by the route uri directly — the legacy `<Viewer>` (whose unmount is what resets the resource) is never mounted. `<ViewerSidebar>` fetches the resource and the store holds it.
2. Hitting **back** to search reuses the cached query and results, so `Search` skips `resetResource()` (it only resets when it actually re-runs a search). The stale resource survives.
3. Opening a **different** result remounts `<ViewerSidebar>`, whose guard was `if (!this.props.resource)` — the stale resource is truthy, so **no fetch happens**. The sidebar, location breadcrumb and the footer's `DownloadButton` (which reads `state.resource`, not the route uri) all act on the previous document, while the main pane correctly fetches the new document's pages by uri.

## Fix

Replace "is a resource present?" with "is the resource **the one for this route**?" via a new `isResourceForUri` helper. It normalises percent-encoding on both sides, since `resource.uri` holds the encoded form while route params can arrive either way (see the long-standing comment on `getCurrentResource`).

- `<ViewerSidebar>` and the legacy `<Viewer>` fetch on mount whenever the stored resource doesn't match the route, and render nothing rather than a stale resource while the right one loads.
- `<PageViewerOrFallback>` treats a non-matching resource as absent, so the footer — and its download button — can never act on the wrong document. This mirrors the uri-tagging pattern already used there for the `pageCount` response.
- The sidebar's uri-change fetch no longer skips while a previous fetch is in flight, which could otherwise pin the in-flight document to the wrong route when stepping quickly between results.

A uri-match check was chosen over dispatching `resetResource()` on unmount because it also covers a stale resource written by the `getBasicResource` paths (`Directory`, `CurrentCollection`), and is immune to an in-flight response landing after a reset.

### DIFF
--- a/frontend/src/js/components/PageViewerOrFallback.tsx
+++ b/frontend/src/js/components/PageViewerOrFallback.tsx
@@ -35,6 +35,7 @@ import {
   COMBINED_VIEW,
   resolveInitialPagedView,
 } from "./PageViewer/resolveInitialPagedView";
+import { isResourceForUri } from "../util/resourceUtils";
 
 function isCombinedOrUnset(view: string | undefined): boolean {
   return !view || view === COMBINED_VIEW;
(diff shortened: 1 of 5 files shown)


philmcmahon (approved):
Noice

PR #806, opened 2026-08-12
Simplify translation for OCR'd files. Replace languageData with translationData.

### LINKED ISSUE #789: Translation index mapping doesn't make sense
In https://github.com/guardian/giant/pull/780/changes#diff-7313b78f0596187098ad28c728d42da778992c1d403d133d5ebbfd45b39deaa2 Phil introduced a new field ‘translation’ inside the languageData block.

The languageData block looks like this:

```json
...
languageData: {
  text: {
    detectedLanguageCode: "fr",
    translation: "good day",
  },
  ocr: {
     detectedLanguageCode: {
      english: "fr",
      french: "fr",
    },
    translation: {
       english: "good day",
       french: "good day",
    }
}
```

There's some detail on this in the original language detection PR here https://github.com/guardian/giant/pull/725

We use an explicit mapping for elasticsearch, which means that each time we add new fields we have to update the mapping. On startup, giant does some of this automatically. For example, if a new language is added to Languages.scala then a mapping for that language will be added to all multi language fields.

The current mapping for the languageData block is shown below. There are two key mistakes with it:
 - the emailBody, emailSubject and text fields don't have `term_vector` set, so we can't do highlighting on the contents of those fields
 - in the ocr field, `detectedLanguageCode` and `translation` are full blown multi language fields - with different analyzer depending on the language. This is stupid as detectedLanguageCode doesn't need indexing at all, and definitely not with different analyzers as it's just a 2 character code in the latin alphabet. It's also stupid for `translation` because that will always be in english.

The problems with this are:
 - we can't do highlighting at all on the text/email fields
 - ocr indexing is overly complex and wasteful

```json
"languageData": {
          "properties": {
            "emailBody": {
              "properties": {
                "detectedLanguageCode": {
                  "type": "text"
                },
                "translation": {
                  "type": "text"
                }
              }
            },
            "emailSubject": {
              "properties": {
                "detectedLanguageCode": {
                  "type": "text"
                },
                "translation": {
                  "type": "text"
                }
              }
            },
            "ocr": {
              "properties": {
                "detectedLanguageCode": {
                  "properties": {
                    "arabic": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "arabic"
                    },
                    "english": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "english"
                    },
                    "french": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "french"
                    },
                    "german": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "german"
                    },
                    "persian": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "persian"
                    },
                    "portuguese": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "portuguese"
                    },
                    "russian": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "russian"
                    },
                    "spanish": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "spanish"
                    }
                  }
                },
                "translation": {
                  "properties": {
                    "arabic": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "arabic"
                    },
                    "english": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "english"
                    },
                    "french": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "french"
                    },
                    "german": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "german"
                    },
                    "persian": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "persian"
                    },
                    "portuguese": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "portuguese"
                    },
                    "russian": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "russian"
                    },
                    "spanish": {
                      "type": "text",
                      "fields": {
                        "exact": {
                          "type": "text",
                          "term_vector": "with_positions_offsets",
                          "analyzer": "standard"
                        }
                      },
                      "term_vector": "with_positions_offsets",
                      "analyzer": "spanish"
                    }
                  }
                }
              }
            },
            "text": {
              "properties": {
                "detectedLanguageCode": {
                  "type": "text"
                },
                "translation": {
                  "type": "text"
                }
              }
            }
          }
        },
```

## How to fix this
In https://github.com/guardian/giant/pull/785 I attempt to fix this by updating ElasticsearchSyntax.scala here https://github.com/guardian/giant/pull/785/changes#diff-c0dcd79765470ddbdccc49538f2ed970f248e64c675c1c5a4b55216dba6667fb so that `languageData.text` gets a proper index that can be used for highlights. However, I ran into problems when trying to apply the mapping update on playground - you can't do this:

```bash
curl -X PUT "http://localhost:19200/pfi/_mapping" \
  -H 'Content-Type: application/json' \
  -d '{
  "properties": {
    "languageData": {
      "properties": {
        "text": {
          "properties": {
            "translation": {
              "type": "text",
              "fields": {
                "exact": {
                  "type": "text",
                  "analyzer": "standard",
                  "term_vector": "with_positions_offsets"
                }
              },
               "term_vector": "with_positions_offsets",
              "analyzer": "english"
            }
          }
        }
      }
    }
  }
}'
```

So what to do instead? We can either make a new index and swap over to it (sounds hard) or maybe just abandon languageData.text.translation and introduce languageData.text.translation2 - that's what I did here https://github.com/guardian/giant/pull/785/changes/95818854aa3e817ee7236142d22ae0945ba86b89 - and I think that would have worked _except_ that it would introduce a translation2 field to languageData.ocr as well, using the current stupid multi language setup.

## Proposed fix
I think we should change languageData so that none of the fields are multilingual - they should all just be plain english text fields.  OCR should either:
 - To keep things simple, we could stop supporting the situation where a document has been OCRd in multiple languages, and instead just pick whichever OCR field has more words in it or something, and then just have a single `languageData.ocr.translation` field. 
 - Or, we can keep the current structure of `languageData.ocr.translation` but make it a simple objectField with a dynamic mapping, so we aren't generating the language mapping for every language in languages.scala **Phil this this might be simplest **

For either solution, we would still need to introduce a `translation2` field or something to deal with the fact that we can't change the mapping of an existing index. 

Sorry to dump this right before going away. I think it's a day or two of work to sort out

philmcmahon:
## What does this change?

This PR aims to resolve https://github.com/guardian/giant/issues/789. It significantly simplifies the way we are modelling translation data in giant.

Prior to this PR, we stored translation information in a block that looked like this

```
text: "Bonjour",
ocr: { french: "bonjour", english: "bonjour"},
languageData: {
  text: {
    detectedLanguageCode: "fr",
    translation: "good day",
  },
  ocr: {
     detectedLanguageCode: {
      english: "fr",
      french: "fr",
    },
    translation: {
       english: "good day",
       french: "good day",
    }
}
```

Note the `ocr` block has multiple different translation/detectedLanguageCode entries for the different languages we told tesseract to use to OCR the document.

This was a bit of a nightmare in terms of complexity and due to the reasons discussed in the issue above. This PR simplifies the structure by changing the translation behaviour so that we pick the 'best' OCR output and only translate that. The new structure looks like this:

```
text: "Bonjour",
ocr: { french: "bonjour", english: "bonjour"},
translationData: {
  text: {
    detectedLanguageCode: "fr",
    translation: "good day",
  },
  ocr: {
     ocrLanguage: "french",
     detectedLanguageCode: "fr"
    translation: "good day"
}
```

## Choosing the 'best' OCR version

Selecting the 'best' OCR output to use for translation is handled by the function `bestOcrLanguage` in the BaseOcrExtractor. To get this working I had to add a new `languageDetectChunked` endpoint that splits a document into 3 chunks, runs language detection on all of them and returns the count of chunks where the most common language matched. This chunked approach helps to deal with situations where you, e.g. OCR a screenshot of a russian site with an english language cookie banner, telling tesseract to use 'english'. All the russian text will get garbled by tesseract but the cookie banner will be nicely OCRd. In this situation the 'chunk score' is useful as it reveals that even though tikka is 99% confident that that the cookie banner has valid english in it, this english only appears in one of the 3 chunks.

It picks the OCR text version based on:
 - does detected language code matches the language the document was OCR'd in (most important)
 - what was the confidence score of the  tikka language detector
 - which language appears in the most 'chunks' of the chunked language detection

I've (asked the llm to) write a load of unit tests for this 'best' selector, using text I manually generated using tesseract. 

## Mapping awkwardness

Note that above I've renamed `languageData` to translationData - this is because in elasticsearch you can't change the mapping once you've added it - the process would be to create a new index with the correct mapping and then reindex everything. Not impossible, but we'd have to scale up to make enough space for having 2x the amount of data.

Since the translation functionality has only existed for a few weeks and we have only one reported user, I'm proposing to just abandon languageData - it will still exist in the mapping, but no new documents will write to it and there will be no references in the code base. If I get a chance I can go back and migrate all the documents uploaded with a languageData block to the new translationData format just so as not to confuse future devs wondering what languageData is doing in the mapping.

 - [x] Tested locally
 - [x] Tested on playground

## Tasks prior to deployment to PROD
Migrate documents we know about from users who are using the translation feature to the new format - would be a shame for them all to break. I think it might be easiest to manually edit them in elasticsearch assuming there's only one or two

## Things to think about


backend/app/extraction/ocr/BaseOcrExtractor.scala
```diff
@@ -50,3 +55,49 @@ abstract class BaseOcrExtractor(scratchSpace: ScratchSpace, index:Index)  (impli
...
+  case class BestLanguageResult(score: Double, chunkCount: Int, language: Language, matchesOcrLanguage: Boolean)
+
+  case class ChunkedDetectionResult(languageCode: String, chunkCount:Int)
+
+  /**
+   * We OCR once per ingestion language, so a Russian document is also OCR'd in English, producing garbage. Here we use
+   * 3 signals to pick the 'best' language - does it match the OCR language, what is the tikka confidence and how many
+   * chunks did that language get selected in.
+   */
+  private[ocr] def bestOcrLanguage(textByLanguage: Map[Language, String], languageDetector: LanguageDetector): Option[Language] = {
```
philmcmahon, line 68: this is where the magic happens - main interesting bit of logic in this pr

backend/app/extraction/ExternalTranscriptionWorker.scala
```diff
@@ -173,11 +173,14 @@ class ExternalTranscriptionWorker(manifest: WorkerManifest, sqsClient: SqsClient
 
   // we shouldn't trust arbitrary field names we read from SQS
   private def discardInvalidFields(fields: List[TranslationField]): List[TranslationField] = {
-    val validFields = List(IndexFields.languageData.textField, IndexFields.languageData.emailBodyField, IndexFields.languageData.emailSubjectField)
-    val validOcrFields = Languages.all.map( lang => s"${IndexFields.languageData.ocr}_${lang.key}").toList
-    fields.filter { field =>
-      validFields.contains(field.name) || validOcrFields.contains(field.name)
-    }
```
philmcmahon, line 180: validOcrFields no longer needed with the simplified approach

backend/app/model/index/TranslationData.scala
```diff
@@ -0,0 +1,18 @@
+package model.index
```
philmcmahon, line 1: This replaces LanguageData

AGENTS.md
```diff
@@ -5,3 +5,10 @@ frontend files to be in typescript. When editing existing javascript files they
 
 We prefer a functional style of programming in both scala and typescript but prioritise readability over functional purity, 
 and simplicity over sophistication.
+
+
+When making changes. step back and think really hard:
+
+    What is the goal of this change?
+    How can we make this simpler and dumber while still achieving our goals?
+    How can we achieve our goals using a shorter diff that makes it easier to review and introduces less risk?
```
philmcmahon, line 14: not relevant to this PR but I think an improvement...

philmcmahon: Note - copied from https://github.com/guardian/agent-skills/blob/main/prompts/simplify-goal.prompt.md

backend/app/extraction/ocr/BaseOcrExtractor.scala
```diff
@@ -50,3 +55,44 @@ abstract class BaseOcrExtractor(scratchSpace: ScratchSpace, index:Index)  (impli
...
+  /**
+   * Attempts to pick the best OCR language and, if non english, add a translation extractor TODO
+   */
+  def handleOcrTranslation(uri: Uri, textByLanguage: Map[Language, String], index: Index,
+                           ingestionServices: IngestionServices, params: ExtractionParams)(implicit ec: ExecutionContext): Unit = {
+
+    val bestLanguage = bestOcrLanguage(textByLanguage, ingestionServices.languageDetector.get())
+    // if we get a decent match, save it in the translation data even if it's english
+    bestLanguage.foreach { lang =>
+      index.addDocumentOcrTranslationData(uri, lang, lang.iso6391Code).awaitEither(10.second)
```
twrichards, line 89: this should probably reference the fact that the `englishTranslation` field will be populated empty, and only filled out with the result of the external translation job if its not english

backend/app/services/ingestion/IngestionServices.scala
```diff
@@ -37,21 +37,77 @@ private case class UriParentPair(child: Uri, parent: Uri) extends UriParent
...
-  def detectLanguage(blobUri: String, text: String): Option[String]
+  def detectLanguage(blobUri: String, text: String): Option[LanguageDetect]
+  def detectLanguageChunked(blobUri: String, text: String): Option[LanguageDetect]
   def addTranslationTodo(blobUri: Uri, params: ExtractionParams, extractorName: String): Either[Failure, Unit]
 }
 
 object IngestionServices extends Logging {
 
+  def isNotEnglish(languageCode: String): Boolean = {
+    languageCode.toLowerCase != English.iso6391Code
```
twrichards: worth lowercasing "iso6391Code" too (I know they're supposed to be lowercase already) but just to be safe, given you feel the need to lowercase the incoming `languageCode` which presumably is expected to be a iso6391 from somewhere else anyway

backend/app/services/ingestion/IngestionServices.scala
```diff
@@ -37,21 +37,77 @@ private case class UriParentPair(child: Uri, parent: Uri) extends UriParent
...
+  }
+
+  /***
+    * Uses tika to detect the iso639-1 language code for the given text, along with how confident it is in that
+    * result. Callers that store the language should only use the result when it is `certain`, to try and maintain
+    * index quality. See https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes for a list of codes
+    */
+  def detectLanguage(languageDetector: LanguageDetector, fieldIdentifier: String, text: String): Option[LanguageDetect] = {
+    // clean a large block of text in case the file starts with lots of junk
+    val textToClean = text.take(50000)
```
twrichards, line 81: word it be worth taking the middle X amount

twrichards (approved):
LGTM and I know you've tested in playground

Awesome work 👏  and thank you for the PR description and IRL chats about this PR.

backend/app/extraction/ocr/BaseOcrExtractor.scala
philmcmahon: this made me realise there was no need to populate englishTranslation at all till we have something to put in it - resolved in https://github.com/guardian/giant/pull/806/commits/febf9cab84d8c8e01905ca80bf6b69cafa28c96e

backend/app/services/ingestion/IngestionServices.scala
philmcmahon: we could, but I think the middle could also be full of junk, this is more to just get a big sample size which then gets filtered down by the junk cleaner

PR #809, opened 2026-08-18
Add devcontainer config and docs to giant

philmcmahon:
## What does this change?
This PR adds configuration for a [dev container](https://containers.dev/overview) using [guardian/devenv](https://github.com/guardian/devenv)

The readme updates explain how it works. The aim here is to provide an easily usable sandboxed dev environment to protect the dev machine from supply chain attacks or AI accidents, and to potentially make it easier to run giant locally.

Most of the changes in this PR to existing code are to get our various scripts working on linux - hence the tweaks to the bash code. 

Note that I've used the same ports for the dev container as you would running giant directly locally so you won't be able to run giant on your laptop and in the dev container at the same time.

To test this PR:
 - check out this branch
 - run `mise install` to make sure you've got devenv installed
 - run `devenv generate` - this will generate your 'user' devenv configuration. @adamnfish-gu and I noticed that currently `devenv` only sets the right memory settings within the 'user' configuration.
 - open up .devcontainer/user/devcontainer.json and right click> dev containers>create dev container and clone sources (you can also mount sources if you want to sync your local project dir and the dev container, with some loss of isolation)
 - this will eventually open a new (or replace your current) intellij window, where you would then follow the readme instructions using the IDE terminal to setup giant. You can close the 'local' intellij window in most cases, I kept it open whilst doing this work as it has some convenient shortcuts for restarting the dev container etc - but that can also be done with docker.

 - [x] Tested locally


srbd (approved):
I tested this, works 👌 nice

love containers

PR #810, opened 2026-08-18
Don't return an error in translation extractor if nothing to translate

philmcmahon:
## What does this change?
I noticed a lot of errors in giant related to translation failing because there wasn't anything to translate. I don't think this is an error case, so I've stopped failing the extractor in this situation.

I've also added a minimum length at which text will be translated of 10 characters - could potentially increase this for text and ocr fields but keep it short for emailSubject, but let's see how this goes for now

 - [ ] Tested locally
 - [x] Tested on playground

### DIFF
--- a/backend/app/extraction/DocumentBodyExtractor.scala
+++ b/backend/app/extraction/DocumentBodyExtractor.scala
@@ -53,7 +53,7 @@ class DocumentBodyExtractor(tika: Tika, index: Index, ingestionServices: Ingesti
         val rawMetadata = metadata.names().map(name => name -> metadata.getValues(name).toSeq).toMap
         val documentBodyDetectedLanguage = ingestionServices.detectLanguage(blob.uri.value, body.trim()).flatMap(_.certainLanguage)
         // if we managed to detect the language and it's not english then add a translation extractor TODO
-        if (documentBodyDetectedLanguage.exists(_ != "en")) {
+        if (documentBodyDetectedLanguage.exists(_ != "en") && body.trim().length > IngestionServices.TRANSLATION_MINIMUM_LENGTH) {
           ingestionServices.addTranslationTodo(blob.uri, params, classOf[ExternalDocumentTranslationExtractor].getSimpleName)
         }
         val enrichedMetadata = MetadataEnrichment.enrich(rawMetadata)
--- a/backend/app/extraction/ocr/BaseOcrExtractor.scala
+++ b/backend/app/extraction/ocr/BaseOcrExtractor.scala
@@ -89,7 +89,7 @@ object BaseOcrExtractor extends Logging {
       index.addDocumentOcrTranslationData(uri, lang, lang.iso6391Code).awaitEither(10.second)
     }
     // if the best language is not english, add translation extractor TODO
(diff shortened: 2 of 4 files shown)


backend/app/services/ingestion/IngestionServices.scala
```diff
@@ -67,6 +67,8 @@ trait IngestionServices {
 
 object IngestionServices extends Logging {
 
+  val TRANSLATION_MINIMUM_LENGTH = 10
```
twrichards: hmm, I wonder if this should be lower, maybe 3? I can think of one word email subjects that would benefit from translation.

backend/app/extraction/DocumentBodyExtractor.scala
```diff
@@ -53,7 +53,7 @@ class DocumentBodyExtractor(tika: Tika, index: Index, ingestionServices: Ingesti
         val rawMetadata = metadata.names().map(name => name -> metadata.getValues(name).toSeq).toMap
         val documentBodyDetectedLanguage = ingestionServices.detectLanguage(blob.uri.value, body.trim()).flatMap(_.certainLanguage)
         // if we managed to detect the language and it's not english then add a translation extractor TODO
-        if (documentBodyDetectedLanguage.exists(_ != "en")) {
+        if (documentBodyDetectedLanguage.exists(_ != "en") && body.trim().length > IngestionServices.TRANSLATION_MINIMUM_LENGTH) {
```
twrichards, line 56: strictly it needs to be a greater than or equal to if the variable is 'minimum'
```suggestion
        if (documentBodyDetectedLanguage.exists(_ != "en") && body.trim().length >= IngestionServices.TRANSLATION_MINIMUM_LENGTH) {
```

backend/app/extraction/ocr/BaseOcrExtractor.scala
```diff
@@ -98,7 +98,7 @@ object BaseOcrExtractor extends Logging {
       index.addDocumentOcrTranslationData(uri, lang, lang.iso6391Code).awaitEither(10.second)
     }
     // if the best language is not english, add translation extractor TODO
-    bestLanguage.filter(lang => isNotEnglish(lang.iso6391Code))
+    bestLanguage.filter(lang => isNotEnglish(lang.iso6391Code) && textByLanguage.get(lang).exists(_.length > IngestionServices.TRANSLATION_MINIMUM_LENGTH))
```
twrichards: strictly it needs to be a greater than or equal to if the variable is 'minimum'
```suggestion
    bestLanguage.filter(lang => isNotEnglish(lang.iso6391Code) && textByLanguage.get(lang).exists(_.length >= IngestionServices.TRANSLATION_MINIMUM_LENGTH))
```

twrichards (approved):
nice improvement! couple of suggested tweaks

backend/app/services/ingestion/IngestionServices.scala
philmcmahon: I'll drop to 5, I was thinking of situations where the subject is just `FW:` or `RE:` which aren't valid english words but could get detected as another language potentially

PR #811, opened 2026-08-21
If a PDF page contains a massive vector, rasterize it with ghostscript

philmcmahon:
## What does this change?

Giant, specifically ocrmypdf, has a problem with PDF files containing massive vector images. (e.g this one https://drive.google.com/drive/u/0/folders/15YG293eQRWbueWXS0Ps4xZlT_FoV_hKD) 

Here we use PDFBox to detect pages with such a massive image and use ghostscript to rasterize the document at 300 DPI, deliberately discarding vector content. Note that the default ocrmypdf behaviour does rasterize at 400DPI but it preserves vector content (flag png16m vs pdfimage24)

A downside of this approach is that it rasterizes the whole PDF even if there's just one page with a giant vector in it. It's definitely possible to just rasterize the page with the vector in it, but it adds quite a bit of complexity - you have to split the PDF and then re-merge it back together I think. I think we can accept the reduced quality because:
 - a document with a vector in it is almost definitely a digital PDF rather than a scan, so the machine readable text will suffice without OCR for most of the text
 - OCR will still be run on the rasterized text so it should be searchable
  - we are hoping to improve our ocr engine anyway, at which points these pre process steps will probably change

Given that we are possibly going to replace ocrmypdf entirely, I think this might be ok for now (though I'm sure Visual OCR will provide a new host of problems...)

 - [x] Tested locally
 - [x] Tested on playground
 - [x] Tested in prod 😨

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,2 @@
 nodejs 24.15.0
 java corretto-21.0.10.7.1
-sbt 1
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,3 @@
 nodejs 24.15.0
 java corretto-21.0.10.7.1
+sbt 1
--- a/backend/app/extraction/ocr/OcrMyPdfExtractor.scala
+++ b/backend/app/extraction/ocr/OcrMyPdfExtractor.scala
@@ -54,22 +54,21 @@ class OcrMyPdfExtractor(scratch: ScratchSpace, index: Index, pageService: Pages,
 
     try {
       pdDocuments = params.languages.map { lang =>
-        // A single pdfbox parse of the original file, used both for the page count and to spot pages with
-        // pathologically large content streams
-        val inspection = Using(PDDocument.load(file)) { doc =>
+
+        val docInspection: Try[(Int, Boolean)] = Using(PDDocument.load(file)) { doc =>
           (doc.getNumberOfPages, Ocr.hasLargeVectorContent(file, doc))
         }
-        inspection.failed.foreach(e => logger.warn(s"Failed to inspect ${blob.uri} with pdfbox", e))
+        docInspection.failed.foreach(e => logger.warn(s"Failed to inspect ${blob.uri} with pdfbox", e))
 
-        val pages = inspection.map(_._1).toOption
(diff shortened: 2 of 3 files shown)


backend/app/extraction/ocr/OcrMyPdfExtractor.scala
```diff
@@ -54,13 +54,21 @@ class OcrMyPdfExtractor(scratch: ScratchSpace, index: Index, pageService: Pages,
...
       pdDocuments = params.languages.map { lang =>
-        val pages = Using(PDDocument.load(file))(_.getNumberOfPages).toOption
+
+        val docInspection: Try[(Int, Boolean)] = Using(PDDocument.load(file)) { doc =>
+          (doc.getNumberOfPages, Ocr.hasLargeVectorContent(file, doc))
+        }
+        docInspection.failed.foreach(e => logger.warn(s"Failed to inspect ${blob.uri} with pdfbox", e))
+
+        val pages = docInspection.map(_._1).toOption
+        val largeVectors = docInspection.map(_._2).getOrElse(false)
```
twrichards: nit: some sort of pattern match would be a it nicer than `_1` and `_2`

.tool-versions
```diff
@@ -1,2 +1,3 @@
 nodejs 24.15.0
 java corretto-21.0.10.7.1
+sbt 1
```
twrichards: was this intentional?

twrichards (approved):
LGTM and you've tested in PROD already and it's been successful

.tool-versions
philmcmahon: oops no, accidentally added from https://github.com/guardian/giant/pull/809

backend/app/extraction/ocr/OcrMyPdfExtractor.scala
philmcmahon: good shout [github.com/guardian/giant/pull/811/commits/06103e888f4dc7e2e696b1ffa10a96e54a3e2719](https://github.com/guardian/giant/pull/811/commits/06103e888f4dc7e2e696b1ffa10a96e54a3e2719)

PR #818, opened 2026-09-03
Migrate frontend build from Create React App to Vite

philmcmahon:
# Human note

I've tested this on playground. Once you get past the mountain of js > jsx renames this is a pretty short PR to review. Create react app was EOL in february last year so this feels like a good change to make.

I didn't want to end up with 'npm legacy peer resolution' on forever so there's already a follow up PR that bumps elastic ui so we can switch that off

# AI notes
## What does this change?

- Replaces the Create React App development and production build with Vite
- Renames JavaScript files containing JSX
- Reproduces the existing API, setup, and third-party development proxies
- Keeps production output in `frontend/build` for the existing Play packaging flow
- Upgrades TypeScript and adds an explicit type-check before production builds
- Keeps the existing CRA test runner temporarily; a stacked follow-up PR migrates tests to Vitest and removes CRA completely

## Compatibility notes

- Uses esbuild for CSS minification because Lightning CSS rejects a selector in the legacy Semantic UI stylesheet
- Adds the missing `moment-timezone` peer dependency required by Elastic Charts
- Enables npm legacy peer resolution because the pinned Elastic EUI version declares TypeScript 4.5 even though the project and current transitive types type-check on TypeScript 5.9 (fixed in #820)

## Validation

- `npm ci`
- `npm run typecheck`
- `npm run build`
- `CI=true npm test -- --runInBand` (27 suites, 237 tests)


frontend/src/js/util/store.js
```diff
@@ -27,14 +27,6 @@ function configureStore(history) {
     ),
   );
 
-  // eslint-disable-next-line no-redeclare
-  /* globals module:false */
-  if (module.hot) {
-    module.hot.accept("../reducers", () => {
```
twrichards, line 33: I take it HMR still works after this change?

twrichards (commented):
LGTM - nice work (good to be off CRA) - take it you've tested

frontend/src/js/util/store.js
philmcmahon: yep I tested that

philmcmahon:
Thanks for the review @twrichards  - I just rebased and tested again on playground. Could you give the final ✅  ?

zekehuntergreen (approved):
looks good!

PR #819, opened 2026-09-03
Migrate frontend tests from Create React App to Vitest

philmcmahon:
# Human bit 

Following from #818 this gets rid of the rest of create-react-app by moving to Vitest. 

# AI bit
## What does this change?

This is stacked on #818 and should be reviewed/merged after it.

- Replaces the CRA/Jest test runner with Vitest and jsdom
- Converts the small number of Jest-specific mock calls to Vitest equivalents
- Removes `react-scripts`, Jest types, and the CRA ESLint preset
- Adds explicit ESLint configuration and runs linting in CI
- Updates Ramda from 0.26 to 0.29, matching the existing type package and avoiding the old exported `then` function being treated as a thenable by the ESM test loader

## Validation

- Clean `npm ci`
- `npm run lint`
- `npm run build`
- `npm test` (27 suites, 237 tests)
- `npm run prettier:check`

PR #820, opened 2026-09-03
Upgrade TypeScript and Elastic UI

philmcmahon:
# Human bit

Following #818 and #819, this upgrades TypeScript and Elastic UI as far as the current React version allows without a larger application migration. My primary motivation here was to get rid of the legacy-peer-deps setting added in #818 - but it's quite a massive elastic version bump so hopefully will resolve various vulnerabilities as well.

Tested on playground

# AI bit
## What does this change?

This is the third PR in the stack and should be reviewed/merged after #819.

- Upgrades TypeScript from 5.9.2 to 5.9.3
- Upgrades Elastic UI from 77.2.0 to 106.0.0
- Adds EUI's required Borealis theme peer dependency
- Makes two small EUI API substitutions and removes obsolete packaged CSS imports
- Declares `react-input-autosize` directly because it was previously supplied transitively by EUI
- Removes the temporary `legacy-peer-deps` setting introduced in #818; EUI 106 officially supports TypeScript 5

EUI 106.0.0 is the latest release compatible with React 16. Later EUI releases require a React upgrade, which is intentionally outside this change.

## Validation

- Clean `npm ci` with strict peer dependency resolution
- `npm run typecheck`
- `npm run build`
- `npm test` (27 suites, 237 tests)
- `npm run lint` (no errors; existing warnings only)
- `npm run prettier:check`


twrichards (approved):
LGTM

PR #824, opened 2026-09-04
[WIP] Use schema generated directly from transcription service types

philmcmahon:
## What does this change?
This PR needs quite a bit more work but I wanted to see what the line count diff looked like and keep an eye on the build status so posting this up to track the work.

This PR builds on another branch which has a vibe coded JSON schema to scala converter in it. This branch is the one that actually starts using the generated types

 - [ ] Tested locally
 - [ ] Tested on playground


twrichards (approved):
lovely stuff 🙌

PR #832, opened 2026-09-11
Add daily JavaScript-to-TypeScript migration PRs

philmcmahon:
# Human description
Giant needs migrating to typescript, this feels like a good task for automation with AI. This is the first time I've tried doing this kind of work running in an automated fashion rather than typing prompts in on my laptop, I'm interested to see if it's useful or not.

My basic idea here was that one-shotting the typescript migration ended in a bit of a mess last time I tried it, and there are actually quite a lot of decision to make along the way as you get into the redux bits. We also don't have a testing/review mechanism beyond devs having to read thousands of lines of migrated code or do manual testing. This approach means that the PRs should be relatively short (the size should be adjustable) and human-reviewable. At the same time I'm hoping [this work](https://github.com/guardian/giant/pull/835) on E2E tests will give us greater confidence in the PRs and reduce manual testing work

Here is an example PR from the first step in the plan https://github.com/guardian/giant/pull/839/changes

# AI description
Adds a daily GitHub Actions workflow that uses Codex to migrate roughly five related frontend JavaScript files to TypeScript and opens a PR. Batch size can vary to keep the diff reviewable, with useful types and Zod schemas using `safeParse` at JSON boundaries.

An ordered Markdown backlog covers all 119 remaining JS/JSX files in 34 steps: utilities, API contracts and Redux slices, UI consumers, then store wiring and startup. It records existing type overlaps, backend contract differences and smaller batches for large components. Each migration PR removes its completed file entries and empty steps from the plan, retaining incomplete work.

The workflow runs at 07:23 UTC, manually, or when its workflow file changes on main. It starts from the default branch and waits while the previous migration PR remains open. It checks the baseline and generated changes with build/typecheck, lint, tests, and formatting. Generated changes are limited to frontend/src and the migration plan, and must include an actual conversion.

Setup: add the `OPENAI_API_KEY` Actions secret and enable Actions PR creation. An optional `TYPESCRIPT_MIGRATION_MODEL` variable selects the model. The included README documents operation and the fact that PRs created with `GITHUB_TOKEN` do not trigger other workflows.

Validation: actionlint passed; the plan lists every tracked JS/JSX file exactly once and matches the fetched main inventory; eight isolated scope-guard scenarios passed, including plan updates, consolidation into existing TypeScript, and rejection of unrelated edits, plan-only changes, and plan deletion. The existing frontend formatting check passed during initial setup. The API-backed workflow has not been run.


twrichards (approved):
such a wicked idea @philmcmahon - thank you for setting this up, looking forward to seeing what it produces!

PR #833, opened 2026-09-11
Move fetchwork test to the start

philmcmahon:
## What does this change?
This test is still randomly failing  for reasons discussed here https://github.com/guardian/giant/pull/777 and here https://github.com/guardian/giant/pull/712

The separate manifest was a stupid idea as the underlying database container is still the same so it didn't help with anything

This is a much dumber fix that I think will probably work - just making this test run first so that there's no other data to mess with it. At the same time I've removed the redundant separate manifest

If CI passes I think this can safely go out


philmcmahon:
@copilot+gpt-5.6-luna I've attempted to fix the fetchWork test in this PR by moving it to the top of the file but I've messed something up - the CI build failed. Can you work out what I've done wrong and fix it

PR #835, opened 2026-09-11
Add initial playwright test infrastructure and genesis user test

philmcmahon:
## What does this change?
We currently spend quite a lot of time manually testing giant. In other teams people are having some success using AI generated playwright tests to check an application is working as expected.

This PR adds playwright to the frontend/ client project along with an initial test that checks that a genesis user can be created and we can login with that user after it has been created.

The tests have their own docker-compose setup keeping them isolated from whatever the local state of dev giant databases is. 

In a follow up branch I've added a test that checks that we can upload a document and the extractors pass. I think this is quite exciting as that's one of the first things I do when testing giant.

A big caveat is that the external extractors (transcription/translation) are disabled at the moment, we'll need to sort out the git submodule or another approach to run these e2e tests with both services.

## Methodology
Initially I did some work using the playwright agent stuff here https://playwright.dev/docs/test-agents but I could never really work out where the agent skills were being used versus the llm doing its own thing so I've not checked those in for now. 

Following a chat with @dominickendrick I refactored this to follow the pattern they're using on the editorial tools modernisation project (e.g. [restorer](https://github.com/guardian/flexible-restorer/tree/main/e2e-tests)):
 - Cucumber used to define tests in human readable language. In general this bit is probably best written by a human rather than generated [see here](https://github.com/guardian/giant/pull/835/changes#diff-331876ad110a9de66ecabe250021b59b093e60fb3491450d0122b89b4ec2847d).
 - All the e2e tests live in their own special folder

How the cucumber testing works:
 - edit or add a `.feature` file in e2e-tests/features
 - run `npm run test:list` - it will give you some code to copy paste into the relevant spec file
 - Then you can fill out the test manually or get the LLM to do it

## Future work
We could dockerize giant with all the various packages needed to run it rather than having to install them directly on the CI agent.

We'll be writing more tests. As we do this we are likely to want to parallelise them at which point we might want to start using https://testcontainers.com/ to manage having multiple different databases running at the same time. For now I think this setup keeps things simple and gets us off the ground with e2e testing.

Trying this out
 - tests are executed via `npm test` or if you want to watch `npm test:ui` in the e2e directory

 - [x] Tested locally
 - [x] Tested in CI


frontend/e2e/genesis.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    await expect(
+      loginPage.getByRole("heading", { name: "Login", exact: true }),
+    ).toBeVisible();
+    await loginPage
+      .getByPlaceholder("Username", { exact: true })
+      .fill(username);
+    await loginPage
+      .getByPlaceholder("Password", { exact: true })
+      .fill(password);
+    await loginPage.getByRole("button", { name: "Login", exact: true }).click();
```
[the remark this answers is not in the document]
philmcmahon: wrong-  frontend/src/js/App.jsx already redirects

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Upload test report and service logs
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: e2e-results
+          path: |
+            frontend/playwright-report/
+            frontend/test-results/
+            frontend/e2e-artifacts/
+          retention-days: 7
```
philmcmahon, line 46: Example report page from these artifacts. Also includes all the logs from the app and docker containers

Image

backend/conf/e2e.conf
```diff
@@ -0,0 +1,21 @@
...
+elasticsearch.hosts = ["http://127.0.0.1:19200"]
+elasticsearch.shards = 1
+s3.endpoint = "http://127.0.0.1:13900"
+
+pekko.remote.artery.canonical.port = 11234
+pekko.cluster.seed-nodes = ["pekko://pfi@localhost:11234"]
+
+# Genesis needs storage and authentication, but no background extraction services.
+worker.enabled = false
+worker.useExternalExtractors = false
```
philmcmahon, line 19: for now this is necessary. once we've done the submodule thing or something else we should be able to run the transcription service worker in these tests as well (though might need separate queues for the tests for when they're run on our local machines - could be quite a lot of test queues following queue gardens)

scripts/test-e2e.sh
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env bash
```
philmcmahon, line 1: this is probably the messiest part of the testing infrastructure. I was wondering about rewriting it in typescript but given that it's mostly calling out to npm/sbt etc maybe bash is fine

twrichards (approved):
despite being mostly a test skeptic, this is really nice @philmcmahon - nice one!!

(nicely explained too in PR description and README 🙏)

PR #843, opened 2026-09-16
Generate transcription worker interface from json schema

philmcmahon:
## What does this change?
This PR adds a new sbt project `transcriptionWorkerInterface` to giant to handle the interface with the transcription service.

It has 3 tasks:
 - fetchSchema - takes an optional branch argument, fetches [this file](https://github.com/guardian/transcription-service/blob/main/packages/common/schemas/worker-interface-schema.json), checks that it's valid JSON and saves it in the `transcription-worker-interface` folder
 - `generateInterface` - this converts the JSON schema into scala case classes, with a single output file in transcription-worker-interface/src/main/scala called `TranscriptionWorkerInterface.scala`. The code that actually does the converstion is  in `project/TranscriptionWorkerInterfaceGenerator.scala`. The conversion code is vibe coded and, whilst I've eyeballed it, I wasn't planning on doing a thorough review of. I think instead we can just review the output scala code. 
 - `checkInterface` - this is a basic check that ensures that the JSON schema we have checked in matches the generated scala case classes (i.e. have we run `generateSchema` after running `fetchSchema`)

## Reasoning behind this approach
I previous tried to achieve a similar result using https://github.com/joelittlejohn/jsonschema2pojo - which worked fine, but it's much nicer having proper scala case classes. The downside of this 'roll your own json to scala' approach is that we'll have to do any updates manually. I don't think json schema is moving that fast though, and am not imagining the interface between giant and the transcription service to demand the latest and greates JSON schema features. 

Another alternative would have been to use apache thrift, but that would have involved greater changes to the transcription service and rewriting the transcription schema as thrift. This way we have Zod>JSON Schema>Scala and no extra libraries to import.

This doesn't actually change giant itself - it just adds the functionality to generate the scala code from the schema, and adds CI steps to ensure the schema is up to date. The first step to actually using the schema is here https://github.com/guardian/giant/pull/824

 - [x] Tested locally
 - [ ] Tested on playground

## Things to think about
**No, not yet, but nearly!!**


project/plugins.sbt
```diff
@@ -7,4 +7,8 @@ addSbtPlugin("com.gu" % "sbt-riffraff-artifact" % "1.1.18")
 
 addSbtPlugin("com.eed3si9n" % "sbt-buildinfo" % "0.11.0")
 
-addSbtPlugin("com.github.sbt" % "sbt-native-packager" % "1.11.1")
\ No newline at end of file
+addSbtPlugin("com.github.sbt" % "sbt-native-packager" % "1.11.1")
+
+// Used by TranscriptionWorkerInterfaceGenerator (invoked by the
+// `generateTranscriptionWorkerInterface` task) to read the transcription worker interface JSON schema
+libraryDependencies += "com.lihaoyi" %% "ujson" % "3.1.4"
```
twrichards: is there a reason we can't use play json for this reading of the JSON?

project/TranscriptionSchemaFetcher.scala
```diff
@@ -0,0 +1,85 @@
...
+      case Seq(ref)     => ref
+      case moreThanOne  =>
+        fail(
+          "fetchSchema takes at most one argument, the branch (or tag/commit) to fetch the schema from, " +
+            s"but got: ${moreThanOne.mkString(" ")}"
+        )
+    }
+
+    val schemaUrl =
+      s"https://raw.githubusercontent.com/guardian/transcription-service/$branch/$schemaPathInRepository"
```
twrichards, line 43: if we go ahead with git submodules (in https://github.com/guardian/giant/pull/792) we can probably do this from the local copy

project/TranscriptionSchemaFetcher.scala
```diff
@@ -0,0 +1,85 @@
...
+    val downloaded = IO.read(tempFile, StandardCharsets.UTF_8)
+
+    try ujson.read(downloaded)
+    catch {
+      case e: Exception =>
+        fail(
+          s"The content downloaded from $schemaUrl is not valid JSON (${e.getMessage}). " +
+            s"Refusing to overwrite $schemaFile."
+        )
+    }
```
twrichards: I think with play json you could just do something like `Json.parse(downloaded).validate[JsObject]`

twrichards (approved):
this is pretty great - nice one @philmcmahon ! couple of comments regarding the additional dependency vs just using play json. 

Looking forward to seeing #824 when it's ready 🙏

project/plugins.sbt
philmcmahon: good shout, we still need to add it as a dependency here so it's available in the tasks but nice to not add another json library, resolved in https://github.com/guardian/giant/pull/843/commits/9b8b5b186dfb605f31db6efe30aa777930488ae1

PR #844, opened 2026-09-16
Speed up CI setup, container startup and packaging

philmcmahon:
# human 
Giant builds are pretty slow, running between 10-13mins. This PR aims to speed them up by:

 - Using the sbt action rather than our custom setup
 - Starting the elastic/neo4j containers at the same time rather than in serial
 - Not bundling scala docs with the output artifact
 - Not generating a zip package for anything except the cli (we only use the debs)

In a sample size of one it took 8.5 mins for the build, which is a 2-4 minute speedup. Not life changing but seems worth it for a fairly harmless change

# 🤖 
## What does this change?

Reduce CI setup, container startup and packaging work. A recent successful build took 10m32s, including 52s installing sbt, 5m12s running integration tests and 48s packaging artifacts.

- Replace the apt-based sbt installation with `sbt/setup-sbt@v1`, already used by the E2E workflow. Keep the apt index refresh beside the remaining native-library installation.
- Start Elasticsearch and Neo4j concurrently with `Startables.deepStart(...).join()` in the five suites that need both. Each suite still owns and stops its own containers, and its tests retain their existing ordering.
- Disable Scaladoc artifact publication and Play's inclusion of API documentation in deployment packages. Packaging previously generated documentation twice.
- Scope tarball generation to `cli/Universal/packageZipTarball`; both Debian packages are still built, but the unused backend tarball is no longer generated.

Baseline: https://github.com/guardian/giant/actions/runs/35115553267. End-to-end savings need measurement on the CI runner.

- [x] `runAllTests`: 237 tests passed, including all 118 integration tests. Logs confirm concurrent container startup.
- [x] `Debian/packageBin cli/Universal/packageZipTarball`: both Debian packages and the CLI tarball built successfully (after installing the local `fakeroot` prerequisite).
- [x] Checked all three artifacts contain their launchers and libraries, with no API documentation; no backend tarball or Scaladoc directories were generated.
- [x] Frontend production build, workflow YAML parsing and shell syntax checks.
- [x] Tested on playground (not deployed).


twrichards (approved):
brilliant - well worth it - thank you!

PR #846, opened 2026-09-16
Add Cucumber workspace upload and extraction E2E test

philmcmahon:
## What does this change?

Adds a Cucumber scenario that logs in, creates a workspace, uploads a PDF and verifies that extraction finishes with a document icon and `processed` status. An extraction failure cannot count as success.

The scenario reuses the existing login step and explicitly depends on the genesis project. Local extraction workers are enabled, CI installs the required PDF tools, and the E2E README documents the setup.

The PDF fixture is rasterised at 300 DPI so Giant rebuilds its text through OCR. This avoids malformed references and font mappings in the original printout that caused CI to detect Welsh and request an unavailable translation extractor.

- [x] Tested locally: `npm test` in `e2e-tests` passed both genesis and workspace upload scenarios against a fresh disposable Giant instance.
- [x] `npm run test:list`, `npm run typecheck`, and `npm run format:check` passed.
- [x] Reproduced the original fixture failure with Ubuntu 24.04 extraction tools and Giant's PDFBox/language detector; the repaired fixture is detected as English in all three text chunks.
- [ ] Tested on playground

## Things to think about

External transcription and translation extractors remain disabled in the E2E environment. This scenario exercises local PDF extraction.


e2e-tests/fixtures/toast_sandwich_en_wiki.pdf
SHession, line 1: 🥪

SHession (commented):
Change look good, I noticed a couple of things running it locally worth considering:

1. The playwright port is not current included in the `devcontainer.json` spec. Should it be added?
2. When running the tests in the Playwright UI, if I attempt to rerun the test the genesis tests are also rerun which results in the suite failing. I could get around this be unticking the genesis project

Image

SHession (approved):

philmcmahon:
> Change look good, I noticed a couple of things running it locally worth considering:
> 
>     1. The playwright port is not current included in the `devcontainer.json` spec. Should it be added?
> 
>     2. When running the tests in the Playwright UI, if I attempt to rerun the test the genesis tests are also rerun which results in the suite failing. I could get around this be unticking the genesis project
> 
> 

good points, I've created an issue here https://github.com/guardian/giant/issues/856 to track this - think we should resolve both things, but think it could go in a follow up bit of work

